PR #6, opened 2020-07-30
[datadog] add `install_info` file

L3n41c:

Add the `install_info` file that is used by the agent to report how it was installed.

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/install_info-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+  annotations:
+    checksum/install_info: {{ printf "%s-%s" .Chart.Name .Chart.Version | sha256sum }}
+data:
+  install_info: |
+    ---
+    install_method:
+      tool: helm
+      tool_version: {{ .Release.Service }}
```
vboulineau, line 18: From doc `.Release.Service` is always `Helm`. Not sure if we can get Helm version, otherwise maybe we should put `Capabilities.KubeVersion`

charts/datadog/templates/install_info-configmap.yaml
```diff
@@ -0,0 +1,19 @@
...
+      installer_version: {{ .Chart.Name}}-{{ .Chart.Version }}
```
vboulineau, line 19: nit: space `.Chart.Name }}`

charts/datadog/templates/install_info-configmap.yaml
L3n41c: I saw that in the doc. So, I wondered what the value would be with Spinnaker. (Given the fact that Spinnaker re-implements the Helm features without relying on it.)

`KubeVersion` can be interesting, but it’s not really linked to the installation method.

charts/datadog/templates/install_info-configmap.yaml
L3n41c: Fixed !

charts/datadog/templates/container-agent.yaml
```diff
@@ -83,6 +83,14 @@
 {{ toYaml .Values.agents.containers.agent.env | indent 4 }}
 {{- end }}
   volumeMounts:
+    - name: installinfo
+      subPath: install_info
+      {{- if eq .Values.targetSystem "windows" }}
+      mountPath: C:/datadog-agent/install_info
```
vboulineau, line 89: I think default path on Windows is `c:\programdata\datadog\`

L3n41c: Good catch! Fixed!

PR #8, opened 2020-08-04
Add datadog-operator chart

vboulineau:

Add chart for `datadog-operator` in our main repository

Allows to use operator easily

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/templates/deployment.yaml
+++ b/charts/datadog-operator/templates/deployment.yaml
@@ -1,0 +1,85 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "datadog-operator.fullname" . }}
+  labels:
+{{ include "datadog-operator.labels" . | indent 4 }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: {{ include "datadog-operator.name" . }}
+      app.kubernetes.io/instance: {{ .Release.Name }}
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: {{ include "datadog-operator.name" . }}
+        app.kubernetes.io/instance: {{ .Release.Name }}
+      annotations:
+        ad.datadoghq.com/{{ .Chart.Name }}.check_names: '["openmetrics"]'
+        ad.datadoghq.com/{{ .Chart.Name }}.init_configs: '[{}]'
+        ad.datadoghq.com/{{ .Chart.Name }}.instances: |
+          [{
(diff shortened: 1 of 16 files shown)


charts/datadog-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+        app.kubernetes.io/instance: {{ .Release.Name }}
+      annotations:
+        ad.datadoghq.com/{{ .Chart.Name }}.check_names: '["openmetrics"]'
+        ad.datadoghq.com/{{ .Chart.Name }}.init_configs: '[{}]'
+        ad.datadoghq.com/{{ .Chart.Name }}.instances: |
+          [{
+            "prometheus_url": "http://%%host%%:8383/metrics",
+            "namespace": "datadog.operator",
+            "metrics": ["*"]
+          }]
```
clamoriniere, line 26: we should have this as optional until we have a proper "datadog-operator" integration.
Else this metrics will be seen as custom metrics.

charts/datadog-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,85 @@
...
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+            - name: OPERATOR_NAME
+              value: {{ .Chart.Name }}
+          args:
+            - --zap-level={{ .Values.logLevel }}
+            - --zap-encoder=console
+            # - --zap-stacktrace-level=panic // TODO: uncomment after releasing v0.3.0
```
clamoriniere, line 51: I think we can use this option now.

crds/datadoghq.com_datadogmetrics_crd.yaml
```diff
@@ -0,0 +1,112 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
clamoriniere, line 1: maybe the installation of this CRD should be optional.

charts/datadog-operator/templates/deployment.yaml
vboulineau: We need to do an operator (docker) release first

charts/datadog-operator/templates/deployment.yaml
vboulineau: Indeed, we should not trigger custom metrics OOB

charts/datadog-operator/templates/deployment.yaml
clamoriniere: I think now it is ok

clamoriniere (approved):

PR #13, opened 2020-08-06
[orchestrator] add scrubbing option for helm charts

nammn:
related PR: https://github.com/DataDog/datadog-agent/pull/6158
#### Checklist
- [x] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -384,6 +384,12 @@ datadog:
     #
     enabled: false
 
+    ## @param scrubbing - boolean - required
+    ## Enable the scrubbing of containers for sensitive information.
+    ## The container scrubbing is taking significant resources during data collection.
```
hkaj, line 389: let's be more specific, this is for the kubernetes resources YAML, not just any containers. And after "The container scrubbing is taking significant resources during data collection." we can say "if you notice that the cluster-agent uses too much CPU in larger clusters turning this option off will improve the situation.

nammn: Makes sense for the second part. I will update this 👍 
But we only scrub `container` and `initContainers`. I could make the text to something like: `enable the scrubbing of containers in the kubernetes resource YAML...`

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -181,6 +181,10 @@ spec:
           - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
             value: "true"
           {{- end }}
+          {{- if .Values.datadog.orchestratorExplorer.container_scrubbing.enabled }}
```
hkaj, line 184: this doesn't work actually, we should check the value, not just if it's set. If i set it to false it still gets enabled

hkaj: same issue with `{{- if .Values.datadog.orchestratorExplorer.enabled }}` above (my fault!). Could you fix both please?

hkaj: i think we can just always set the env vars, but set them to the value of the options from values.yaml (since they're mandatory)

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -177,9 +177,13 @@ spec:
                 key: token
           - name: DD_KUBE_RESOURCES_NAMESPACE
             value: {{ .Release.Namespace }}
-          {{- if .Values.datadog.orchestratorExplorer.enabled }}
+          {{- if (.Values.datadog.orchestratorExplorer.enabled | quote) }}
```
vboulineau, line 180: Why changing this check with a quote? Quoting for `value` is useful but in `{{ if }}` I don't think so.

charts/datadog/templates/container-process-agent.yaml
```diff
@@ -24,13 +24,15 @@
     {{- end }}
     {{- if .Values.datadog.orchestratorExplorer.enabled }}
     - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
-      value: "true"
+      value: {{ .Values.datadog.orchestratorExplorer.enabled | quote }}
+    {{- if eq (.Values.datadog.orchestratorExplorer.enabled | quote) "true" }}
```
vboulineau, line 28: Same issue here

charts/datadog/templates/container-process-agent.yaml
```diff
@@ -24,13 +24,15 @@
     {{- end }}
     {{- if .Values.datadog.orchestratorExplorer.enabled }}
```
vboulineau, line 25: This `if` is not required and duplicated with the setting `DD_ORCHESTRATOR_CLUSTER_ID`

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -179,7 +179,11 @@ spec:
             value: {{ .Release.Namespace }}
           {{- if .Values.datadog.orchestratorExplorer.enabled }}
```
vboulineau, line 180: Same here, this if is not necessary

charts/datadog/templates/container-process-agent.yaml
```diff
@@ -22,9 +22,9 @@
     - name: DD_SYSTEM_PROBE_ENABLED
       value: {{ .Values.datadog.systemProbe.enabled | quote }}
     {{- end }}
-    {{- if .Values.datadog.orchestratorExplorer.enabled }}
     - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
-      value: "true"
+      value: {{ .Values.datadog.orchestratorExplorer.enabled | quote }}
+    {{- if .Values.datadog.orchestratorExplorer.enabled }}
```
nammn, line 27: just tested this with `template`. It still works. If the explorer is set to false, it will not reference the `datadog-cluster-id`

PR #16, opened 2020-08-07
Adding NET_BROADCAST capability for system-probe

hmahmood:
system-probe will need this capability to monitor nat info outside of the root namespace.

#### Checklist
- [x] Chart Version bumped


vboulineau:
Is this capabilities always used when system-probe is activated, or is it only required with some feature?

You also need to update PSP/SCC part.

hmahmood:
> Is this capabilities always used when system-probe is activated, or is it only required with some feature?
> 
> You also need to update PSP/SCC part.

This is only used when system-probe is activated.

hmahmood:
@vboulineau is this good to go?

PR #18, opened 2020-08-10
Add eventfd and eventfd2 to allowed syscalls for system-probe

brycekahle:

The cilium-based library needs these syscalls for perf buffer epoll usage.

#### Checklist
- [x] Chart Version bumped


CharlyF:
ah, seems like something was merged in between. Can you bump the version @brycekahle ? I will merge thereafter

PR #24, opened 2020-08-14
Fix external metrics on GKE due to Google fix on recent versions

vboulineau:

With latest GKE release (R26 regular/rapid channels), they fixed their issue with missing `external-metrics-reader` ClusterRole.

Thus adapting our chart not to create this ClusterRole on newer versions

  - fixes #

#### Checklist
- [x] Chart Version bumped


clamoriniere (approved):

danopia:
Hello, should there be a 1.15 version here too? I've updated a gke cluster `v1.15.12-gke.2` ▶️ `v1.15.12-gke.20` and see a `external-metrics-reader` ClusterRole created around the same time.

```
$ kubectl get ClusterRole|grep external-metrics-reader
datadog-cluster-agent-external-metrics-reader                          140d
external-metrics-reader                                                6d22h
```

I'm trying to update my Datadog chart `2.3.41` ▶️  `2.4.20` and it's trying to move to the external-metrics-reader name, which already exists. (It seems the user workaround is disabling `clusterAgent.metricsProvider.createReaderRbac` since gke seemingly made the RBAC already.)

vboulineau:
@danopia It's possible yes. When I created the PR, none of the 1.15 versions had the automatic creation of `external-metrics-reader`, but it might have received an update since. I will do a PR to add the version `v1.15.12-gke.20` there.

PR #25, opened 2020-08-17
fix PodSecurityPolicy

michaelst:
The PSP is currently not accessible to the service account so no pods can startup. Additionally if you are using dogstatsd it would not start due to using host port 8125.

This only gets the agent pods working with a PSP, cluster-agent and kube-state-metrics do not work with PSP enabled.

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/agent-psp.yaml
```diff
@@ -17,6 +17,11 @@ metadata:
 spec:
   privileged: {{ .Values.agents.podSecurity.privileged }}
   hostNetwork: {{ .Values.agents.useHostNetwork }}
+  {{- if .Values.datadog.dogstatsd.useHostPort }}
```
vboulineau, line 20: In fact we also need that for APM, so we can cover 8125-8126 is any of these two features is activated

```suggestion
  {{- if or .Values.datadog.dogstatsd.useHostPort .Values.datadog.apm.enabled }}
    hostPorts:
  - min: 8125
    max: 8126
```

michaelst: Thanks, I have applied the suggestion

vboulineau (approved):

PR #38, opened 2020-09-04
Add External Metric Aggregator config on Chart

brunopaiuca:
It defines the standard value as average for the variable for the
variable DD_EXTERNAL_METRICS_AGGREGATOR like the default behavior of
cluster agent, and also enable the user to use the other configs among
the options currently available (avg sum min max).

  - fixes #

#### Checklist
- [ ] Chart Version bumped
- [ ] Variables are documented in the README.md


vboulineau:
Hello @brunopaiuca,

Thanks for your contribution. There are few issues with your contribution before we can merge it:
- Bump Chart version
- Add variable in README
- We usually deal with default value by setting them in the `values.yaml` directly (no need to use the `default` function)

Also, if you're using the `DD_EXTERNAL_METRICS_AGGREGATOR`, you may be interested in the new DatadogMetric CRD, which allows custom queries and is already supported by this Helm chart, see:
https://docs.datadoghq.com/agent/cluster_agent/external_metrics/#autoscaling-with-custom-queries-using-datadogmetric-crd-cluster-agent-v1-7-0

brunopaiuca:
Thank you, I will adjust accordingly. I would like to use the CRD, but my
old k8s version don’t allow me.

PR #46, opened 2020-09-22
Change `securityContext` variable name to allow setting the PSP/SCC seLinux `type` or `rule`

vboulineau:

Change the `datadog.podSecurity.securityContext` to `datadog.podSecurity.seLinuxContext` for two reasons:
- Allow setting the `rule: ` or `type: ` field instead of hardcoded
- Clarify that it's not the same format as Pod `securityContext` field

#### Checklist
- [x] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -803,10 +803,11 @@ agents:
       #
       create: false
 
-    ## @param securityContext - object - required
-    ## Provide securityContext configuration
+    ## @param seLinuxContext - object - required
+    ## Provide seLinuxContext configuration for PSP/SCC
     #
-    securityContext:
+    seLinuxContext:
```
clamoriniere, line 809: this is a breaking change. 
I think we need to style support the previous variable name, we can comment it and flag it as deprecated

vboulineau: We do support it, though it's not documented anymore.

charts/datadog/templates/NOTES.txt
```diff
@@ -82,3 +82,14 @@ This version of the chart does not support deploying Agent 5.x.
...
 {{- end }}
+
+{{- if .Values.agents.podSecurity.securityContext }}
+
+#################################################################
+####               WARNING: Deprecation notice               ####
+#################################################################
+
+You are using the datadog.podSecurity.securityContext parameter, which has been renamed datadog.podSecurity.seLinuxContext.
+This version still supports datadog.podSecurity.securityContext, but it will be removed in a future version of our Helm chart.
```
clamoriniere, line 93:
```suggestion
This version still supports datadog.podSecurity.securityContext, but it will be removed in the next major version of our Helm chart.
```

clamoriniere (approved):

PR #50, opened 2020-09-27
Run datadog agent as a deployment

### LINKED ISSUE #48: How to run as a deployment instead of a daemonset?
**Describe what happened:**
On the README it is said: 
```
By default, the Datadog Agent runs in a DaemonSet. It can alternatively run inside a Deployment for special use cases.
```

I expected some kind of value do be able to run the agent as a Deployment instead of a replica

I'll have to fork the whole chart to be able to run the agent as a deployment. Is there another way?

AWS, EKS

bmbferreira:
On the README.md it is said that "By default, the Datadog Agent runs in a DaemonSet. It can alternatively run inside a Deployment for special use cases.", however with the current helm chart it is not possible to run the agent as a deployment, only as a daemonset. This can be usefull for instance if the user is not interested in the kubernetes metrics and wants to run only some custom metrics and some custom python scripts in the checks.d (which afaik is not possible with the cluster agent).

  - fixes #48 

I was not sure if I should maintain the name of the file as `daemonset.yaml` because now it can be a Deployment. 
I checked other examples and the [ambassador helm chart](https://github.com/helm/charts/tree/82151373e4c07f9b76b2ed2af3ffe3ce5859b13f/stable/ambassador) just called it [deployment.yaml](https://github.com/helm/charts/blob/82151373e4c07f9b76b2ed2af3ffe3ce5859b13f/stable/ambassador/templates/deployment.yaml) and it sets the kind like this:
```
{{- if .Values.daemonSet }}
kind: DaemonSet
{{- else }}
kind: Deployment
{{- end }}
```

While on the [haproxy ingress helm chart](https://github.com/helm/charts/tree/67de216e850b80f0622fe821ffb7eacfa7d37244/incubator/haproxy-ingress) the daemonset and deployment are separated in two different files like this:
```
# controller-deployment.yaml
{{ if eq .Values.controller.kind "Deployment" -}}   apiVersion: apps/v1   
kind: Deployment
```
and
```
# controller-daemonset.yaml
{{ if eq .Values.controller.kind "DaemonSet" -}}
apiVersion: apps/v1
kind: DaemonSet
```

Please tell me which you think it's best.

#### Checklist
- [x] Chart Version bumped


vboulineau:
Hi @bmbferreira,

Thanks for your contribution. Supporting the Agent as a deployment has been removed since version 2.0.0 of our Helm chart. Using the Agent as a deployment presents a number of issues, notably around tagging.

This chart represents the recommended way of deploying the Agent on K8S to have all features working out-of-the-box. As it's not something we can propose with the Agent as a deployment, we prefer not to have this option in this chart.

bmbferreira:
Hi @vboulineau, thanks for the quick reply. 
Our main goal with running it as a deployment would be to have just one pod to execute some custom metrics and checks since we were not interested in the metrics automatically gathered from the agent in kubernetes and it wouldn't make sense running it as a daemonset just for these custom metrics/checks. 

We ended up forking the repo and doing these changes but I thought that someone out there could also be interested in having this. Nevertheless, the documentation should be updated on the README if this is not supported.

vboulineau:
We should indeed clarify that using a Deployment is possible but not recommended, nor supported by this chart.

From what you describe, it may be possible to cover your use case with Cluster Agent/Cluster Check Runner (without Agent deployment).

bmbferreira:
ok, I'll try with the cluster agent

PR #71, opened 2020-10-26
Allow the system-probe to be run without network performance monitoring being enabled

leeavital:

To preserve backwards compatibility, there is some special behavior when systemProbe.enabled=true.

Please see the following truth table for what behaviors should be
available when different values are set:

```
systemProbe.enabled     runtimeSecurity.enabled        networkConfig.enabled  agent features
no                      no                             no                     nothing enabled
no                      no                             yes                    system probe enabled, network config enabled
no                      yes                            no                     system probe enabled, network config disabled
no                      yes                            yes                    system probe enabled, network config enabled
yes                     no                             no                     system probe enabled, network config enabled
yes                     no                             yes                    system probe enabled, networkConfig enabled
yes                     yes                            no                     system probe enabled, network config disabled
yes                     yes                            yes                    system probe enabled, network config enabled
```

This PR does a few things to make this work correctly:
* add a new networkConfig block that can be used to disable/enable
  network performance monitoring
* set systemProbe.enabled=true based on networkConfig
* use networkConfig.enabled when deciding whether or not to render
  system-probe-manifests

Additionally, this exposed a bug in agent 7.23.0 where
DD_SYSTEM_PROBE_ENABLED must be set to true for the connection check to
run, which is fixed as part of this PR. This was an impossible situation
before this PR, but now might happen if the system probe is enabled
without network performance monitoring.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -461,9 +461,9 @@ helm install --name <RELEASE_NAME> \
 | datadog.logs.containerCollectAll | bool | `false` | Enable this to allow log collection for all containers |
 | datadog.logs.containerCollectUsingFiles | bool | `true` | Collect logs from files in /var/log/pods instead of using container runtime API |
 | datadog.logs.enabled | bool | `false` | Enables this to activate Datadog Agent log collection |
-| datadog.neworkMonitoring.container_scrubbing.enabled | bool | `true` |  |
 | datadog.neworkMonitoring.enabled | bool | `false` |  |
 | datadog.nodeLabelsAsTags | object | `{}` | Provide a mapping of Kubernetes Node Labels to Datadog Tags |
+| datadog.orchestratorExplorer.container_scrubbing | object | `{"enabled":true}` | Enable the scrubbing of containers in the kubernetes resource YAML for sensitive information |
 | datadog.orchestratorExplorer.enabled | bool | `false` | Set this to true to enable the orchestrator explorer |
 | datadog.podAnnotationsAsTags | object | `{}` | Provide a mapping of Kubernetes Annotations to Datadog Tags |
 | datadog.podLabelsAsTags | object | `{}` | Provide a mapping of Kubernetes Labels to Datadog Tags |
@@ -484,7 +484,7 @@ helm install --name <RELEASE_NAME> \
 | datadog.systemProbe.enableConntrack | bool | `true` | Enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data |
 | datadog.systemProbe.enableOOMKill | bool | `false` | Enable the OOM kill eBPF-based check |
 | datadog.systemProbe.enableTCPQueueLength | bool | `false` | Enable the TCP queue length eBPF-based check |
-| datadog.systemProbe.enabled | bool | `false` | Enable the system-probe agent. |
+| datadog.systemProbe.enabled | bool | `false` | Set this to true to enable system-probe agent |
 | datadog.systemProbe.seccomp | string | `"localhost/system-probe"` | Apply an ad-hoc seccomp profile to the system-probe agent to restrict its privileges |
 | datadog.systemProbe.seccompRoot | string | `"/var/lib/kubelet/seccomp"` | Specify the seccomp profile root directory |
 | datadog.tags | list | `[]` | List of static tags to attach to every metric, event and service check collected by this Agent. |
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -457,8 +457,9 @@ helm install --name <RELEASE_NAME> \
 | datadog.logs.containerCollectAll | bool | `false` | Enable this to allow log collection for all containers |
 | datadog.logs.containerCollectUsingFiles | bool | `true` | Collect logs from files in /var/log/pods instead of using container runtime API |
 | datadog.logs.enabled | bool | `false` | Enables this to activate Datadog Agent log collection |
+| datadog.networkConfig.container_scrubbing.enabled | bool | `true` |  |
+| datadog.networkConfig.enabled | bool | `false` | Enable network performance monitoring. |
 | datadog.nodeLabelsAsTags | object | `{}` | Provide a mapping of Kubernetes Node Labels to Datadog Tags |
-| datadog.orchestratorExplorer.container_scrubbing | object | `{"enabled":true}` | Enable the scrubbing of containers in the kubernetes resource YAML for sensitive information |
 | datadog.orchestratorExplorer.enabled | bool | `false` | Set this to true to enable the orchestrator explorer |
 | datadog.podAnnotationsAsTags | object | `{}` | Provide a mapping of Kubernetes Annotations to Datadog Tags |
 | datadog.podLabelsAsTags | object | `{}` | Provide a mapping of Kubernetes Labels to Datadog Tags |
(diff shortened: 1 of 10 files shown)


charts/datadog/README.md
```diff
@@ -479,7 +480,7 @@ helm install --name <RELEASE_NAME> \
 | datadog.systemProbe.enableConntrack | bool | `true` | Enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data |
 | datadog.systemProbe.enableOOMKill | bool | `false` | Enable the OOM kill eBPF-based check |
 | datadog.systemProbe.enableTCPQueueLength | bool | `false` | Enable the TCP queue length eBPF-based check |
-| datadog.systemProbe.enabled | bool | `false` | Set this to true to enable system-probe agent |
+| datadog.systemProbe.enabled | bool | `false` | Enable the system-probe agent. DEPRECATED: use networkConfig or securityAgent blocks. |
```
clamoriniere, line 483: I think `Network` and `SecurityAgent` is not the 2 only functionality present in SystemProbe.
we also have `systemProbe.enableOOMKill` `systemProbe.enableTCPQueueLength `

charts/datadog/values.yaml
```diff
@@ -283,6 +283,10 @@ datadog:
     ## ref: TODO - add doc link
     enabled: false
 
+  networkConfig:
```
clamoriniere, line 286: should we move also `enableConntrack` and `collect_dns_stats` into this new section?

leeavital: That would be a breaking change, and these options should be changed very infrequently. I'd rather leave these here than try to add them to networkConfig in a backwards compatible way.

charts/datadog/templates/containers-init-linux.yaml
```diff
@@ -37,7 +37,7 @@
       mountPath: {{ print "/host/" (dir (include "datadog.dockerOrCriSocketPath" .)) | clean }}
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
-    {{- if or .Values.datadog.systemProbe.enabled .Values.datadog.securityAgent.runtime.enabled }}
+    {{- if or .Values.datadog.systemProbe.enabled .Values.datadog.securityAgent.runtime.enabled $.Values.datadog.networkConfig.enabled }}
```
clamoriniere, line 40: I think `or` only works with 2 args.

clamoriniere: Also this test is used several time. It could be good to have a function in `_helpers.yaml` to remove the duplication.

leeavital: I did a simple test and render:

```
helloworld: {{ or false false true }}
```

and it seems that `or` accepts any number of args

charts/datadog/values.yaml
clamoriniere: Sorry I haven't used the proper word.
do you think it could be a good idea to deprecate the `systemProbe.enableConntrack` and `systemProbe.collectDNSStats` and to replace them by `networkConfig.enableConntrack` and `networkConfig.collectDNSStats`

charts/datadog/README.md
leeavital: This is already a rather large change. Can I change the wording here and do those handle those two fields in a follow up change? 

system_probe.enabled will continue to work for tcp_queue_length and oom_kill as expected, so this won't break anything.

charts/datadog/values.yaml
clamoriniere, line 286: Have you considered to add `systemProbe.enableNetwork` (`true` by default for backward compatibility) instead of created a new section?

leeavital: There is a network_config section in the agent config, where new config will eventually be added. So I figured adding a new "namespace" was better for future compatibility.

leeavital: I don't feel like it's worth the effort of maintaining, these features are both fairly stable and on by default.

charts/datadog/README.md
clamoriniere: we could deprecate `systemProbe.enabled` in another PR too.

charts/datadog/values.yaml
clamoriniere, line 286: I'm wondering if the "namespace" should be more align with the product name `networkPerformanceMonitoring` or just "npm". 

I feel like with `networkConfig` people can misunderstood the purpose of this section.

leeavital: I agree that `networkConfig` could be misleading.   I chose  `networkConfig` for parity with the `nework_config` section in the system-probe.yaml file. That's already in 7.23.0 so we can't easily change it.

I'm not sure how much parity we should keep with the system-probe.yaml formatting in the helm options, which I could use feedback on. I'd avoid `npm` though, since that acronym usually refers to "node package manager"

clamoriniere: I think we don't need feature parity between the system-probe.yaml config file and the option in the helm chart. 
I know that in the past we weren't very "clean" about naming convention, but we should try to use  name that "talk" to the user.

clamoriniere: what do you think about a section like this
```yaml
  networkMonitoring:
    enabled: true
    config:
      // any config parameter here
```

leeavital:
I would up:
* removing the deprecation notice about systemProbe.enabled (in a follow up, I will add new flags to enable OOM kill and tcp_queue_length, and then deprecate systemProbe.enabled)
* changing the new config block to `networkMonitoring`. The right way to enable NPM is now to use `networkMonitoring.enabled`

charts/datadog/templates/container-process-agent.yaml
```diff
@@ -22,9 +22,13 @@
     {{- end }}
     - name: DD_LOG_LEVEL
       value: {{ .Values.agents.containers.processAgent.logLevel | default .Values.datadog.logLevel | quote }}
-    {{- if .Values.datadog.systemProbe.enabled }}
+    {{- if or .Values.datadog.systemProbe.enabled .Values.datadog.networkMonitoring.enabled }}
```
clamoriniere, line 25: why not using `{{- if eq (include "should-enable-system-probe" .) "true" }}` here?

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -13,14 +13,18 @@ metadata:
 data:
   system-probe.yaml: |
     system_probe_config:
-      enabled: {{ or $.Values.datadog.systemProbe.enabled $.Values.datadog.securityAgent.runtime.enabled }}
+      enabled: {{ or $.Values.datadog.systemProbe.enabled $.Values.datadog.securityAgent.runtime.enabled $.Values.datadog.networkMonitoring.enabled }}
```
clamoriniere, line 16: why not using `{{- if eq (include "should-enable-system-probe" .) "true" }}` here too

clamoriniere: Also since in line1 we already surround  the Configmap with `{{- if eq (include "should-enable-system-probe" .) "true" }}` 
we should be able to just set the value to `true`

```suggestion
      enabled: true
```

clamoriniere (commented):
Hi @leeavital 
sorry for the delay, Looks good 👍 
just few possible optimisations.

charts/datadog/templates/container-process-agent.yaml
leeavital: The process-agent should only be started with the system-probe for the network-check, so turning on just runtime security monitoring should not set  ENABLED=true.

clamoriniere (approved):

PR #83, opened 2020-11-04
Add options to set pod and container securityContext

### LINKED ISSUE #82: Add DD agent container securityContext field
Original issue by @billyshambrook (https://github.com/helm/charts/issues/16353) with fix PR by @coreypobrien (https://github.com/helm/charts/pull/17274 - stale)

The new Network Performance feature requires setting some container securityContext.capabilities - they are shown here under the "kubernetes" tab https://docs.datadoghq.com/graphing/infrastructure/network_performance_monitor/?tab=kubernetes

Currently only pod securityContext is exposed through values, not container securityContext.

Could we expose container securityContext as an optional value?

Cheers!

icelynjennings:
Adds options to specify securityContext settings for both pods and containers to handle special capabilities or more stringent security policies.

Fixes #82 

This PR is based on an older PR (https://github.com/helm/charts/pull/17274)

#### Checklist
- [x] Chart Version bumped


vboulineau:
LGTM, though you need to add the new options in `values.yaml` with empty default value. The doc generation will then generate the doc correctly.

icelynjennings:
Thank you @vboulineau. Docs checks now pass; merge when ready.

icelynjennings:
@vboulineau one of the lint actions failed to spin up: https://github.com/DataDog/helm-charts/pull/83/checks?check_run_id=1353327585

charts/datadog/values.yaml
```diff
@@ -976,6 +992,12 @@ clusterChecksRunner:
   additionalLabels: {}
     # key: "value"
 
+  # clusterChecksRunner.securityContext -- Allows you to overwrite the default PodSecurityContext on the clusterchecks pods.
+  securityContext: {}
+
+  # clusterChecksRunner.containerSecurityContext -- Allows you to overwrite the default container SecurityContext on the clustercheck containers.
+  containerSecurityContext: {}
```
vboulineau, line 999: I'm not sure the `containerSecurityContext` fields are useful as we only have 1 container / pod ATM, it may add confusion. We will add them if we need to when we put multiple containers in these pods.

charts/datadog/values.yaml
```diff
@@ -343,6 +343,12 @@ clusterAgent:
     pullSecrets: []
     #   - name: "<REG_SECRET>"
 
+  # clusterAgent.securityContext -- Allows you to overwrite the default PodSecurityContext on the cluster-agent pods.
+  securityContext: {}
+
+  # clusterAgent.containerSecurityContext -- Allows you to overwrite the default container SecurityContext for the cluster-agent.
+  containerSecurityContext: {}
```
vboulineau, line 350: Same for cluster agent

charts/datadog/values.yaml
icelynjennings: Though my need is strictly to set capabilities on the `system-probe` container, I've added these to resolve https://github.com/helm/charts/pull/17274#issuecomment-538536277 as discussed between @coreypobrien and @CharlyF when the conversation had originally spun from the `system-probe` container issue.

vboulineau: Yeah I agree with this. So having the `containerSecurityContext` in all containers is fine, though currently in `clusterAgent` and `clusterChecksRunner`, as we have 1 container/pod, so I believe we should only keep `securityContext` there ATM.

icelynjennings: Removed

charts/datadog/values.yaml
icelynjennings: Removed

icelynjennings:
Squashed, removed `containerSecurityContext` fields proposed by original PR @vboulineau

vboulineau (approved):

PR #93, opened 2020-11-07
Makes handling of useHostPID consistent

yafernandes:
The enabling `securityAgent.compliance` but not `useHostPID`, it created an inconsistency where it would be enabled in the daemonset but not enabled in the SCC for OpenShift.  The result was failed deployment in OpenShift.

Tested with OpenShift 4 on Azure (ARO).


vboulineau:
Hi, @yafernandes,

Thanks for your contribution. Looks LGTM to me.
However for the CI to pass, you'll need to bump the version in `charts/datadog/Chart.yaml` and update the documentation by running `./.github/helm-docs.sh`

PR #100, opened 2020-11-19
Change default values to give best experience OOTB

vboulineau:
  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -1,6 +1,6 @@
 {{- define "container-agent" -}}
 - name: agent
-  image: "{{ .Values.agents.image.repository }}:{{ .Values.agents.image.tag }}"
+  image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   command: ["agent", "run"]
   {{- if .Values.agents.containers.agent.securityContext }}
--- a/charts/datadog/templates/_helpers.tpl
+++ b/charts/datadog/templates/_helpers.tpl
@@ -197,3 +197,14 @@ true
 false
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return a remote image path based on `.Values` (passed as root) and `.` (any `.image` from `.Values` passed as parameter)
+*/}}
+{{- define "image-path" -}}
+{{- if .image.repository -}}
+{{- .image.repository -}}:{{ .image.tag }}
+{{- else -}}
+{{ .root.registry }}/{{ .image.name }}:{{ .image.tag }}
+{{- end -}}
+{{- end -}}
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -11,6 +11,13 @@ fullnameOverride:  # ""
 # targetSystem -- Target OS for this deployment (possible values: linux, windows)
 targetSystem: "linux"
 
+# registry -- Registry to use for all Agent images (default gcr.io)
+## Currently we offer Datadog Agent images on:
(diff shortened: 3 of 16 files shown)


charts/datadog/values.yaml
```diff
@@ -75,7 +75,7 @@ datadog:
   ## Autodiscovery via Kube Service annotations is automatically enabled
   clusterChecks:
     # datadog.clusterChecks.enabled -- Enable the Cluster Checks feature on both the cluster-agents and the daemonset
-    enabled: false
```
vboulineau, line 78: Should we activate CLC by default?

charts/datadog/values.yaml
```diff
@@ -141,14 +141,14 @@ datadog:
 
     # datadog.dogstatsd.nonLocalTraffic -- Enable this to make each node accept non-local statsd traffic
     ## ref: https://github.com/DataDog/docker-dd-agent#environment-variables
-    nonLocalTraffic: false
+    nonLocalTraffic: true
```
clamoriniere, line 144: this is set to false on purpose for security reason: mainly because if a node is not properly protected by a firewall, anyone can push traces to the agent.

charts/datadog/values.yaml
```diff
@@ -377,7 +376,7 @@ clusterAgent:
 
     # clusterAgent.metricsProvider.useDatadogMetrics -- Enable usage of DatadogMetric CRD to autoscale on arbitrary Datadog queries
     ## NOTE: You need to install the `DatadogMetric` CRD before
-    useDatadogMetrics: false
+    useDatadogMetrics: true
```
clamoriniere, line 379: I think it is ok, but we need to document that we only supported kubernetes > 1.10 (to support crds)

charts/datadog/values.yaml
vboulineau: Not sure as `useHostPort: false` by default

charts/datadog/values.yaml
vboulineau: So we agree to drop Kube < 1.10 (I think ATM we support 1.8?)

charts/datadog/values.yaml
clamoriniere: no. we will activate only the cluster-agent by default

clamoriniere:
@vboulineau 
any reason why this PR is still a draft pr?

vboulineau:
> @vboulineau
> any reason why this PR is still a draft pr?

Not anymore :)

clamoriniere (commented):
/LGTM

could you also update the default agent tag?

charts/datadog/templates/_container-agent.yaml
```diff
@@ -45,7 +45,7 @@
       value: {{ tpl (.Values.datadog.dogstatsd.tags | join " " | quote) . }}
     {{- end }}
     {{- if not .Values.clusterAgent.enabled }}
-    {{- if .Values.datadog.leaderElection }}
+    {{- if and (not .Values.clusterAgent.enabled) .Values.datadog.leaderElection }}
```
hkaj, line 48: is this change needed? The line directly above already ensures the `not .Values.clusterAgent.enabled` condition

vboulineau: Indeed, good catch, I must have missed it! Fixed

xornivore (approved):

charts/datadog/templates/_helpers.tpl
```diff
@@ -197,3 +197,14 @@ true
 false
 {{- end -}}
 {{- end -}}
+
+{{/*
+Return true if the system-probe container should be created.
```
hkaj, line 202: copy pasta? :D

vboulineau: Credo di sì :D

PR #104, opened 2020-11-23
[Private-Synthetics] Conditionally create configMap and docs

### LINKED ISSUE #103: [Private-Synthetics] Conditionally create config map
While installing the private synthetics chart, I noticed that I was forced to create a config map with the configuration. Since this contains secret information, I want to create this out of band (or using another tool that handles secrets). It would be nice if the chart allowed me to conditionally create the configMap (or potentially used a secrets so I could use external secrets for this).

johnkost:
This PR allows specifying whether the configMap should be created and allows specifying the name. This allows the configMap to be created out of band and passing the name of the configMap in to the chart. 

  - fixes #103 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-private-location/templates/_helpers.tpl
+++ b/charts/synthetics-private-location/templates/_helpers.tpl
@@ -64,6 +64,6 @@ Create the name of the service account to use
 {{/*
 Determine name of the configmap
 */}}
-{{- define "synthetics-privatet-location.configMapName" -}}
+{{- define "synthetics-private-location.configMapName" -}}
 {{- default ( printf "%s-config" (include "synthetics-private-location.fullname" .)) .Values.configMap.name }}
 {{- end }}
--- a/charts/synthetics-private-location/templates/_helpers.tpl
+++ b/charts/synthetics-private-location/templates/_helpers.tpl
@@ -60,3 +60,10 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Determine name of the configmap
+*/}}
+{{- define "synthetics-privatet-location.configMapName" -}}
+{{- default ( printf "%s-config" (include "synthetics-private-location.fullname" .)) .Values.configMap.name }}
+{{- end }}
--- a/charts/synthetics-private-location/templates/deployment.yaml
+++ b/charts/synthetics-private-location/templates/deployment.yaml
@@ -61,4 +61,4 @@ spec:
       volumes:
       - name: worker-config
         configMap:
-          name: {{ include "synthetics-private-location.fullname" . }}-config
+          name: {{ include "synthetics-privatet-location.configMapName" . }}
--- a/charts/synthetics-private-location/templates/deployment.yaml
+++ b/charts/synthetics-private-location/templates/deployment.yaml
@@ -61,4 +61,4 @@ spec:
       volumes:
       - name: worker-config
         configMap:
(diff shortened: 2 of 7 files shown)


charts/synthetics-private-location/templates/deployment.yaml
```diff
@@ -61,4 +61,4 @@ spec:
       volumes:
       - name: worker-config
         configMap:
-          name: {{ include "synthetics-private-location.fullname" . }}-config
+          name: {{ include "synthetics-private-location.configMapName" . }}
```
clamoriniere, line 64: I think the volume should to be added only if `.Values.configMap.enabled`

charts/synthetics-private-location/templates/_helpers.tpl
```diff
@@ -60,3 +60,10 @@ Create the name of the service account to use
 {{- default "default" .Values.serviceAccount.name }}
 {{- end }}
 {{- end }}
+
+{{/*
+Determine name of the configmap
+*/}}
+{{- define "synthetics-private-location.configMapName" -}}
+{{- default ( printf "%s-config" (include "synthetics-private-location.fullname" .)) .Values.configMap.name }}
+{{- end }}
```
clamoriniere, line 69:
```suggestion
{{- end }}

```

charts/synthetics-private-location/templates/deployment.yaml
johnkost: In my use case, I need the volume added but will be creating it out of band.

johnkost: I could add an extra volumes/volumeMounts section to the chart and only conditionally add the configMap that was created when enabled?

clamoriniere: I think we can simplify the logic by removing `configMap.enabled` parameter and test `{{- if .Values.configMap.name == "" }}` to create or configmap in the file `charts/synthetics-private-location/templates/configmap.yaml`

WDYT ?

johnkost: 👍 Let me make that change

johnkost: Thinking about this more, I think we should keep the `enabled` parameter. I see the following 2 use cases:
1. As currently written where the configMap is created and managed by the Helm chart
2. ConfigMap is managed out of band and used by the synthetics.

For option 1, we probably don't want to specify the name (allow the defaults). For option 2, we may want to specify the name or use the default. Based off this, I'd say the name only matters when providing your own config map so we shouldn't use it for the default case.

clamoriniere: I agree with the 2 use cases that you mention but:
1. the current generated configMap name is generated with the release name.
2. If someone provide `.values.configMap.name` it means that the ConfigMap is managed outside the chart. So we dont really need to have the `enabled` field.

Maybe the issue is only how we name this new field. WDYT if we renamed it: `externalConfiguration.configMap.name`

```yaml
externalConfiguration:
  configMap:
    # externalConfiguration.configMap.name -- ConfigMap name containing `synthetics-check-runner.json` configuration file. if set, it overrides the `configFile` parameter.
    name: ""
```

johnkost: That makes it clear it's external. I'll make that change

loutPhilipps:
👋🏼 This should have been covered by https://github.com/DataDog/helm-charts/pull/682, closing the Pull Request

PR #122, opened 2020-12-10
Non k8s .yaml files should be renamed like _helpers.tpl file

anthony-pastor:
Those .yaml files are not templates and should be considered like _helpers files.
Any document without metadata.name entry isn't a valid Kubernetes resource, and therefore should be considered like the _helpers.tpl file.

Without this _helm lint_ will drop an error for any of those files:
 _[ERROR] templates/containers-common-env.yaml: object name does not conform to Kubernetes naming requirements: "": invalid metadata name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])+$ and the length must not longer than 253_


ahmed-mez:
Hi @anthony-pastor, thank you for the contribution. Could you rebase your branch and resolve the conflict? thanks!

anthony-pastor:
@ahmed-mez I recreated it here: https://github.com/DataDog/helm-charts/pull/135

PR #135, opened 2021-01-08
Those .yaml files are not templates and should be considered as _helpers

anthony-pastor:
Those .yaml files are not templates and should be considered like _helpers files.
Any document without metadata.name entry isn't a valid Kubernetes resource, and therefore should be considered like the _helpers.tpl file.

Without this helm lint will drop an error for any of those files:
[ERROR] templates/containers-common-env.yaml: object name does not conform to Kubernetes naming requirements: "": invalid metadata name, must match regex ^(([A-Za-z0-9][-A-Za-z0-9.]*)?[A-Za-z0-9])+$ and the length must not longer than 253_


ahmed-mez:
hey @anthony-pastor - the changes look good, could you bump the chart version and add a release note? thanks!

anthony-pastor:
Good to go @ahmed-mez

vboulineau (approved):

anthony-pastor:
@vboulineau is this log expected?
https://github.com/DataDog/helm-charts/pull/135/checks?check_run_id=1681509535#step:4:4719

vboulineau:
> @vboulineau is this log expected?
> https://github.com/DataDog/helm-charts/pull/135/checks?check_run_id=1681509535#step:4:4719

Yes, the install CI does not have `system-probe` installed by default.

PR #137, opened 2021-01-11
Improve CI and add /tmp as an empty dir to prepare for VOLUME removal in Dockerfile

vboulineau:

Improve Helm CI
Mount `/tmp` as an empty dir in all Agent containers

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -93,6 +93,9 @@
       subPath: install_info
       mountPath: /etc/datadog-agent/install_info
       readOnly: true
+    - name: tmpdir
+      mountPath: /tmp
+      readOnly: false
     {{- end }}
     - name: config
       mountPath: {{ template "datadog.confPath" . }}
--- a/charts/datadog/templates/cluster-agent-rbac.yaml
+++ b/charts/datadog/templates/cluster-agent-rbac.yaml
@@ -233,7 +233,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name | quote }}
     app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
     app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-  name: {{ template "datadog.fullname" . }}-cluster-agent:system:auth-delegator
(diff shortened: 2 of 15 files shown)


charts/datadog/templates/_container-agent.yaml
```diff
@@ -93,6 +93,9 @@
       subPath: install_info
       mountPath: /etc/datadog-agent/install_info
       readOnly: true
+    - name: tmpdir
```
clamoriniere, line 96: should we shared the same empty dir across containers?

clamoriniere: I don't see a specific issue except if 2 agents process generate the same file, like a flare.

vboulineau: Indeed, it theory it could cause some issues. though creating one volume (emptyDir) / container may be a bit complex in terms of setup.

clamoriniere: I think it is ok to make it like you did because when the agent is installed on the host, the agents also shared the same `/tmp` folder

clamoriniere (approved):

clamoriniere:
hi @vboulineau
Could you add in the Contribution guide the info that people need to run the `: .github/helm-docs.sh` script after bumping the `chart.yaml`

charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -233,7 +233,7 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name | quote }}
     app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
     app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
-  name: {{ template "datadog.fullname" . }}-cluster-agent:system:auth-delegator
+  name: {{ template "datadog.fullname" . }}-cluster-agent-system-auth-delegator
```
tonglil, line 236: You shouldn't arbitrarily rename kubernetes resources like this.

Without an easy way to clean things up for non-helm users, this leaves 2 CRBs floating around in clusters.. Also there was no note of this change, and this repo has no changelog. it's not a great experience for datadog users

PR #124, opened 2020-12-11
remove the datadog-cluster-id configmap mount from process-agent

hkaj:
#### What this PR does / why we need it:

Remove the cluster-id configmap mount for process-agent. This is not needed starting with agent 7.25 and dca 1.11 since we use a cluster agent api endpoint for this now.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Special notes for your reviewer:

WIP

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #138, opened 2021-01-12
Enable the orchestrator explorer by default

FlorianVeaux:
This PR enables the orchestrator explorer by default.
#### Special notes:

The important point of that PR is that it shouldn't introduce major incompatibilities for users, upgrading the helm chart shouldn't make the agent fail. As many variables comes into play, let's list the various scenarios.

Pre-requisites:
- Up to agent 7.24.x, the pod check on the process-agent requires that the `DD_ORCHESTRATOR_CLUSTER_ID` env variable is set. This is taken care of by our helm chart here: https://github.com/DataDog/helm-charts/blob/master/charts/datadog/templates/_container-process-agent.yaml#L38-L44
- With the next agent release (7.25.x), the pod check is now able to query the cluster agent to get the cluster_id. As long as the cluster agent is recent enough (future release 1.11), agent 7.25.x is able to live without the `DD_ORCHESTRATOR_CLUSTER_ID` env var.
- In the short-term, the helm chart will stop populating this environment variable: https://github.com/DataDog/helm-charts/pull/124 so the current PR should be able to work with this future change.

Decision tree:
Enable orchestration

In this diagram:

- `Config #1`: is when the user is explicitly defining both `clusterAgent.enabled` and `orchestratorExplorer.enabled`. In that case nothing happens with this PR and the values used are the ones provided by the user. Note that this can lead to enabling the orchestrator without the clusterAgent but that was already the case before.
- `Config #2`: is when the user explicitly sets `clusterAgent.enabled` to false and `orchestratorExplorer.enabled` is enabled (either via explicit configuration or via the new default). This is a new scenario where the orchestrator is enabled but can't work because `clusterAgent` is disabled. In that specific scenario, the helm chart disables the orchestrator explorer and logs a warning. Otherwise, the process-agent container would fail to start
- `Config #3`: is any other config basically.
- The red box `pod check is failing` means that on the process-agent container, the pod check will fail to get the cluster id and won't be able to run. In the end the orchestrator explorer page won't work at all for pods but the process-agent will continue to run.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.8.0
+
+* Enable the orchestrator explorer by default.
+
 ## 2.7.0
 
 * Changes default values to activate a maximum of built-in features to ease configuration.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.7.0
+version: 2.8.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.7.0](https://img.shields.io/badge/Version-2.7.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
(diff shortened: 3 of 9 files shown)


vboulineau:
I'd expect this to go on top of all default review we have in:
https://github.com/DataDog/helm-charts/pull/100

hkaj (approved):
this is great! Let's merge as soon as we get approval from container-integrations :)

hkaj (approved):

vboulineau (approved):

clamoriniere (approved):

PR #146, opened 2021-01-21
Add configuration for runtime compiler in system-probe

brycekahle:

Adds configuration options for enabling/configuring the runtime compiler in the system-probe, due to be released in 7.27.0.

See https://github.com/DataDog/datadog-agent/pull/6978

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_daemonset-volumes-linux.yaml
+++ b/charts/datadog/templates/_daemonset-volumes-linux.yaml
@@ -51,13 +51,19 @@
   name: debugfs
 - name: sysprobe-socket-dir
   emptyDir: {}
-{{- if or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill }}
+{{- if or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill .Values.datadog.systemProbe.enableRuntimeCompiler }}
 - hostPath:
     path: /lib/modules
   name: modules
 - hostPath:
     path: /usr/src
   name: src
+{{- end }}
+{{- if .Values.datadog.systemProbe.enableRuntimeCompiler }}
+- hostPath:
+    path: {{ .Values.datadog.systemProbe.runtimeCompilerOutputDir }}
+    type: DirectoryOrCreate
+  name: runtime-compiler-output-dir
 {{- end }}
 {{- end }}
 {{- if or .Values.datadog.processAgent.enabled (eq (include "should-enable-system-probe" .) "true") (eq (include "should-enable-security-agent" .) "true") }}
--- a/charts/datadog/templates/system-probe-configmap.yaml
+++ b/charts/datadog/templates/system-probe-configmap.yaml
@@ -36,6 +36,7 @@ data:
       collect_dns_stats: {{ $.Values.datadog.systemProbe.collectDNSStats }}
       max_tracked_connections: {{ $.Values.datadog.systemProbe.maxTrackedConnections }}
       conntrack_max_state_size: {{ $.Values.datadog.systemProbe.conntrackMaxStateSize }}
+      enable_runtime_compiler: {{ $.Values.datadog.systemProbe.enableRuntimeCompiler }}
     network_config:
       enabled: {{ $.Values.datadog.networkMonitoring.enabled }}
     runtime_security_config:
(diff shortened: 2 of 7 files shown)


charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -51,6 +51,11 @@
     path: /usr/src
   name: src
 {{- end }}
+{{- if .Values.datadog.systemProbe.enableRuntimeCompiler }}
+- hostPath:
+    path: {{ .Values.datadog.systemProbe.runtimeCompilerOutputDir }}
+  name: runtime-compiler-output-dir
```
clamoriniere, line 57: it is better to add also the `type: DirectoryOrCreate` if the folder doesn't exist yet

```suggestion
- hostPath:
    path: {{ .Values.datadog.systemProbe.runtimeCompilerOutputDir }}
    type: DirectoryOrCreate
  name: runtime-compiler-output-dir
```

charts/datadog/values.yaml
```diff
@@ -282,6 +282,12 @@ datadog:
     # datadog.systemProbe.collectDNSStats -- Enable DNS stat collection
     collectDNSStats: false
 
+    # datadog.systemProbe.enableRuntimeCompiler - Enable the runtime compiler for eBPF probes
+    enableRuntimeCompiler: false
+
+    # datadog.systemProbe.runtimeCompilerOutputDir - Specify an output directory for the compiled assets from the eBPF runtime compiler
+    runtimeCompilerOutputDir: /var/tmp/datadog-agent/system-probe/build
```
clamoriniere, line 289: I'm wondering why this path and not: `/var/run/datadog/system-probe/build` ?

doesn't /var/tmp` have some specific cleanup policy?

clamoriniere: ok today I learn about `/var/tmp` :) 
https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch05s15.html

clamoriniere (approved):

stb2132 (approved):

charts/datadog/values.yaml
```diff
@@ -327,6 +327,11 @@ datadog:
     # datadog.systemProbe.conntrackMaxStateSize -- the maximum size of the userspace conntrack cache
     conntrackMaxStateSize: 131072  # 2 * maxTrackedConnections by default, per  https://github.com/DataDog/datadog-agent/blob/d1c5de31e1bba72dfac459aed5ff9562c3fdcc20/pkg/process/config/config.go#L229
 
+    # datadog.systemProbe.enableRuntimeCompiler -- Enable the runtime compiler for eBPF probes
+    enableRuntimeCompiler: false
```
leeavital, line 331: Is it worth enforcing a minimum agent version for this?

We do something similar here: https://github.com/DataDog/helm-charts/blob/master/charts/datadog/templates/system-probe-configmap.yaml#L7

brycekahle: It feels unnecessary to me to add version checking for every new config option. I understand the `networkMonitoring` one because that determines whether the system-probe even runs or not. This (for now) is still an opt-in feature.

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -36,6 +36,7 @@ data:
       collect_dns_stats: {{ $.Values.datadog.systemProbe.collectDNSStats }}
       max_tracked_connections: {{ $.Values.datadog.systemProbe.maxTrackedConnections }}
       conntrack_max_state_size: {{ $.Values.datadog.systemProbe.conntrackMaxStateSize }}
+      enable_runtime_compiler: {{ $.Values.datadog.systemProbe.enableRuntimeCompiler }}
```
leeavital, line 39: This could cause problems when the default value in helm disagrees with the default value in the agent.

1. user uses a helm chart with default enableRuntimeCompiler=false
2. user installs a future agent version with enableRuntimeCompiler=true by default
3. because of using an old helm chart, the user will incorrectly be using offset guessing.

brycekahle: What are you suggesting to fix that?

brycekahle: We decided to leave it as-is.

brycekahle:
@clamoriniere This is ready for a merge when you get a minute.

PR #161, opened 2021-02-08
[datadog] Support prometheus autodiscovery

ahmed-mez:

Configure/enable prometheus AD. Requires Datadog Agent 7/6.26+ and Datadog Cluster Agent 1.11+.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -216,6 +216,12 @@ spec:
           - name: DD_COMPLIANCE_CONFIG_CHECK_INTERVAL
             value: {{ .Values.datadog.securityAgent.compliance.checkInterval | quote }}
           {{- end }}
+          {{- if .Values.datadog.prometheusScrape.enabled }}
+          - name: DD_PROMETHEUS_SCRAPE_ENABLED
+            value: "true"
+          - name: DD_PROMETHEUS_SCRAPE_SERVICE_ENDPOINTS
+            value: {{ .Values.datadog.prometheusScrape.serviceEndpoints | quote }}
+          {{- end }}
           {{- end }}
 {{- if .Values.clusterAgent.env }}
 {{ toYaml .Values.clusterAgent.env | indent 10 }}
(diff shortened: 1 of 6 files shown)


charts/datadog/templates/cluster-agent-config-configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.clusterAgent.datadog_cluster_yaml }}
+{{- if or (.Values.clusterAgent.datadog_cluster_yaml) (eq (include "should-add-additional-prometheus-configs" .) "true") }}
```
vboulineau, line 1: As it's repeated many times, I think the helper should rather be "should-use-configmap" or something

ahmed-mez: yes good point thanks, we'll need two separate helpers though (for the DCA and the Agent)

ahmed-mez: done

clamoriniere (approved):

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -216,6 +216,12 @@ spec:
           - name: DD_COMPLIANCE_CONFIG_CHECK_INTERVAL
             value: {{ .Values.datadog.securityAgent.compliance.checkInterval | quote }}
           {{- end }}
+          {{- if .Values.datadog.prometheusScrape.enabled }}
```
tomislater, line 219: Hmm, I think that this `if` block should be put after the last `end` block.

At this moment, when I enable `prometheusScrape` nothing happens. I see that this `if` is in another block (`{{- if eq  (include "should-enable-security-agent" .) "true" }}`).

tomislater: I can create PR for this.

ahmed-mez: Indeed, I fixed it part of this PR https://github.com/DataDog/helm-charts/pull/201 actually which will be merged when Agent 7.27 is released.
But this fix itself can be fixed and merged immediately without waiting for the Agent release, so feel free to submit a fix 😄  Thanks!

tomislater: Hey 🖐

Here is the fix: https://github.com/DataDog/helm-charts/pull/202

PR #164, opened 2021-02-17
Fix `networkMonitoring.enabled` behavior when set to `false`

p-lambert:

Currently it is not possible to disable the network monitoring (while leaving system-probe enabled) by setting `datadog.networkMonitoring.enabled` to `false`. This PR fixes that.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.8.6](https://img.shields.io/badge/Version-2.8.6-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 2.9.0](https://img.shields.io/badge/Version-2.9.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 [Datadog](https://www.datadoghq.com/) is a hosted infrastructure monitoring platform. This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally depends on the [kube-state-metrics chart](https://github.com/kubernetes/charts/tree/master/stable/kube-state-metrics). For more information about monitoring Kubernetes with Datadog, please refer to the [Datadog documentation website](https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/).
 
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -483,7 +483,7 @@ helm install --name <RELEASE_NAME> \
 | datadog.logs.containerCollectAll | bool | `false` | Enable this to allow log collection for all containers |
 | datadog.logs.containerCollectUsingFiles | bool | `true` | Collect logs from files in /var/log/pods instead of using container runtime API |
 | datadog.logs.enabled | bool | `false` | Enables this to activate Datadog Agent log collection |
-| datadog.networkMonitoring.enabled | bool | `nil` | Enable network performance monitoring |
+| datadog.networkMonitoring.enabled | bool | `false` | Enable network performance monitoring |
 | datadog.networkPolicy.cilium.dnsSelector | object | kube-dns in namespace kube-system | Cilium selector of the DNS server entity |
 | datadog.networkPolicy.create | bool | `false` | If true, create NetworkPolicy for all the components |
 | datadog.networkPolicy.flavor | string | `"kubernetes"` | Flavor of the network policy to use. Can be: * kubernetes for networking.k8s.io/v1/NetworkPolicy * cilium     for cilium.io/v2/CiliumNetworkPolicy |
@@ -532,7 +532,6 @@ Some options above are not working/not available on Windows, here is the list of
 | `datadog.dogstatsd.useSocketVolume`      | Unix sockets not supported on Windows            |
 | `datadog.dogstatsd.socketPath`           | Unix sockets not supported on Windows            |
 | `datadog.processAgent.processCollection` | Unable to access host/other containers processes |
-| `datadog.systemProbe.enabled`            | System probe is not available for Windows        |
 | `datadog.systemProbe.seccomp`            | System probe is not available for Windows        |
 | `datadog.systemProbe.seccompRoot`        | System probe is not available for Windows        |
 | `datadog.systemProbe.debugPort`          | System probe is not available for Windows        |
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.8.4](https://img.shields.io/badge/Version-2.8.4-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 2.8.5](https://img.shields.io/badge/Version-2.8.5-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 [Datadog](https://www.datadoghq.com/) is a hosted infrastructure monitoring platform. This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally depends on the [kube-state-metrics chart](https://github.com/kubernetes/charts/tree/master/stable/kube-state-metrics). For more information about monitoring Kubernetes with Datadog, please refer to the [Datadog documentation website](https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/).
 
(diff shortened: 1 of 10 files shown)


charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -1,3 +1,14 @@
+{{- if .Values.datadog.networkMonitoring.enabled }}
+{{- $version := (.Values.agents.image.tag | toString | trimSuffix "-jmx") }}
+{{- $length := len (split "." $version ) -}}
+{{- if (gt $length 1) }}
+{{- if not (semverCompare "^6.23.0-0 || ^7.23.0-0" $version) -}}
```
leeavital, line 5: This should respect `agents.image.doNotCheckTag`

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -1,3 +1,14 @@
...
+{{- fail "datadog.networkMonitoring.enabled requires agent >= 7.23.0" }}
+{{- end }}
+{{- end }}
+{{- end }}
```
clamoriniere, line 9: could you use [doNotCheckTag](https://github.com/DataDog/helm-charts/blob/f003f0de258e7670699ed4bb73629f585a9244cf/charts/datadog/values.yaml#L597) here, like it is used [here](https://github.com/DataDog/helm-charts/blob/fa68a6f473c5521a8830db60899af6b2e8ecbe87/charts/datadog/templates/_helpers.tpl#L4)

We have introduced this setting to allow users that build custom agent images which doesn't use SemVer tag to bypass the version check.

leeavital (commented):
LGTM, don't forget to bump the chart version and changelog

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -21,10 +32,12 @@ data:
       enable_tcp_queue_length: {{ $.Values.datadog.systemProbe.enableTCPQueueLength }}
       enable_oom_kill: {{ $.Values.datadog.systemProbe.enableOOMKill }}
       collect_dns_stats: {{ $.Values.datadog.systemProbe.collectDNSStats }}
-    {{- if $.Values.datadog.networkMonitoring.enabled }}
     network_config:
-      enabled: true
-    {{- end }}
+      {{- if (eq ($.Values.datadog.networkMonitoring.enabled | toString) "<nil>") }}
```
clamoriniere, line 36: `datadog.networkMonitoring.enabled` from the `values.yaml` file should be `false` or `true`, so why checking nil ?

p-lambert: Long story short, we need that sentinel value for backwards compatibility. We want to meet the following conditions:

* systemProbe.enabled -> defaults to false
* networkConfig.enabled -> defaults to false
* systemProbe.enabled= false, networkConfig.enabled = not set -> system probe is off
* systemProbe.enabled=false, networkConfig.enabled = false -> system probe is off
* systemProbe.enabled=false, networkConfig.enabled=true -> system probe is on, but only the network module
* systemProbe.enabled = true, networkConfig.enabled = not set -> system probe is on, with network module enabled (for back compat reasons)
* systemProbe.enabled = true, networkConfig.enabled = false (explicitly) -> system probe is on, with network module disabled
* systemProbe.enabled = true, networkConfig.enabled = true (explicitly) -> system probe is on, with network module enabled

p-lambert: The issue is that the auto-generated docs now want this variable to be documented as `string` which might be misleading to customers :(
```
+| datadog.networkMonitoring.enabled | string | `nil` | Enable network performance monitoring |
```

p-lambert: nevermind, found a workaround

charts/datadog/README.md
```diff
@@ -483,7 +483,7 @@ helm install --name <RELEASE_NAME> \
 | datadog.logs.containerCollectAll | bool | `false` | Enable this to allow log collection for all containers |
 | datadog.logs.containerCollectUsingFiles | bool | `true` | Collect logs from files in /var/log/pods instead of using container runtime API |
 | datadog.logs.enabled | bool | `false` | Enables this to activate Datadog Agent log collection |
-| datadog.networkMonitoring.enabled | bool | `false` | Enable network performance monitoring |
+| datadog.networkMonitoring.enabled | bool | `nil` | Enable network performance monitoring |
```
leeavital, line 486: Let's make explcitit that systemProbe.enabled is used if `nil`

leeavital (approved):

charts/datadog/README.md
clamoriniere: sorry but a bool have only two value: `true` and `false`.
Can we get ride of these three states?

charts/datadog/templates/system-probe-configmap.yaml
clamoriniere: you need also to consider the `securityagent.runtime` and `[oom,tcpqueue] probe` in systemprobe enabling process.

p-lambert: I'm not sure what you mean by your comment. System-probe enabling is already being enabled by these flags.

charts/datadog/README.md
p-lambert: I explained the reason here: https://github.com/DataDog/helm-charts/pull/164#discussion_r577845106
Without the `nil` value we can't preserve backwards compatibility and customers would inadvertently disable NPM monitoring.

charts/datadog/templates/system-probe-configmap.yaml
clamoriniere: I think having this "nil/null" value will confuse users.

In previous a PR, we deprecate "systemProbe.enabled" option. The idea was to not consider the `systemProbe` as a feature, but just an implementation details.

And so instead of a global `systemProbe.enabled`, we have for each feature/product a boolean to enabled the corresponding module in `systemProbe`. 

From your previous use cases description, I think the only case which is problematic is `systemProbe.enable=true` with `networkMonitoring.enabled=false`.
But we can avoid having this case by adding a test in the chart, that refuses the deployment, and return an error message that said: "networkMonitoring can't be disable if `systemProbe.enable=true`, please remove the "systemProbe.enable" option from your configuration."

WDYT?

p-lambert: That sounds reasonable to me. @leeavital any objections to this?

leeavital: Are you suggesting that:
* the defaults are `networkMonitoring.enabled=false` && `systemProbe.enabled=false`   
* `networkMonitoring.enabled=false` && `systemProbe.enabled=true` causes the chart to fail
?

I think that's fine for new users of the chart, but breaks all the customers who have been using the old form of config.  Generally, as a helm user, I expect upgrading a chart to "just work" and this will break that.

Perhaps I'm misunderstanding what you're proposing ?

clamoriniere: We advice users since several versions to avoid using systemProbe.enabled. 
So I think it ok to move forward and try to enforce proper configuration. Do make this change we can increment the minor version.

leeavital: It has been only a few months since we introduced the `networkMonitoring` field.     I don't have a good sense of how long is "right".    

If we already have precedent for breaking configurations (and it sounds like we do) to enforce config migrations, then I think your suggestion makes a lot of sense.

p-lambert: @clamoriniere I fully removed `systemProbe.enabled` option and did a minor version bump. Please give it a review let us know if that makes sense.

charts/datadog/templates/NOTES.txt
```diff
@@ -183,14 +183,7 @@ More information about this change: https://github.com/DataDog/helm-charts/pull/
...
 {{- if .Values.datadog.systemProbe.enabled }}
-
-#################################################################
-####               WARNING: Deprecation notice               ####
-#################################################################
-
-You are using datadog.systemProbe.enabled which has been deprecated in favor using one of networkMonitoring.enabled, systemProbe.enableTCPQueueLength, systemProbe.enableOOMKill, or securityAgent.runtime.enabled.
-More information about this change: https://github.com/DataDog/helm-charts/pull/101
-
+{{- fail "You are using datadog.systemProbe.enabled which has been superseded by networkMonitoring.enabled, systemProbe.enableTCPQueueLength, systemProbe.enableOOMKill, and securityAgent.runtime.enabled" }}
```
clamoriniere, line 186: could you add a sentence which explains this choice like.
```suggestion
{{- fail "You are using datadog.systemProbe.enabled which has been superseded by networkMonitoring.enabled, systemProbe.enableTCPQueueLength, systemProbe.enableOOMKill, and securityAgent.runtime.enabled. The new options provided a better features activation granularity." }}
```

clamoriniere (approved):
Just a small comment about the error message, else it looks good.
Thanks you for this PR.

PR #165, opened 2021-02-17
Make agents able to connect to an existing DCA

hkaj:
#### What this PR does / why we need it:

Make agents able to connect to an existing DCA

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Special notes for your reviewer:

the following `values.yaml` file can be use to instantiate `datadog` chart on windows nodes.

```yaml
targetSystem: windows

existingClusterAgent:
  join: true
  serviceName: dd-datadog-cluster-agent
  tokenSecretName: dd-datadog-cluster-agent

datadog-crds:
  crds:
    # datadog-crds.crds.datadogMetrics -- Set to true to deploy the DatadogMetrics CRD
    datadogMetrics: false

# Disable kube-state-metrics deployement
datadog:
  kubeStateMetricsEnabled: false
```

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #189, opened 2021-03-10
When joining an existing DCA, do not let node agents take cluster checks

L3n41c:

#165 introduced the possibility for the node agents of an Helm release to join an “external” DCA (i.e. managed by another Helm release).
By doing so, the node agents were also taking part of the cluster check dispatch.
If the DCA is “external”, we should rather let the cluster checks be handled by the agents (CLC) deployed with the DCA.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -79,7 +79,7 @@
       value: {{ .Values.datadog.dogstatsd.socketPath | quote }}
     {{- end }}
     {{- if and (eq (include "cluster-agent-enabled" .) "true") .Values.datadog.clusterChecks.enabled }}
-    {{- if .Values.clusterChecksRunner.enabled }}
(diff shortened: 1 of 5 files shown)


clamoriniere (commented):
Thanks for opening this contribution.

I think we need to let the user decide if the cluster check can be schedule on the node agents.
we can add a new parameter `existingClusterAgent.clustercheckEnabled` with a default value set to `true` because in the helm chart `CLC` are not deployed by default. 
WDYT?

clamoriniere (approved):

charts/datadog/templates/_container-agent.yaml
```diff
@@ -79,7 +79,8 @@
       value: {{ .Values.datadog.dogstatsd.socketPath | quote }}
     {{- end }}
     {{- if and (eq (include "cluster-agent-enabled" .) "true") .Values.datadog.clusterChecks.enabled }}
-    {{- if .Values.clusterChecksRunner.enabled }}
+    {{- if or (and (not .Values.existingClusterAgent.join)     .Values.clusterChecksRunner.enabled)
+              (and      .Values.existingClusterAgent.join (not .Values.existingClusterAgent.clusterchecksEnabled)) }}
```
clamoriniere, line 83: I feel it need to be put on the same line to make the linter apply.

PR #209, opened 2021-03-29
[DCA] Use Role instead of ClusterRole for secrets RBAC

ahmed-mez:

Apply this change https://github.com/DataDog/datadog-agent/pull/7754

DCA 1.12 must be released before merging this PR

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-rbac.yaml
+++ b/charts/datadog/templates/cluster-agent-rbac.yaml
@@ -154,9 +154,6 @@ rules:
   resources:
   - mutatingwebhookconfigurations
   verbs: ["get", "list", "watch", "update", "create"]
-- apiGroups: [""]
-  resources: ["secrets"]
-  verbs: ["get", "list", "watch", "update", "create"]
 - apiGroups: ["batch"]
   resources: ["jobs", "cronjobs"]
   verbs: ["get"]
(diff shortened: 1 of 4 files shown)


charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -314,6 +311,42 @@ metadata:
...
+apiVersion: {{ template "rbac.apiVersion" . }}
+kind: RoleBinding
+metadata:
+  labels:
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    app.kubernetes.io/name: "{{ template "datadog.fullname" . }}"
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+  name: "{{ template "datadog.fullname" . }}-cluster-agent-rolebinding"
```
L3n41c, line 340: The `ClusterRoleBinding` is just named `{{ template "datadog.fullname" . }}-cluster-agent`. I.e. the name of the object is not suffixed by its type.
As this `RoleBinding` is a different kind, we shouldn’t have any name conflict.

So, we could use:
```suggestion
  name: "{{ template "datadog.fullname" . }}-cluster-agent"
```

L3n41c (approved):
Just left a small remark about the fact that we usually not suffix object names with their kind.

charts/datadog/templates/cluster-agent-rbac.yaml
ahmed-mez: we actually have a rolebinding with this name https://github.com/DataDog/helm-charts/blob/master/charts/datadog/templates/cluster-agent-rbac.yaml#L353

L3n41c (approved):
👍🏻

PR #211, opened 2021-04-05
Enable dns stats by default

mbakht:

We have enabled DNS stats collection by default everywhere. Updating the helm chart to be in sync with that.

  - fixes #

#### Checklist
- [ ] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -313,7 +313,7 @@ datadog:
     enableOOMKill: false
 
     # datadog.systemProbe.collectDNSStats -- Enable DNS stat collection
-    collectDNSStats: false
+    collectDNSStats: true
```
clamoriniere, line 316: does it means that SystemProbe is required by default?

mbakht: It is the latter. DNS stats will get collected only when `networkMonitoring.enable` is also `true`

clamoriniere: Thanks for the answer. 
If you can rebase you pull request, we will be able to merge it.

mbakht: Done

clamoriniere (approved):

PR #219, opened 2021-04-12
[datadog] DCA 1.12 + Support KSM core basic config

ahmed-mez:

- Bump the DCA version
- Support configuring the ksm core check in the cluster agent

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -77,6 +77,13 @@ datadog:
     # datadog.kubeStateMetricsNetworkPolicy.create -- If true, create a NetworkPolicy for kube state metrics
     create: false
 
+  kubeStateMetricsCore:
+    # datadog.kubeStateMetricsCore.enabled -- Enable the kubernetes_state_core check in the Cluster Agent (Requires Cluster Agent 1.12.0+)
+    enabled: false
+
+    # datadog.kubeStateMetricsCore.ignoreLegacyKSMCheck -- Disable the auto-configuration of lagacy kubernetes_state check (taken into account only when datadog.kubeStateMetricsCore.enabled is true)
+    ignoreLegacyKSMCheck: true
+
   ## Manage Cluster checks feature
   ## ref: https://docs.datadoghq.com/agent/autodiscovery/clusterchecks/
   ## Autodiscovery via Kube Service annotations is automatically enabled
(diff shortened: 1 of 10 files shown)


charts/datadog/values.yaml
```diff
@@ -77,6 +77,17 @@ datadog:
     # datadog.kubeStateMetricsNetworkPolicy.create -- If true, create a NetworkPolicy for kube state metrics
     create: false
 
+  # datadog.kubeStateMetricsCore -- Kube-State-Metrics core check configuration
```
vboulineau, line 80: Only leaf nodes are documented

charts/datadog/values.yaml
```diff
@@ -77,6 +77,13 @@ datadog:
     # datadog.kubeStateMetricsNetworkPolicy.create -- If true, create a NetworkPolicy for kube state metrics
     create: false
 
+  kubeStateMetricsCore:
+    # datadog.kubeStateMetricsCore.enabled -- Enable the kubernetes_state_core check in the Cluster Agent
+    enabled: false
+
+    # datadog.kubeStateMetricsCore.ignoreLegacyKSMCheck -- Disable the auto-configuration of lagacy kubernetes_state check (taken into account only when datadog.kubeStateMetricsCore.enabled is true)
+    ignoreLegacyKSMCheck: true
```
vboulineau, line 85: Is it necessary to offer this? Setting `datadog.kubeStateMetricsCore.enabled` to false if, for some reason, core does not work well, should be enough no?

ahmed-mez: `ignoreLegacyKSMCheck` is only taken into consideration if `kubeStateMetricsCore.enabled=true`. so yes disabling `kubeStateMetricsCore` is enough and guaranteed already.
`ignoreLegacyKSMCheck` can be useful though during the migration phase, users might be interested in enabling both checks at first

charts/datadog/values.yaml
```diff
@@ -77,6 +77,13 @@ datadog:
     # datadog.kubeStateMetricsNetworkPolicy.create -- If true, create a NetworkPolicy for kube state metrics
     create: false
 
+  kubeStateMetricsCore:
+    # datadog.kubeStateMetricsCore.enabled -- Enable the kubernetes_state_core check in the Cluster Agent
```
clamoriniere, line 81: could you add in comment: require cluster-agent >=1.12 version

ahmed-mez: Done!

clamoriniere (approved):

PR #224, opened 2021-04-26
[networks] Add missing syscalls to seccomp profile

p-lambert:

* Add missing syscalls to system-probe seccomp profile
* `getgroups`, `getpgrp` and`setpgid` were preventing us from `kubectl exec` to the system-probe container;
* `tgkill` seems to be used by some parts of the Go runtime;

#### Checklist
- [x] Chart Version bumped


clamoriniere (commented):
Hi @p-lambert 
Your change looks good.

Could you update the change-log, and also run the `.github/helm-docs.sh` script as explain in the PR checklist.
Thanks

clamoriniere (approved):

brycekahle:
Could we comment the syscalls that were added only for `kubectl exec`? I could see folks not wanting that, and it blurs the line of what the system-probe needs versus k8s.

p-lambert:
> Could we comment the syscalls that were added only for `kubectl exec`? I could see folks not wanting that, and it blurs the line of what the system-probe needs versus k8s.

@brycekahle Do you mean commenting out and leaving the seccomp profile effectively as it is, or simply grouping them in the list like

```
# This syscalls are necessary for kubectl exec
"setpgid",
...
```

brycekahle:
> Do you mean commenting out and leaving the seccomp profile effectively as it is, or simply grouping them in the list

The latter, adding a comment explaining.

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -210,14 +210,19 @@ data:
...
             "umask",
             "uname",
             "unlink",
             "unlinkat",
             "wait4",
             "waitid",
             "waitpid",
-            "write"
+            "write",
+            # The following system calls are only necessary for kubectl exec
```
loganfsmyth, line 222: I think this comment is screwing things up, because it is inside a JSON config file, and JSON doesn't allow comments.

PR #240, opened 2021-05-06
Added option for specifying the Config Map with config

yafernandes:
The goal is to make the json config external to the Chart.  It will allow the user to manage the Config Map with the json config independently to the Helm Chart.


charts/synthetics-private-location/templates/deployment.yaml
```diff
@@ -61,4 +61,8 @@ spec:
       volumes:
       - name: worker-config
         configMap:
+        {{- if .Values.configConfigMap }}
+          name: {{ .Values.configConfigMap | quote }}
+        {{- else }}
           name: {{ include "synthetics-private-location.fullname" . }}-config
+        {{- end }}
```
clamoriniere, line 68:
```suggestion
        {{- end }}

```

clamoriniere (commented):
The feature seems legit.
Could you add a `NOTES.txt` file that return message (+ an error) if both `configConfigMap` and `configFile` are set. to explain to the user that it is not possible.

you can find an example [here](https://github.com/DataDog/helm-charts/blob/master/charts/datadog/templates/NOTES.txt)

yafernandes:
Should I also update the image to 1.10?

clamoriniere (approved):
/lgtm 👍

PR #265, opened 2021-05-28
Add warning about containerInclude usage.

ludaires:

Provide warnings when using only  ```container_include``` . It does nothing since the default behavior is to include everything. The expectation would be to exclude everything else if include is specified.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/NOTES.txt
+++ b/charts/datadog/templates/NOTES.txt
@@ -255,3 +255,27 @@ helm install ksm https://charts.helm.sh/stable/packages/kube-state-metrics-2.9.1
 You are using datadog.kubeStateMetricsCore.enabled but you disabled the cluster agent. This configuration is unsupported and the kube-state-metrics core check can't be configured.
 To enable it please set clusterAgent.enabled to 'true'.
 {{- end }}
+
+{{- $hasContainerInclude := false }}
+{{- range $key := .Values.datadog.env }}
+  {{- if eq $key.name "DD_CONTAINER_INCLUDE" }}
+    {{- $hasContainerInclude = true }}
+  {{- end }}
+{{- end }}
+
+{{- $hasContainerExclude := false }}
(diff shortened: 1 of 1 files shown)


charts/datadog/templates/NOTES.txt
```diff
@@ -255,3 +255,27 @@ helm install ksm https://charts.helm.sh/stable/packages/kube-state-metrics-2.9.1
 You are using datadog.kubeStateMetricsCore.enabled but you disabled the cluster agent. This configuration is unsupported and the kube-state-metrics core check can't be configured.
 To enable it please set clusterAgent.enabled to 'true'.
 {{- end }}
+
+{{- $hasContainerInclude := false }}
+{{- range $key := .Values.datadog.env }}
```
celenechang, line 260: maybe we should also check for the presence of `.Values.datadog.containerInclude` as well:
https://github.com/DataDog/helm-charts/blob/5056593ac4f07a463f48378f6d90a703a0279646/charts/datadog/templates/_containers-common-env.yaml#L67

charts/datadog/templates/NOTES.txt
```diff
@@ -255,3 +255,27 @@ helm install ksm https://charts.helm.sh/stable/packages/kube-state-metrics-2.9.1
...
+  {{- end }}
+{{- end }}
+
+{{- if and $hasContainerInclude (not $hasContainerExclude) }}
+
+#################################################################
+####               WARNING: Configuration notice             ####
+#################################################################
+
+You are using datadog.containerInclude but you haven't exlude any container. The default behavior is to include everything, if the intent is to exclude all other containers, set datadog.containerExclude to '.*' .
```
celenechang, line 279: typo: `exlude` -> `excluded`

typo: `any container` -> `any containers`

grammar: `everything,` -> `everything;`

PR #271, opened 2021-06-04
Fixes for the cilium network policies and additional rules for metrics

martijnvdp:

More fixes for the cilium network policies and additional rules for metrics

#### Checklist
- [ X] Chart Version bumped
- [ X] `CHANGELOG.md` has beed updated

### DIFF
--- a/charts/datadog/templates/agent-cilium-network-policy.yaml
+++ b/charts/datadog/templates/agent-cilium-network-policy.yaml
@@ -35,6 +35,12 @@ specs:
           - ports:
               - port: "123"
                 protocol: UDP
+      - toCIDR:
+          - 169.254.169.123/32
+        toPorts:
+          - ports:
+              - port: "123"
+                protocol: UDP
   - description: "Egress to metadata server"
     endpointSelector:
       matchLabels:
@@ -57,7 +63,8 @@ specs:
         {{ toYaml .Values.agents.podLabels | indent 8 }}
         {{- end }}
     egress:
-      - {{ toYaml .Values.datadog.networkPolicy.cilium.dnsSelector | nindent 8 }}
+      - toEndpoints:
+{{ toYaml .Values.datadog.networkPolicy.cilium.dnsSelector.toEndpoints | indent 10 }}
         toPorts:
           - ports:
               - port: "53"
@@ -107,6 +114,40 @@ specs:
         toPorts:
           - ports:
               - port: "10250"
+                protocol: TCP
+              - port: "9090"
+                protocol: "TCP"
+              - port: "6942"
+                protocol: "TCP"
+  - description: "Egress to cluster agent"
+    endpointSelector:
+      matchLabels:
+        app: {{ template "datadog.fullname" . }}
+        {{- if .Values.agents.podLabels }}
+        {{ toYaml .Values.agents.podLabels | indent 8 }}
+        {{- end }}
+    egress:
+      - toEndpoints:
+          - matchLabels:
+              app: {{ template "datadog.fullname" . }}-cluster-agent
+        toPorts:
+          - ports:
+              - port: "5005"
+                protocol: TCP
+  - description: "Egress to kube-state-metrics"
+    endpointSelector:
+      matchLabels:
+        app: {{ template "datadog.fullname" . }}
+        {{- if .Values.agents.podLabels }}
+        {{ toYaml .Values.agents.podLabels | indent 8 }}
+        {{- end }}
+    egress:
+      - toEndpoints:
+          - matchLabels:
+              name: kube-state-metrics
+        toPorts:
+          - ports:
+              - port: "8080"
                 protocol: TCP
 {{- if $.Values.datadog.dogstatsd.port }}
   - description: "Ingress for dogstatsd"
--- a/charts/datadog/templates/agent-cilium-network-policy.yaml
+++ b/charts/datadog/templates/agent-cilium-network-policy.yaml
@@ -115,9 +115,21 @@ specs:
           - ports:
               - port: "10250"
                 protocol: TCP
-              - port: "9090"
-                protocol: "TCP"
               - port: "6942"
+                protocol: "TCP"
+  - description: "Egress to Prometheus metrics"
+    endpointSelector:
+      matchLabels:
+        app: {{ template "datadog.fullname" . }}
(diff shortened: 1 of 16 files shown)


charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -35,6 +35,12 @@ specs:
           - ports:
               - port: "123"
                 protocol: UDP
+      - toCIDR:
+          - 169.254.169.123/32
```
clamoriniere, line 39: I would prefer to have a list on ip range in the `values.yaml` to easily update/add other ip range

clamoriniere: Also quick questions (because I'n not an expert on cilium network policy)
* what is the behaviour of the rule if `toCIDR` is empty? does it mean every IPs are allow, or every IPs are disallow?

charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -108,6 +115,40 @@ specs:
           - ports:
               - port: "10250"
                 protocol: TCP
+              - port: "9090"
```
clamoriniere, line 118: it can be useful to document the purpose of each port.

charts/datadog/templates/agent-cilium-network-policy.yaml
martijnvdp: Leaving it empty is not an option than the policy format gets messed up, i dont think you can do a null check on a helm value , so maybe with a boolean. `datadog.networkPolicy.cilium.ntp.enabled` and `datadog.networkPolicy.cilium.ntp.toCIDR`

charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -108,6 +115,52 @@ specs:
           - ports:
               - port: "10250"
                 protocol: TCP
+              - port: "6942"
+                protocol: "TCP"
```
L3n41c, line 119: What is the `6942` port about ?
Whereas I could find the `10250` as being the default kubelet port on [this page](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/), I couldn’t find any mention of `6942`.

charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -108,6 +115,52 @@ specs:
...
+        {{- end }}
+    egress:
+      - toEntities:
+          - host
+        toPorts:
+          - ports:
+              - port: "9090"
+                protocol: "TCP"
+  - description: "Egress to cluster agent"
+    endpointSelector:
```
L3n41c, line 135: Cluster agent specific parts should be generated only if the cluster agent is enabled.
```suggestion
{{- if eq (include "cluster-agent-enabled" .) "true" }}
  - description: "Egress to cluster agent"
    endpointSelector:
```

charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -108,6 +115,52 @@ specs:
...
+        {{ toYaml .Values.agents.podLabels | indent 8 }}
+        {{- end }}
+    egress:
+      - toEndpoints:
+          - matchLabels:
+              app: {{ template "datadog.fullname" . }}-cluster-agent
+        toPorts:
+          - ports:
+              - port: "5005"
+                protocol: TCP
```
L3n41c, line 148: … and here ends the cluster agent specific part.
```suggestion
                protocol: TCP
{{- end }}
```

charts/datadog/templates/agent-cilium-network-policy.yaml
```diff
@@ -108,6 +115,52 @@ specs:
...
+        {{ toYaml .Values.agents.podLabels | indent 8 }}
+        {{- end }}
+    egress:
+      - toEndpoints:
+          - matchLabels:
+              name: kube-state-metrics
+        toPorts:
+          - ports:
+              - port: "8080"
+                protocol: TCP
```
L3n41c, line 163: `kube-state-metrics` is only one of the numerous checks that a node agent can run.
There is currently no mechanism to dynamically open or close ports in the NetworkPolicy based on the checks that are scheduled by the auto-discovery.
That’s why there’s already [an “allow all egress traffic” rule here](https://github.com/DataDog/helm-charts/blob/7fb6a6ea7e079778d403ae284f3f1f45eaf3b1ac/charts/datadog/templates/agent-cilium-network-policy.yaml#L143-L162).

Rather than introducing a `kube-state-metrics` specific rule, I’d be in favour of fixing the “allow all egress” rule if it’s currently broken.

charts/datadog/templates/agent-clusterchecks-cilium-network-policy.yaml
```diff
@@ -64,6 +67,18 @@ specs:
...
+      matchLabels:
+        app: {{ template "datadog.fullname" . }}-clusterchecks
+    egress:
+      - toEndpoints:
+          - matchLabels:
+              name: kube-state-metrics
+        toPorts:
+          - ports:
+              - port: "8080"
+                protocol: TCP
```
L3n41c, line 81: I would have the same remark as for the node agent: the cluster checks runners have [an “allow all egress traffic” rule here](https://github.com/DataDog/helm-charts/blob/7fb6a6ea7e079778d403ae284f3f1f45eaf3b1ac/charts/datadog/templates/agent-clusterchecks-cilium-network-policy.yaml#L67-L83).

We should rather fix this “allow all egress” rule rather than adding a `kube-state-metric` specific one.

charts/datadog/templates/kube-state-metrics-cilium-network-policy.yaml
```diff
@@ -33,21 +33,32 @@ specs:
...
+      matchLabels:
+        app.kubernetes.io/name: kube-state-metrics
+    ingress:
+      - fromEndpoints:
+          - matchLabels:
+              app:  {{ template "datadog.fullname" . }}-clusterchecks
+        toPorts:
+          - ports:
+              - port: "8080"
+                protocol: TCP
```
L3n41c, line 63: Isn’t `kube-state-metrics` monitored by the node agent running on the same node rather than by a cluster checks runner ?

PR #267, opened 2021-05-31
OpenShift: Automatically use built-in SCCs instead of failing if create SCC option is not used

vboulineau:
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #278, opened 2021-06-08
[datadog] Update apiVersion in Chart.yaml to v2

shibataka000:

#267 introduce `lookup` function to `datadog` chart.

https://github.com/DataDog/helm-charts/blob/cc00bee0eb1b228658d9c61eae4063520ef66a17/charts/datadog/templates/_helpers.tpl#L28

`lookup` function can be used by not helm 2 but only helm 3.

According to following guide of `apiVersion` field, `apiVersion` in `charts/datadog/Chart.yaml` should be `v2` .

https://helm.sh/docs/topics/charts/#the-apiversion-field

> The apiVersion field should be v2 for Helm charts that require at least Helm 3. Charts supporting previous Helm versions have an apiVersion set to v1 and are still installable by Helm 3.

This PR update `apiVersion` in `charts/datadog/Chart.yaml` to `v2` .

None

None

#### Checklist
- [x] Chart Version bumped


vboulineau:
Hi @shibataka000,

Thanks for your contribution.
Unfortunately dropping support for Helm v2 is not possible, at least without releasing a major change.
We need to solve this issue though, so it means we probably need to revert or workaround the change introduced by #267 .

shibataka000:
Hi @vboulineau , thank you for your comment.

> We need to solve this issue though, so it means we probably need to revert or workaround the change introduced by #267 .

I agree your opinion. Do you have any plan to do it?

vboulineau:
Of course, I'll have a look in the coming days

shibataka000:
@vboulineau I created issue #279 instead of this PR. I'll close this PR. Thank you!

PR #289, opened 2021-06-24
Use a single entrypoint for the datadog agent docker image

L3n41c:

Use a single entrypoint for the datadog agent docker image.

GKE auto-pilot requires to have a fixed entrypoint in the docker image.

This new version of the chart requires an agent docker image that includes DataDog/datadog-agent#8490.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/agent-clusterchecks-deployment.yaml
+++ b/charts/datadog/templates/agent-clusterchecks-deployment.yaml
@@ -64,9 +64,9 @@ spec:
       - name: init-volume
         image: "{{ include "image-path" (dict "root" .Values "image" .Values.clusterChecksRunner.image) }}"
         imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
-        command: ["bash", "-c"]
-        args:
-          - cp -r /etc/datadog-agent /opt
+        env:
+          - name: ENTRYPOINT
+            value: init-volume
         volumeMounts:
           - name: config
             mountPath: /opt/datadog-agent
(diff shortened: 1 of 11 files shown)


charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -75,9 +75,10 @@ spec:
       - name: init-config
         image: "{{ include "image-path" (dict "root" .Values "image" .Values.clusterChecksRunner.image) }}"
         imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
-        command: ["bash", "-c"]
-        args:
-          - for script in $(find /etc/cont-init.d/ -type f -name '*.sh' | sort) ; do bash $script ; done
+        env:
+          - name: ENTRYPOINT
+            value: init-config
+          {{- include "containers-common-env" . | nindent 10 }}
```
clamoriniere, line 81: before the `containers-common-env` wasn't added.

L3n41c: It is needed because the `init-config` init container executes all the [`cont-init.d` scripts](https://github.com/DataDog/datadog-agent/tree/0ffb5ac1eaaaeed37c15a245f5c09eac4acb9583/Dockerfiles/agent/entrypoint), among which the [`01-check-apikey.sh` script](https://github.com/DataDog/datadog-agent/blob/0ffb5ac1eaaaeed37c15a245f5c09eac4acb9583/Dockerfiles/agent/entrypoint/01-check-apikey.sh) which is checking that `DD_API_KEY` is set.

And `DD_API_KEY` is set by `containers-common-env`.
The reason why it was not a problem before was because [errors used to be silently ignored](https://github.com/DataDog/helm-charts/pull/289/files#diff-2d5d99450009939b13edaea8bc6c04ed07dbdcb45f59247bc27b3fc63e6593f5L18) and [they are now checked](https://github.com/DataDog/datadog-agent/pull/8490/files#diff-79990b709c64f33937cf346d8c67adcb942795a68d6b9b2fe2477af5d4dd1f45R2).

brycekahle (approved):

PR #300, opened 2021-07-05
[datadog] Expose UDS for Dogstatsd in an emptyDir volume by default

clamoriniere:

Expose the Unix Domain Sockect for Dogstatsd in an emptyDir volume by default. It will allow
the "JMX-Fetch" process to always use the UDS socket to send metrics to the `core-agent`.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-trace-agent.yaml
+++ b/charts/datadog/templates/_container-trace-agent.yaml
@@ -66,15 +66,17 @@
       mountPath: {{ print "/host/" (dir (include "datadog.dockerOrCriSocketPath" .)) | clean }}
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
-    - name: dsdsocket
-      mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
-      readOnly: true
     {{- end }}
     {{- if eq .Values.targetSystem "windows" }}
     - name: runtimesocket
       mountPath: {{ template "datadog.dockerOrCriSocketPath" . }}
     {{- end }}
-    {{- if .Values.datadog.apm.useSocketVolume }}
+    - name: dsdsocket
+      mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
+    {{- if not .Values.datadog.apm.useSocketVolume }}
+      readOnly: true
+    {{- else if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
+      readOnly: true
     - name: apmsocket
       mountPath: {{ (dir .Values.datadog.apm.socketPath) }}
     {{- end }}
--- a/charts/datadog/templates/_container-trace-agent.yaml
+++ b/charts/datadog/templates/_container-trace-agent.yaml
@@ -41,6 +41,10 @@
     - name: DD_APM_RECEIVER_SOCKET
       value: {{ .Values.datadog.apm.socketPath | quote }}
   {{- end }}
+  {{- if eq .Values.targetSystem "linux" }}
+    - name: DD_DOGSTATSD_SOCKET
+      value: {{ .Values.datadog.dogstatsd.socketPath | quote }}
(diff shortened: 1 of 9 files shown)


L3n41c (approved):

olivielpeau (commented):
This makes sense!

Note that trace-agent (and process-agent, and maybe other agent containers) also submit dogstatsd metrics, and at least the trace-agent picks up `DD_DOGSTATSD_SOCKET` when it's set. I think it's fine to leave them to use UDP by default, we'll simply want to make sure `DD_DOGSTATSD_SOCKET` is only set on their containers when the socket volume is mounted as well, which is true today (the env var is never set on `trace-agent`, and the volume isn't mounted)

Implementation comment: I think the `emptyDir` volume needs to be mounted in the core agent's container as well, when `datadog.dogstatsd.useSocketVolume` is `false`?

clamoriniere:
> Note that trace-agent (and process-agent, and maybe other agent containers) also submit dogstatsd metrics, and at least the trace-agent picks up `DD_DOGSTATSD_SOCKET` when it's set. I think it's fine to leave them to use UDP by default, we'll simply want to make sure `DD_DOGSTATSD_SOCKET` is only set on their containers when the socket volume is mounted as well, which is true today (the env var is never set on `trace-agent`, and the volume isn't mounted)

Yes we can extend the usage of the UDS for the other agent containers (except system-probe). 

> Implementation comment: I think the `emptyDir` volume needs to be mounted in the core agent's container as well, when `datadog.dogstatsd.useSocketVolume` is `false`?

It should be the case, the idea is:

* `datadog.dogstatsd.useSocketVolume=false` => the socket is created in the "emptyDir" volume
* `datadog.dogstatsd.useSocketVolume=true` => the socket is created in the "hostPath" volume

So in any case the socket exist. And it is why I removed the [condition](https://github.com/DataDog/helm-charts/pull/300/files#diff-0973c8baa0f902471c709f7be2b0682aa0b89d92368289528c1d4b42118980a3L77) in the `volumeMounts` list.

clamoriniere:
@olivielpeau 
good catch I forgot remove the condition on the `dsdsocket` volumeMount creation for the "core agent".

L3n41c (approved):

L3n41c (approved):

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -67,7 +71,10 @@
     - name: runtimesocket
       mountPath: {{ template "datadog.dockerOrCriSocketPath" . }}
     {{- end }}
-    {{- if .Values.datadog.apm.useSocketVolume }}
+    - name: dsdsocket
+      mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
+    {{- if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
+      readOnly: true
```
olivielpeau, line 77: Technically I think `readOnly: true` could be set even if `.Values.datadog.apm.useSocketVolume` is false (the only condition that matters here is `ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)`).

olivielpeau (approved):
Just a question on the trace-agent container, but LGTM in any case 👍

charts/datadog/templates/_container-trace-agent.yaml
clamoriniere: I think we are still missing a corner case, when `datadog.apm.useSocketVolume` not `true` but `datadog.dogstatsd.socketPath != datadog.apm.socketPath`

```yaml
    - name: dsdsocket
    {{- if not .Values.datadog.apm.useSocketVolume) }}
      readOnly: true
    {{- else if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
      readOnly: true
    - name: apmsocket
      mountPath: {{ (dir .Values.datadog.apm.socketPath) }}
    {{- end }}
```

PR #313, opened 2021-07-09
[datadog] Fix volumeMount for the `trace-agent` on windows

clamoriniere:
#### What this PR does / why we need it:

* Fix `dsdsocket` volumeMount for the `trace-agent` on windows deployment.

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #318, opened 2021-07-15
Remove ref to readonly for socket mounts in trace agent container for Windows

Simwar:
This PR fixes the trace agent container by removing the reference to socket read only mounts on Windows.
Pipes cannot be mounted as read only on windows and this PR introduced a read only mounts by mistake: https://github.com/DataDog/helm-charts/pull/313 for the runtimesocket mount.

Maybe I overlooked a logic introduced by https://github.com/DataDog/helm-charts/pull/300 for the readonly mounts, but I think this should not be there anyway.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-trace-agent.yaml
+++ b/charts/datadog/templates/_container-trace-agent.yaml
@@ -64,17 +64,23 @@
       readOnly: false
     - name: dsdsocket
       mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
+    {{- if not .Values.datadog.apm.useSocketVolume }}
+      readOnly: true
+    {{- else if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
+      readOnly: true
+    {{- end }}
     - name: runtimesocketdir
       mountPath: {{ print "/host/" (dir (include "datadog.dockerOrCriSocketPath" .)) | clean }}
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
+    {{- if .Values.datadog.apm.useSocketVolume }}
+    - name: apmsocket
+      mountPath: {{ (dir .Values.datadog.apm.socketPath) }}
+    {{- end }}
     {{- end }}
     {{- if eq .Values.targetSystem "windows" }}
     - name: runtimesocket
       mountPath: {{ template "datadog.dockerOrCriSocketPath" . }}
-    {{- end }}
-    - name: apmsocket
-      mountPath: {{ (dir .Values.datadog.apm.socketPath) }}
     {{- end }}
     {{- if .Values.datadog.kubelet.hostCAPath }}
 {{ include "datadog.kubelet.volumeMount" . | indent 4 }}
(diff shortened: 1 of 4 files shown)


vboulineau:
The section you removed is not under the
```
{{- if eq .Values.targetSystem "windows" }}
``` section.

I think the `apmsocket` part should be inside the Linux part ATM.

File: charts/datadog/templates/_container-trace-agent.yaml [hunk 1]
```diff
@@ -89,4 +91,4 @@
   livenessProbe:
 {{- $live := .Values.agents.containers.traceAgent.livenessProbe }}
 {{ include "probe.tcp" (dict "port" .Values.datadog.apm.port "settings" $live ) | indent 4 }}
-{{- end -}}
+{{- end -}}
```
clamoriniere, line 94:
```suggestion
{{- end -}}

```

clamoriniere:
Hi @Simwar 

because by default in the helm chart, the APM and DSD socket folder is the same, but can be override by the user. the idea was to set the readonly option only if the APM and DSD socket folder was different, to have the more restricted permission configuration.

* If APM and DSD sockets are in the same volume: we need to be in Read/Write because the trace-agent needs to create the socket.
* If APM and DSD sockets are in different volume:
   * the APM socket volume mount will be in Read/Write to allow the trace-agent to create the socket if neeeded
   * the DSD socket volume mount will be in ReadOnly because the trace-agent only need to access the socket. 

I wasn't aware of a different behaviour on windows.

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -64,22 +64,24 @@
...
+      readOnly: true
+    {{- end }}
     - name: runtimesocketdir
       mountPath: {{ print "/host/" (dir (include "datadog.dockerOrCriSocketPath" .)) | clean }}
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
+    {{- if .Values.datadog.apm.useSocketVolume }}
+    - name: apmsocket
+      mountPath: {{ (dir .Values.datadog.apm.socketPath) }}
+    {{- end }}
```
clamoriniere, line 79: could you move `apmsocket` just after the `dsdsocket 🙏

clamoriniere (approved):

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -64,6 +64,15 @@
       readOnly: false
     - name: dsdsocket
       mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
+    {{- if not .Values.datadog.apm.useSocketVolume }}
+      readOnly: true
+    {{- else if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
+      readOnly: true
+    {{- end }}
```
clamoriniere, line 71: lets simplify think here, and allow always read/write

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -64,6 +64,15 @@
...
+    {{- if .Values.datadog.apm.useSocketVolume }}
```
clamoriniere, line 72: Here you need to check that the path is different from the `dsdsocket`

```suggestion
    {{- if and .Values.datadog.apm.useSocketVolume (ne (dir .Values.datadog.dogstatsd.socketPath) (dir .Values.datadog.apm.socketPath)) }}
```

PR #321, opened 2021-07-15
[datadog] fix `should-enable-k8s-resource-monitoring` definition

celenechang:

Fixes issue where disabling cluster agent wasn't sufficient to disable orchestrator explorer. Prior to the fix, it was noticed that `DD_ORCHESTRATOR_EXPLORER_ENABLED` was set to `true` even if `DD_CLUSTER_AGENT_ENABLED` was `false`. This is because the statement `include "existingClusterAgent-configured" .` in the `if` condition was always returning `true`.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.19.3
 
-* Fix condition defining `should-enable-k8s-resource-monitoring`.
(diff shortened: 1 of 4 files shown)


ahmed-mez (approved):
could we make the message clearer for users who don't necessarily know much about the internal helper `should-enable-k8s-resource-monitoring` ?

xlucas (approved):

PR #322, opened 2021-07-16
Add configuration for runtime compilation with kernel header downloading in the system-probe

ISauve:

Supercedes #146

- Adds configuration options for enabling/configuring the runtime compiler with kernel header downloading in the system-probe.
- If kernel header downloading is enabled, mounts directories necessary for kernel header downloading:
   - kernel-headers-download-dir (`/var/tmp/datadog-agent/system-probe/kernel-headers` by default, though this is configurable)
   - package management folders (`/etc/apt`, `/etc/yum.repos.d`, `/etc/pki`, `/etc/rhsm`, `/etc/zypp`)
 - Mounts OS release files (`/etc/os-release`, `/etc/redhat-release`, `/etc/fedora-release`, `/etc/lsb-release`) to `/host/etc` for correct host OS detection in a containerized environment (see https://github.com/DataDog/datadog-agent/pull/10125)
- Adds `flock`, `renameat` and `symlinkat` to the list of allowed syscalls in the seccomp profile

  - fixes #

#### Checklist
- [x] Chart Version bumped


brycekahle:
What do you think about only mounting `/var/tmp/datadog-agent/system-probe` from the host. That way if we need further subdirectories within, we don't need another mount and config?

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -32,6 +32,8 @@ data:
       collect_dns_stats: {{ $.Values.datadog.systemProbe.collectDNSStats }}
       max_tracked_connections: {{ $.Values.datadog.systemProbe.maxTrackedConnections }}
       conntrack_max_state_size: {{ $.Values.datadog.systemProbe.conntrackMaxStateSize }}
+      enable_runtime_compiler: {{ $.Values.datadog.systemProbe.enableRuntimeCompiler }}
+      kernel_headers_download_dir: {{ $.Values.datadog.systemProbe.kernelHeadersDownloadDir }}
```
brycekahle, line 36: I think you are missing the setting for `runtime_compiler_output_dir`

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -54,14 +54,45 @@
...
+  name: yum-repos-dir
+- hostPath:
+    path: /etc/pki
+  name: public-key-dir
+- hostPath:
+    path: /etc/rhsm
+  name: rhel-subscription-dir
+- hostPath:
+    path: /etc/zypp
+  name: opensuse-repos-dir
```
clamoriniere, line 94: the problem with this configuration is that it will created a new folder if it doesn't exist. So user can be bother if the agent deployment create directory on the host.
Also on a harden host, the host file system can be in read only, which prevent to create new folder, and so the agent will be able to start.

Unfortunately kubernetes doesn't provide an option on hostpath to ignore none existing volume...
So I think we should add at least options to:
* disable the mount of all this volume
* (maybe) provide an option to set the path where the linux distribution store the info.

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -47,7 +47,7 @@
       mountPath: /host/proc
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
-{{- if or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill }}
+{{- if or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill .Values.datadog.systemProbe.enableRuntimeCompiler }}
```
clamoriniere, line 50: could you create an helper function to store this condition. I see we are using the same condition in 2 different places.

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -62,6 +62,35 @@
       mountPath: /etc/datadog-agent/runtime-security.d
       readOnly: true
 {{- end }}
+{{- if .Values.datadog.systemProbe.enableRuntimeCompiler }}
+    - name: runtime-compiler-output-dir
+      mountPath: /var/tmp/datadog-agent/system-probe/build
```
brycekahle, line 67: Should this be `{{ .Values.datadog.systemProbe. runtimeCompilerOutputDir }}` ?

charts/datadog/templates/_daemonset-volumes-linux.yaml
brycekahle: > disable the mount of all this volume

We could add a config option for kernel header downloading (separate from runtime compiler config option, but only read if runtime compiler is enabled). We could then disable mounting these volumes if kernel header downloading is disabled. How does that sound?

brycekahle: > * provide an option to set the path where the linux distribution store the info.

Can you clarify what you mean by this?

ISauve: @clamoriniere I've added a new flag, `mountPackageManagementDirs`, which allows users to disable the mounting of all the package management directories - does this resolve your concerns?

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -78,6 +78,43 @@
...
+      readOnly: true
+    - name: opensuse-repos-dir
+      mountPath: /host/etc/zypp
+      readOnly: true
+    - name: public-key-dir
+      mountPath: /etc/pki
+      readOnly: true
+    - name: rhel-subscription-dir
+      mountPath: /etc/rhsm
+      readOnly: true
```
celenechang, line 115: are these intentionally not mounted at `/host/` ?

ISauve: No, they should be under there as well. Good catch :)

charts/datadog/templates/_daemonset-volumes-linux.yaml
clamoriniere: Hi @ISauve,
Yes I think so.

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -78,6 +78,43 @@
...
+    - name: opensuse-repos-dir
+      mountPath: /host/etc/zypp
+      readOnly: true
+    - name: public-key-dir
+      mountPath: /host/etc/pki
+      readOnly: true
+    - name: rhel-subscription-dir
+      mountPath: /host/etc/rhsm
+      readOnly: true
+{{- end }}
```
clamoriniere, line 116: the idea that I had in my was the opposite. 
By default all the paths are added, But if the user doesn't or can't mount these path (for example read-only file system) they can provide the list of path they want.

something like:

```suggestion
{{- if not .Values.datadog.systemProbe.mountPackageManagementDirs }}
    - name: etc-os-release
      mountPath: /host/etc/os-release
      readOnly: true
    - name: etc-redhat-release
      mountPath: /host/etc/redhat-release
      readOnly: true
    - name: etc-fedora-release
      mountPath: /host/etc/fedora-release
      readOnly: true
    - name: etc-lsb-release
      mountPath: /host/etc/lsb-release
      readOnly: true
    - name: apt-config-dir
      mountPath: /host/etc/apt
      readOnly: true
    - name: yum-repos-dir
      mountPath: /host/etc/yum.repos.d
      readOnly: true
    - name: opensuse-repos-dir
      mountPath: /host/etc/zypp
      readOnly: true
    - name: public-key-dir
      mountPath: /host/etc/pki
      readOnly: true
    - name: rhel-subscription-dir
      mountPath: /host/etc/rhsm
      readOnly: true
{{- else }}
{{- range .Values.datadog.systemProbe.mountPackageManagementDirs }}
    - name: {{ .name }}
      mountPath: {{ .mountPath }}
      readOnly: true
{{- end }}
```

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -68,6 +68,45 @@
...
+{{- if eq (include "runtime-compilation-enabled" .) "true" }}
+- hostPath:
+    path: {{ .Values.datadog.systemProbe.runtimeCompilationAssetDir }}/build
+    type: DirectoryOrCreate
+  name: runtime-compiler-output-dir
+- hostPath:
+    path: {{ .Values.datadog.systemProbe.runtimeCompilationAssetDir }}/kernel-headers
+    type: DirectoryOrCreate
+  name: kernel-headers-download-dir
+{{- if .Values.datadog.systemProbe.mountPackageManagementDirs }}
```
clamoriniere, line 80: same comment that for the `volumeMounts`

charts/datadog/values.yaml
```diff
@@ -369,6 +369,15 @@ datadog:
     # datadog.systemProbe.enableOOMKill -- Enable the OOM kill eBPF-based check
     enableOOMKill: false
 
+    # datadog.systemProbe.enableRuntimeCompiler -- Enable the runtime compiler for eBPF probes
+    enableRuntimeCompiler: false
+
+    # datadog.systemProbe.mountPackageManagementDirs -- Enables mounting of various package management directories when runtime compilation is enabled
+    mountPackageManagementDirs: true
```
clamoriniere, line 376: based on my previous comments we can convert this option to an optional list of hostPath configuration

```suggestion
    mountPackageManagementDirs: []
    ## - name: "fedora-release"
    ##   hostPath: /etc/fedora-release
    ##   mountPath: /host/etc/fedora-release
```
And base on this list, the `volumeMounts` and `volume` can be created. 

Also what can be done is to define in the `values.yaml` file all the hostPath that you want to mount by default. So like this if a customer only want one path. they can just comment the others.

please let me know if I was clear

ISauve: Great idea, it's very clear! Thanks for the review :)

charts/datadog/values.yaml
```diff
@@ -369,6 +369,24 @@ datadog:
...
+    # datadog.systemProbe.mountPackageManagementDirs -- Enables mounting of specific package management directories when runtime compilation is enabled
+    mountPackageManagementDirs: []
+    ## For runtime compilation to be able to download kernel headers, the host's package management folders
+    ## must be mounted to the /host directory. For example, for Ubuntu & Debian the following mount would be necessary:
+    # - name: "apt-config-dir"
+    #   hostPath: /etc/apt
+    #   mountPath: /host/etc/apt
+    ## If this list is empty, then all necessary package management directories (for all supported OSs) will be mounted.
+
+    # datadog.systemProbe.osReleasePath -- Specify the path to your os-release file if you don't want all os-releases mounted by default
```
celenechang, line 384:
```suggestion
    # datadog.systemProbe.osReleasePath -- Specify the path to your os-release file if you don't want all `/etc/*-release` file mounts attempted by default
```

celenechang (approved):

PR #333, opened 2021-08-02
[datadog-crds] Adapt script to be able to download CRDs from the EDS repo

davidor:
#### What this PR does / why we need it:

We want to add the charts from the EDS repo in this one. To do that, we'll need to be able to download the CRDs from that repo in the "datadog-crds" chart.

This PR also runs the updated script to add the CRDs from the EDS repo v0.6.0 (latest tag).

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [ ] ~Variables are documented in the `README.md`~

PR #334, opened 2021-08-03
[datadog-crds] Fix script to download correct EDS CRDs

davidor:

This PR fixes some mistakes introduced in my previous PR (#333):

- This PR updates the `charts/datadog-crds/update-crds.sh` script so it only downloads v1beta1 from the EDS repo. There are no releases with valid v1 CRDs.
- Adds missing options in `charts/datadog-crds/values.yaml` to install the EDS CRDs.
- Regenerates the templates and CRDs with the updated `charts/datadog-crds/update-crds.sh` script.

#### Checklist
- [x] Chart Version bumped


charts/datadog-crds/values.yaml
```diff
@@ -9,6 +9,12 @@ crds:
   datadogAgents: false
   # crds.datadogMonitors -- Set to true to deploy the DatadogMonitors CRD
   datadogMonitors: false
+  # crds.extendedDaemonSetReplicaSets -- Set to true to deploy the ExtendedDaemonSetReplicaSets CRD
```
davidor, line 12: For this I followed the same pattern used for the operator CRDs. However, in this case I wonder if we could simplify this to just one option that installs all the EDS CRDs or none.

charts/datadog-crds/values.yaml
```diff
@@ -9,6 +9,12 @@ crds:
...
+  extendedDaemonSetReplicaSets: false
+  # crds.extendedDaemonSets -- Set to true to deploy the ExtendedDaemonSet CRD
+  extendedDaemonSets: false
+  # crds.extendedDaemonSetSettings -- Set to true to deploy the ExtendedDaemonSettings CRD
+  extendedDaemonSetSettings: false
```
clamoriniere, line 17: could you remove the `extendedDaemonSetSettings` from this CRD import script, because we don't want to maintain it in the futur.

clamoriniere (approved):

charts/datadog-crds/values.yaml
clamoriniere: yes I think we can have only one settings for the 3 EDS crds, because it is not possible to disable one of them in the EDS-controller.

davidor: Fixed :+1:

PR #380, opened 2021-09-14
[extended-daemon-set] Add CRDs inside the chart

clamoriniere:

* Add ExtendedDaemonset CRDs directly inside this chart.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/extended-daemon-set/CHANGELOG.md
+++ b/charts/extended-daemon-set/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.2.1
+
+* Add ExtendedDaemonset CRDs directly inside this chart.
+
 ## 0.2.0
 
 * Updated for EDS 0.7.0.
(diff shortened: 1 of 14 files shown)


vboulineau:
I believe it removes the ability to NOT install the CRDs

clamoriniere:
> I believe it removes the ability to NOT install the CRDs

Yes, to you think I should add back this capability? I don't see a real use case for it

vboulineau:
I agree the use case i thin, but I think some customers may handle CRDs without Helm and it may be useful it case something goes wrong during Helm release lifecycle for some reason.

davidor (commented):
I think it would be nice to have some tooling, like the `update-crds.sh` script that we used before, to fetch the CRDs from the EDS repo.

vboulineau (approved):

clamoriniere:
> I think it would be nice to have some tooling, like the `update-crds.sh` script that we used before, to fetch the CRDs from the EDS repo.

@davidor pull request updated with the script and doc

davidor (approved):
Looks good to me :+1:

In a different PR we can try to provide a shared script to update the CRDs.
The one introduced in this PR is almost identical to the one we have under `datadog-crds`. It'd be nice to avoid this duplication.

PR #395, opened 2021-09-24
[datadog] Expose the `labels_as_tags` parameter of the KSM core check

L3n41c:

Expose the `labels_as_tags` setting of the KSM core check introduced by DataDog/datadog-agent#9214 in the Helm chart.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Datadog changelog
 
+## 2.27.2
+
+* Expose the `labels_as_tags` parameter of the KSM core check.
+  This parameter exists only in agent 7.32.0 and above and cluster-agent 1.16.0 and above.
+
 # 2.27.1
 
 * Update README.md to clarify Helm 2 vs. Helm 3 instructions.
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -104,6 +104,9 @@ datadog:
     ##     <label2>: <tag2>
     ##   <resource2>:
     ##     <label3>: <tag3>
+    ##
+    ## Warning: the label must match the transformation done by kube-state-metrics,
+    ## example:  tags.datadoghq.com/version becomes label_tags_datadoghq_com_version 
     labelsAsTags: {}
     #  pod:
     #    app: app
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -96,6 +96,21 @@ datadog:
     ## ref: https://docs.datadoghq.com/agent/cluster_agent/clusterchecksrunner?tab=helm
     useClusterCheckRunners: false
 
+    # datadog.kubeStateMetricsCore.labelsAsTags -- Extra labels to collect from resources and to turn into datadog tag.
+    ## It has the following structure:
+    ## labelsAsTags:
+    ##   <resource1>:        # can be pod, deployment, node, etc.
+    ##     <label1>: <tag1>  # where <label1> is the kubernetes label and <tag1> is the datadog tag
+    ##     <label2>: <tag2>
+    ##   <resource2>:
+    ##     <label3>: <tag3>
+    labelsAsTags: {}
+    #  pod:
+    #    app: app
+    #  node:
+    #    zone: zone
+    #    team: team
+
   ## Manage Cluster checks feature
   ## ref: https://docs.datadoghq.com/agent/autodiscovery/clusterchecks/
   ## Autodiscovery via Kube Service annotations is automatically enabled
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -106,7 +106,7 @@ datadog:
     ##     <label3>: <tag3>
     ##
     ## Warning: the label must match the transformation done by kube-state-metrics,
-    ## example:  tags.datadoghq.com/version becomes label_tags_datadoghq_com_version 
+    ## for example tags.datadoghq.com/version becomes label_tags_datadoghq_com_version.
     labelsAsTags: {}
     #  pod:
     #    app: app
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.27.1
+version: 2.27.2
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.27.1](https://img.shields.io/badge/Version-2.27.1-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 2.27.2](https://img.shields.io/badge/Version-2.27.2-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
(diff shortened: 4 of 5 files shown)


charts/datadog/templates/agent-clusterchecks-confd-configmap.yaml
```diff
@@ -0,0 +1,39 @@
...
+        - daemonsets
+        - deployments
+        - replicasets
+        - statefulsets
+        - cronjobs
+        - jobs
+        - horizontalpodautoscalers
+        - poddisruptionbudgets
+        - storageclasses
+        - volumeattachments
```
clamoriniere, line 35: can we remove this part? 
maybe we should add the logic in the kubernetes_state_core check, that if the list is empty or not define we use the default list of collector.
WDYT?

ahmed-mez: > maybe we should add the logic in the kubernetes_state_core check, that if the list is empty or not define we use the default list of collector.

We already have this logic in the check itself, it would watch all the possible resources. The reason why we specify the collectors explicitly here is to make sure the helm chart knows what RBACs should be added to make the ksm core check work.

charts/datadog/values.yaml
```diff
@@ -87,6 +87,22 @@ datadog:
     ## Migration guide: https://docs.datadoghq.com/integrations/kubernetes_state_core/?tab=helm#migration-from-kubernetes_state-to-kubernetes_state_core
     ignoreLegacyKSMCheck: true
 
+    # datadog.kubeStateMetricsCore.labelsAsTags --
+    ## Extra labels to collect from resources and to turn into datadog tag.
```
ahmed-mez, line 91:
```suggestion
    # datadog.kubeStateMetricsCore.labelsAsTags -- Extra labels to collect from resources and to turn into datadog tag.
```
to make the comment appear in the readme

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.22.8
+
+* Expose the `labels_as_tags` parameter of the KSM core check.
```
ahmed-mez, line 5: Let's mention the required agent/dca versions

charts/datadog/templates/_helpers.tpl
```diff
@@ -503,3 +503,14 @@ true
...
+{{/*
+Return true if the Cluster check runners need a confd configmap
+*/}}
+{{- define "need-clusterchecks-confd" -}}
+{{- if (or .Values.datadog.kubeStateMetricsCore.enabled) -}}
+true
+{{- else -}}
+false
+{{- end -}}
+{{- end -}}
```
ahmed-mez, line 516: Not sure why we'd need this 🤔  The cluster check runners must run the check as a cluster check (dispatched by the DCA) not with static config. Am I missing something?

charts/datadog/templates/agent-clusterchecks-confd-configmap.yaml
clamoriniere: @ahmed-mez 

but the RBAC list is managed by another file? 
should we build a function that generate the collectors list based on the rbac?

ahmed-mez: The RBACs are defined here https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/kube-state-metrics-core-rbac.yaml

> should we build a function that generate the collectors list based on the rbac?

not sure I understand, you mean we should expose the RBACs list in the values file? If this what you meant I think it's a good idea, it can also be useful for custom ksm core configurations.

charts/datadog/CHANGELOG.md
L3n41c: Good idea !
Done: https://github.com/DataDog/helm-charts/pull/395/files#diff-0c47f15ee87d7464ff10968fa5900d272dd547e8963b2aee06a8e83537ad1bd7R6

charts/datadog/values.yaml
L3n41c: 👍🏻 Good idea.
Done !

charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -174,6 +174,11 @@ spec:
             mountPath: /etc/datadog-agent/install_info
             {{- end }}
             readOnly: true
+{{- if eq (include "need-clusterchecks-confd" .) "true" }}
+          - name: confd
+            mountPath: /conf.d
+            readOnly: true
+{{- end }}
```
clamoriniere, line 181: the change that you made here are not needed. since the clustercheckrunner takes configuration from the DCA

charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -191,6 +196,11 @@ spec:
             name: {{ template "datadog.fullname" . }}-installinfo
 {{- if .Values.clusterChecksRunner.volumes }}
 {{ toYaml .Values.clusterChecksRunner.volumes | indent 8 }}
+{{- end }}
+{{- if eq (include "need-clusterchecks-confd" . ) "true" }}
+        - name: confd
+          configMap:
+            name: {{ template "datadog.fullname" . }}-clusterchecks-confd
```
clamoriniere, line 203: same

charts/datadog/templates/agent-clusterchecks-confd-configmap.yaml
clamoriniere: agree all this file can be removed

L3n41c: Removed !

charts/datadog/templates/agent-clusterchecks-deployment.yaml
L3n41c: Removed !

charts/datadog/templates/agent-clusterchecks-deployment.yaml
L3n41c: 👍🏻

charts/datadog/values.yaml
```diff
@@ -96,6 +96,21 @@ datadog:
     ## ref: https://docs.datadoghq.com/agent/cluster_agent/clusterchecksrunner?tab=helm
     useClusterCheckRunners: false
 
+    # datadog.kubeStateMetricsCore.labelsAsTags -- Extra labels to collect from resources and to turn into datadog tag.
+    ## It has the following structure:
+    ## labelsAsTags:
+    ##   <resource1>:        # can be pod, deployment, node, etc.
+    ##     <label1>: <tag1>  # where <label1> is the kubernetes label and <tag1> is the datadog tag
```
ahmed-mez, line 103: I think it's worth adding a concrete example, because the the label must match the transformation done by ksm, example:  `tags.datadoghq.com/version` becomes `label_tags_datadoghq_com_version`

ahmed-mez (approved):

clamoriniere (approved):

charts/datadog/values.yaml
```diff
@@ -96,6 +96,24 @@ datadog:
...
+    ##     <label2>: <tag2>
+    ##   <resource2>:
+    ##     <label3>: <tag3>
+    ##
+    ## Warning: the label must match the transformation done by kube-state-metrics,
+    ## example:  tags.datadoghq.com/version becomes label_tags_datadoghq_com_version
```
clamoriniere, line 109:
```suggestion
    ## for example tags.datadoghq.com/version becomes label_tags_datadoghq_com_version
```

PR #422, opened 2021-10-19
Add option to use CLCW for KSM core

CharlyF:

Adds an option to easily run the KSM Core check as a distributed cluster check.

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -303,6 +233,23 @@ subjects:
     name: {{ template "datadog.fullname" . }}-cluster-agent
     namespace: {{ .Release.Namespace }}
 ---
+{{- if and .Values.datadog.kubeStateMetricsCore.enabled (not .Values.datadog.kubeStateMetricsCore.useClusterCheckWorkers) }}
```
L3n41c, line 236: That’s really a tiny detail, but as `---` is the markdown marker for the beginning of a document, it makes more sense to include the leading `---` in the conditional `{{- if` block.
(The markdown marker for the end of a document is `...` but is usually omitted.)

```suggestion
{{- if and .Values.datadog.kubeStateMetricsCore.enabled (not .Values.datadog.kubeStateMetricsCore.useClusterCheckWorkers) }}
---
```

charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -303,6 +233,23 @@ subjects:
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "datadog.fullname" . }}-ksm-core
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "datadog.fullname" . }}-cluster-agent
+    namespace: {{ .Release.Namespace }}
+---
+{{- end }}
```
L3n41c, line 252: Following the above comment, as `---` is the marker of the beginning of a yaml document, it looks more natural to have the `---` of the next document outside this `{{- if` block.

```suggestion
{{- end }}
---
```

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -149,9 +149,9 @@ spec:
           - name: DD_ADMISSION_CONTROLLER_SERVICE_NAME
             value: {{ template "datadog.fullname" . }}-cluster-agent-admission-controller
           {{- end }}
-          {{- if .Values.datadog.clusterChecks.enabled }}
+          {{- if (eq (include "should-enable-cluster-check-workers" .) "true") }}
```
ahmed-mez, line 152: I don't think we should do this, because we can run cluster checks without cluster check runners. This change will disable the checks for users who use node agents to run cluster checks.

charts/datadog/templates/agent-clusterchecks-rbac.yaml
```diff
@@ -14,6 +14,23 @@ subjects:
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "datadog.fullname" . }}-ksm-core
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "datadog.fullname" . }}-cluster-checks
+    namespace: {{ .Release.Namespace }}
+---
+{{- end }}
```
L3n41c, line 33: We could factorise the creation of the `ClusterRoleBinding` for KSM to have only one definition but with a condition for the `ServiceAccount`.

```suggestion
{{- if and .Values.datadog.kubeStateMetricsCore.enabled  -}}
---
apiVersion: {{ template "rbac.apiVersion" . }}
kind: ClusterRoleBinding
metadata:
  labels:
{{ include "datadog.labels" . | indent 4 }}
  name: {{ template "datadog.fullname" . }}-ksm-core
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: {{ template "datadog.fullname" . }}-ksm-core
subjects:
  - kind: ServiceAccount
    {{- if .Values.datadog.kubeStateMetricsCore.useClusterCheckWorkers }}
    name: {{ template "datadog.fullname" . }}-cluster-checks
    {{- else }}
    name: {{ template "datadog.fullname" . }}-cluster-agent
    {{- endif }}
    namespace: {{ .Release.Namespace }}
{{- end }}
---
```

But in this case, this “generic” `ClusterRoleBinding` definition should move to `charts/datadog/templates/kube-state-metrics-core-cr.yaml`.

charts/datadog/templates/cluster-agent-rbac.yaml
L3n41c, line 236: That’s really a tiny detail, but as `---` is the markdown marker for the beginning of a document, it makes more sense to include the leading `---` in the conditional `{{- if` block.
(The markdown marker for the end of a document is `...` but is usually omitted.)

```suggestion
{{- if and .Values.datadog.kubeStateMetricsCore.enabled (not .Values.datadog.kubeStateMetricsCore.useClusterCheckWorkers) }}
---
```

charts/datadog/templates/cluster-agent-rbac.yaml
L3n41c, line 252: Following the above comment, as `---` is the marker of the beginning of a yaml document, it looks more natural to have the `---` of the next document outside this `{{- if` block.

```suggestion
{{- end }}
---
```

charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -303,6 +233,23 @@ subjects:
...
+  name: {{ template "datadog.fullname" . }}-cluster-agent-ksm-core
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "datadog.fullname" . }}-ksm-core
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "datadog.fullname" . }}-cluster-agent
+    namespace: {{ .Release.Namespace }}
+---
```
L3n41c, line 251: I think this can be removed if we create a templated `ClustorRoleBinding` for both the DCA and the CLC as suggested above.

charts/datadog/templates/_helpers.tpl
```diff
@@ -415,6 +415,17 @@ false
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return true if the Cluster Check Workers have to be deployed
+*/}}
+{{- define "should-enable-cluster-check-workers" -}}
```
ahmed-mez, line 421: nit
```suggestion
{{- define "should-enable-cluster-check-runners" -}}
```

charts/datadog/templates/NOTES.txt
```diff
@@ -86,7 +86,7 @@ Cluster Agent liveness probe port ({{ $liveness.port }}) is different from the c
 Cluster Agent readiness probe port ({{ $readiness.port }}) is different from the configured health port ({{ $healthPort }}).
   {{- end }}
   {{- end }}
-  {{- if and .Values.datadog.clusterChecks.enabled .Values.clusterChecksRunner.enabled }}
+  {{- if (eq (include "should-enable-cluster-check-workers" .) "true") }}
```
ahmed-mez, line 89:
```suggestion
  {{- if (eq (include "should-enable-cluster-check-runners" .) "true") }}
```

charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and (eq (include "should-deploy-cluster-agent" .) "true") .Values.datadog.clusterChecks.enabled .Values.clusterChecksRunner.enabled -}}
+{{- if and (eq (include "should-deploy-cluster-agent" .) "true") (eq (include "should-enable-cluster-check-workers" .) "true") -}}
```
ahmed-mez, line 1:
```suggestion
{{- if and (eq (include "should-deploy-cluster-agent" .) "true") (eq (include "should-enable-cluster-check-runners" .) "true") -}}
```

charts/datadog/templates/agent-clusterchecks-rbac.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.clusterChecksRunner.rbac.create (eq (include "should-deploy-cluster-agent" .) "true") .Values.datadog.clusterChecks.enabled .Values.clusterChecksRunner.enabled .Values.clusterChecksRunner.rbac.dedicated -}}
+{{- if and .Values.clusterChecksRunner.rbac.create (eq (include "should-deploy-cluster-agent" .) "true") (eq (include "should-enable-cluster-check-workers" .) "true") .Values.clusterChecksRunner.rbac.dedicated -}}
```
ahmed-mez, line 1:
```suggestion
{{- if and .Values.clusterChecksRunner.rbac.create (eq (include "should-deploy-cluster-agent" .) "true") (eq (include "should-enable-cluster-check-runners" .) "true") .Values.clusterChecksRunner.rbac.dedicated -}}
```

charts/datadog/templates/cluster-agent-confd-configmap.yaml
```diff
@@ -14,6 +14,9 @@ data:
 {{- end }}
 {{- if .Values.datadog.kubeStateMetricsCore.enabled }}
   kubernetes_state_core.yaml.default: |-
+{{- if .Values.datadog.kubeStateMetricsCore.useClusterCheckRunners }}
+    cluster_checks: true
```
ahmed-mez, line 18: we also need to configure `skip_leader_election: true` in the instance field below when it's a cluster check. (see internal config for ref)

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -179,7 +179,7 @@ spec:
           {{- if .Values.datadog.leaderLeaseDuration }}
           - name: DD_LEADER_LEASE_DURATION
             value: {{ .Values.datadog.leaderLeaseDuration | quote }}
-          {{- else if .Values.datadog.clusterChecks.enabled }}
+          {{- else if (eq (include "should-enable-cluster-check-workers" .) "true") }}
```
ahmed-mez, line 182:
```suggestion
          {{- else if (eq (include "should-enable-cluster-check-runners" .) "true") }}
```

charts/datadog/templates/cluster-agent-confd-configmap.yaml
danemiller, line 18: typo? Should be `cluster_check` right? https://github.com/DataDog/helm-charts/issues/460 @CharlyF

clamoriniere: yes it is a typo.

ahmed-mez: fix https://github.com/DataDog/helm-charts/pull/462 (turned out to be more than just a type 😅 )

PR #434, opened 2021-11-02
Add `agents.localService` parameters

### LINKED ISSUE #429: Restore k8s service name override support
v2.22.8 (https://github.com/DataDog/helm-charts/pull/375/files) removed the `serviceTopology.enabled` and `serviceTopology.serviceName` values. This breaks statsd and APM data collection in deployments with `serviceTopology.serviceName` set. In such deployments, the apps have their `DD_AGENT_HOST` set to the corresponding service name. The helm chart v2.22.8+ doesn't offer an API to override the service name. Also, there is no way to opt out of a service declaration since that change.

I suggest we:
1. Bring the `serviceTopology.*` keys back, interpret them reasonably, and mark them as deprecated
2. Introduce a new value for overriding the service name
3. Introduce a new value for opting out of the service resource declaration

L3n41c:

1. Add an option to set the service name use to communicate with the node agent.
2. Add an option to force the activation of internalTrafficPolicy to support Kubernetes version on which the feature is still alpha.

```yaml
agents:
  localService:
    overrideName: ""
    forceLocalServiceEnabled: false
```

  - fixes #429

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+# 2.23.3
+
+* Add `agents.localService` parameters to customize the internal traffic policy service name and force its creation of Kubernetes 1.21.
+
 # 2.23.2
 
 * Add an `agents.podSecurity.defaultApparmor` setting to allow customizing the default AppArmor profile used by all containers but `system-probe`.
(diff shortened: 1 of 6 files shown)


vboulineau (approved):

hi-artem:
@L3n41c @vboulineau is the value in in the `values.yaml` misspelled and missing d at the end?

Issue #264, opened 2021-05-26
Constant stream of "Error running check prometheus" errors by agent pods


lapidus79:
**Describe what happened:**

Keep getting a lot of error output when running Datadog Helm chart 2.14.0 in AWS EKS 1.19

**Describe what you expected:**

No error ouput.

**Steps to reproduce the issue:**

**Additional environment details (Operating System, Cloud provider, etc):**

```
2021-05-25 10:08:37 UTC | CORE | ERROR | (pkg/collector/runner/runner.go:292 in work) | Error running check prometheus: [{"message": "HTTPConnectionPool(host='REMOVED', port=5000): Max retries exceeded with url: /metrics (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fa3198431f0>: Failed to establish a new connection: [Errno 111] Connection refused'))", "traceback": "Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/urllib3/connection.py\", line 159, in _new_conn
    conn = connection.create_connection(
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/urllib3/util/connection.py\", line 84, in create_connection
    raise err
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/urllib3/util/connection.py\", line 74, in create_connection
    sock.connect(sa)
ConnectionRefusedError: [Errno 111] Connection refused

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/urllib3/connectionpool.py\", line 670, in urlopen
    httplib_response = self._make_request(
  File \"/opt/datadog-agent/embedded/lib/python3.8/site-packages/urllib3/connectionpool.py\", line 392, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File \"/opt/datadog-agent/embedded/lib/python3.8/http/client.py\", line 1255, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File \"/opt/datadog-agent/embedded/lib/python3.8/http/client.py\", line 1301, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File \"/opt/datadog-agent/embedded/lib/python3.8/http/client.py\", line 1250, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
[...]

PR #435, opened 2021-11-03
feat/fix: defined expvarPort, set containerPort for agentmetrics

sebbrandt87:
**What this PR does / why we need it:**
feat/fix: defined expvarPort, set containerPort for agentmetrics

* Add `datadog.expvarPort` parameter to customize the default 
expvar default port to not conflict with the default clusteragent 
metrics port.

* Defined cluster-agent containerPort `agentmetrics` to expose the 
default port, which is set to 5000 and already defined in the 
`NetworkPolicy` for the cluster-agent.

**Special notes for your reviewer:**
See also 
https://github.com/DataDog/datadog-agent/blob/4f63274038167c7487aa44e0126d42cd6b27fba0/pkg/config/config.go#L659-L660
https://github.com/DataDog/datadog-agent/blob/4f63274038167c7487aa44e0126d42cd6b27fba0/pkg/config/config.go#L576-L585

Closes: #264

**Checklist**
- [x] Chart Version bumped
- [x] CHANGELOG.md has beed updated

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Datadog changelog
 
+# 2.23.6
+
+* Add `datadog.expvarPort` parameter to customize the default expvar default port to not conflict with the default clusteragent metrics port if running in hostNetwork mode.
+* Defined cluster-agent containerPort `agentmetrics` to expose the default port, which is set to 5000 and already defined in the `NetworkPolicy` for the cluster-agent.
+
 # 2.23.5
 
 Change OpenShift SCC priorities from 10 to 8 to avoid conflicts with OpenShift Auth operator.
@@ -68,7 +73,7 @@ Change OpenShift SCC priorities from 10 to 8 to avoid conflicts with OpenShift A
 ## 2.22.8
 
 * Add a service with local [internal traffic policy](https://kubernetes.io/docs/concepts/services-networking/service-traffic-policy/) for traces and dogstatsd.
-  This works only on Kubernetes 1.22 or more recent. 
+  This works only on Kubernetes 1.22 or more recent.
 
 ## 2.22.7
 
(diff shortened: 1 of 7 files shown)


juliogreff:
Hi @sebbrandt87, thanks for the contribution! Exposing the `agentmetrics` port seems to be sufficient to solve the mentioned issue of the agent not being able to connect to the cluster agent because of connection refused, and that part of this PR looks good to me.

But it seems that the expvar port change is a no-op for the cluster agent, as it does not open that port, that's just for the agent itself. I tested it by executing `curl http://localhost:6000` from inside the cluster agent pod, and got a connection refused, which for me suggests that indeed there's no such port open even when setting the env var. It's still unclear to me why you were seeing `Error creating expvar server on port 5000: listen tcp 0.0.0.0:5000: bind: address already in use` in the DCA (and that error message is wrong btw, [and it has since been fixed](https://github.com/DataDog/datadog-agent/commit/3601ffa6cd0daed09951e17f301f1c62f95dd905)).

sebbrandt87:
Hey @juliogreff 
it is a pleasure. 😄 

To be exact we see 

```bash
2021-11-08 09:10:30 UTC | CLUSTER | ERROR | (app/app.go:174 in func1) | Error creating telemetry server on port 5000: listen tcp 0.0.0.0:5000: bind: address already in use
```

which matches the current version 😉 .

Okay so we need to check why DD_EXPVAR_PORT is ignored, which shouldn't from what I see in the datadog code.
I take an extra look, could you might take a look also and help seeing what we might not have seen in parsing the environment variables?
Thanks in advance mate.

sebbrandt87:
Okay I think I got it.
It is related to having the port not bound within the container ... but bound on the host already.
I will change my PR accordingly. An expvar server isn't even running within the cluster-agent but in the agent(s).

juliogreff (approved):

Issue #409, opened 2021-10-06
DaemonSet keeps replicas as Pending on GKE Autopilot


tnovau:
**Describe what happened:**
I'm trying to deploy Datadog on GKE Autopilot cluster. I have followed [this tutorial](https://www.datadoghq.com/blog/gke-autopilot-monitoring/#deploy-datadog-on-gke-autopilot). 

When I install the helm chart I'm getting DaemonSet to have less replicas than expected.

**Describe what you expected:**
I expect all the deployed workloads to be green 🟢 

**Steps to reproduce the issue:**
1- Create a K8S Autopilot cluster in GKE (Region: us-west2)
2- Open CloudShell
3- Follow [the tutorial](https://www.datadoghq.com/blog/gke-autopilot-monitoring/#deploy-datadog-on-gke-autopilot)
4- Get this error: 
![image](https://user-images.githubusercontent.com/18268934/136170122-94b07256-f9c7-46bf-a1a1-3eff7d5d3dd0.png)

**Additional environment details (Operating System, Cloud provider, etc):**
Cloud provider: Google (GCP)
Region: us-west2
Release channel: Regular channel 
Version: 1.20.10-gke.301
[Chart](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml) we're using:

__Command and output__ (helm install)
```
helm install datadog-autopilot-test-qa \
>     --set datadog.apiKey=<OMITTED_API_KEY> \
>     --set datadog.appKey=<OMITTED_APP_KEY> \
>     --set clusterAgent.enabled=true \
>     --set clusterAgent.metricsProvider.enabled=true \
>     --set providers.gke.autopilot=true \
>     --set datadog.logs.enabled=true \
>     --set datadog.apm.enabled=true \
>     --set datadog.kubeStateMetricsEnabled=false \
>     --set datadog.kubeStateMetricsCore.enabled=true \
>     datadog/datadog
W1006 08:46:46.419177     675 warnings.go:70] Autopilot set default resource requests for DaemonSet default/datadog-autopilot-test-qa, as resource requests were not specified. See http://g.co/gke/autopilot-defaults.
W1006 08:46:46.670791     675 warnings.go:70] Autopilot set default resource requests for Deployment default/datadog-autopilot-test-qa-cluster-agent, as resource requests were not specified. See http://g.co/gke/autopilot-defaults.
NAME: datadog-autopilot-test-qa
[...]

PR #463, opened 2021-12-09
Adding priorityClass for the DD daemonset in Autopilot environments

CharlyF:

In Autopilot environments, the default NG is too small to accommodate applications and the agent in certain conditions. This ensures the agent is prioritized and let's Autopilot do the scale up as necessary.

https://github.com/DataDog/helm-charts/issues/409

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/agent-priorityclass.yaml
+++ b/charts/datadog/templates/agent-priorityclass.yaml
@@ -1,0 +1,9 @@
+{{- if .Values.agents.priorityClassCreate}}
+apiVersion: scheduling.k8s.io/v1
+description: Used for Datadog Agent Components to be scheduled with higher priority.
+kind: PriorityClass
+metadata:
+  name: {{ .Values.agents.priorityClassName | default (include "datadog.fullname" . ) }}
+preemptionPolicy: PreemptLowerPriority
+value: {{ .Values.agents.priorityClassValue }}
+{{- end }}
--- a/charts/datadog/templates/daemonset.yaml
+++ b/charts/datadog/templates/daemonset.yaml
@@ -99,8 +99,8 @@ spec:
       imagePullSecrets:
 {{ toYaml .Values.agents.image.pullSecrets | indent 8 }}
       {{- end }}
-      {{- if .Values.agents.priorityClassName }}
-      priorityClassName: {{ .Values.agents.priorityClassName }}
+      {{- if or .Values.agents.priorityClassCreate .Values.agents.priorityClassName }}
+      priorityClassName: {{ .Values.agents.priorityClassName | default (include "datadog.fullname" . ) }}
       {{- end }}
       containers:
         {{- include "container-agent" . | nindent 6 }}
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -1135,8 +1135,15 @@ agents:
     rollingUpdate:
       maxUnavailable: "10%"
 
-  # agents.priorityClassName -- Sets PriorityClassName if defineds
+  # agents.priorityClassCreate -- Creates a priorityClass for the Datadog Agent's Daemonset pods.
+  priorityClassCreate: false
+
+  # agents.priorityClassName -- Sets PriorityClassName if defined
   priorityClassName:
+
+  # agents.priorityClassValue -- Value used to specify the priority of the scheduling of Datadog Agent's Daemonset pods.
+  ## The PriorityClass uses PreemptLowerPriority.
+  priorityClassValue: 1000000000
 
   # agents.podLabels -- Sets podLabels if defined
   # Note: These labels are also used as label selectors so they are immutable.
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.28.0
+
+* Create priority Class to better support environments such as GKE Autopilot.
+
 ## 2.27.8
 
 * Default Datadog Agent image to `7.32.3`.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.27.8
+version: 2.28.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.27.8](https://img.shields.io/badge/Version-2.27.8-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 2.28.0](https://img.shields.io/badge/Version-2.28.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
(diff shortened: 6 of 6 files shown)


charts/datadog/values.yaml
```diff
@@ -1135,9 +1135,13 @@ agents:
     rollingUpdate:
       maxUnavailable: "10%"
 
-  # agents.priorityClassName -- Sets PriorityClassName if defineds
+  # agents.priorityClassName -- Sets PriorityClassName if defined
   priorityClassName:
 
+  # agents.priorityClassValue -- Value used to specify the priority of the scheduling of Datadog Agent's Daemonset pods.
+  ## The PriorityClass uses PreemptLowerPriority.
+  priorityClassValue: 1000000000
```
clamoriniere, line 1143: the value seems huge, does the kubernetes documentation have some advice on the possible value? Like from 200000000 to 100000000 for CNI/system pods

charts/datadog/templates/agent-priorityclass.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if or .Values.providers.gke.autopilot .Values.datadog.priorityClassName }}
```
clamoriniere, line 1: IMO it breaks the previous behavior.
.Values.datadog.priorityClassName was use to provide the name of PriorityClass that already exist.

Now it will try to create one with this name.

clamoriniere: The other approach that I see:
Always create the PriorityClass with a default name. If a user provide a Name it means that the PriorityClass already exist

charts/datadog/templates/agent-priorityclass.yaml
```diff
@@ -0,0 +1,9 @@
...
+apiVersion: scheduling.k8s.io/v1
+description: Used for Datadog Agent Components to be scheduled with higher priority.
+kind: PriorityClass
+metadata:
+  name: {{ default "datadog-agent-priority-class" .Values.agents.priorityClassName }}
```
clamoriniere, line 6: The name should contains the release name, else it will not work with 2 charts deployment in the same namespace

charts/datadog/templates/daemonset.yaml
```diff
@@ -99,8 +99,8 @@ spec:
       imagePullSecrets:
 {{ toYaml .Values.agents.image.pullSecrets | indent 8 }}
       {{- end }}
-      {{- if .Values.agents.priorityClassName }}
-      priorityClassName: {{ .Values.agents.priorityClassName }}
+      {{- if or .Values.providers.gke.autopilot .Values.agents.priorityClassName }}
```
clamoriniere, line 102: IMO we can create the PriorityClass all the time.

charts/datadog/values.yaml
CharlyF: I did not find doc, but in Autopilot specifically the value is lower than the value in the priority classes for system pods (critical-cluster and critical-node), which is 2000000000. As the priorityClass has the config to preempt the lower priority it means that we will be scheduled after the kube-system apps but before the applications (or anything < 1000000000).

charts/datadog/templates/agent-priorityclass.yaml
CharlyF: Ah I had misunderstood. I thought this was to specify the name of the PC to be created.
thanks, I will adjust.

charts/datadog/templates/agent-priorityclass.yaml
CharlyF: good point

charts/datadog/templates/daemonset.yaml
CharlyF: my fear if we create the PC all the time is that it's not necessary in most cases and I can foresee users coming back to us having not seen that it's now a default and whether they have one already or have a lifecycle that depends on it for their applications, this could cause some inconvenience.
As it is an advanced feature, I believe we should keep it optional.

clamoriniere: Yes fare enough.
in that case, we can add a new parameter `agent.createPriorityClass: false`

charts/datadog/templates/agent-priorityclass.yaml
clamoriniere: maybe you can create an help for the name. since it will be used in 2 places

charts/datadog/templates/agent-priorityclass.yaml
clamoriniere: linked to my other comment

```suggestion
{{- if or .Values.providers.gke.autopilot .Values.agent.createPriorityClass }}
```

CharlyF:
with `--set providers.gke.autopilot=true` and:
```
agents:
  priorityClassName: datadog-agent-default-nodegroup-autopilot
```
We don't create a PC and we specify:
```
# Source: datadog/templates/daemonset.yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: datadog-autopilot
  namespace: default
[...]
    spec:
      priorityClassName: datadog-agent-default-nodegroup-autopilot
      containers:
      - name: agent
        image: "gcr.io/datadoghq/agent:7.32.1"
```
without `priorityClassName` and with autopilot:
 ```
# Source: datadog/templates/agent-priorityclass.yaml
apiVersion: scheduling.k8s.io/v1
description: Used for Datadog Agent Components to be scheduled with higher priority.
kind: PriorityClass
metadata:
  name: datadog-autopilot
preemptionPolicy: PreemptLowerPriority
value: 1e+09
[...]
# Source: datadog/templates/daemonset.yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: datadog-autopilot
  namespace: default
[...]
spec:
  template:
    spec:
      priorityClassName: datadog-autopilot
```
finally, with  `--set agents.priorityClassCreate=true`
```
# Source: datadog/templates/agent-priorityclass.yaml
apiVersion: scheduling.k8s.io/v1
description: Used for Datadog Agent Components to be scheduled with higher priority.
kind: PriorityClass
metadata:
  name: datadog-autopilot
preemptionPolicy: PreemptLowerPriority
value: 1e+09
[...]
# Source: datadog/templates/daemonset.yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: datadog-autopilot
  namespace: default
[...]
spec:
  template:
    spec:
      priorityClassName: datadog-autopilot
```

File: charts/datadog/templates/agent-priorityclass.yaml [hunk 5]
```diff
@@ -0,0 +1,9 @@
+{{- if and (or .Values.providers.gke.autopilot .Values.agents.priorityClassCreate) (not .Values.agents.priorityClassName)}}
+apiVersion: scheduling.k8s.io/v1
+description: Used for Datadog Agent Components to be scheduled with higher priority.
+kind: PriorityClass
+metadata:
+  name: {{ template "datadog.fullname" . }}
```
clamoriniere, line 6: What if the user want to set a specific name for the pc?
I think you can reuse the logic used in the pod template

charts/datadog/templates/agent-priorityclass.yaml
```diff
@@ -0,0 +1,9 @@
+{{- if .Values.agents.priorityClassCreate}}
+apiVersion: scheduling.k8s.io/v1
+description: Used for Datadog Agent Components to be scheduled with higher priority.
+kind: PriorityClass
+metadata:
+  name: {{ .Values.agents.priorityClassName | default (include "datadog.fullname" . ) }}
```
clamoriniere, line 6: the default name should contains the namespace since `PriorityClass` resources are cluster wide

clamoriniere: hum other cluster wild resource such as clusterrole use also "datadog.fullname". 

So lets consider it here ok

clamoriniere (approved):

Issue #464, opened 2021-12-10
cluster-agent-deployment is missing DD_CLUSTER_CHECKS_ENABLED env variable


danemiller:
I'd like to restrict cluster checks to a dedicated deployment, and disable cluster checks on the daemonset agents.  

When `datadog.clusterChecks.enabled = false`, cluster agent does not schedule any cluster checks.  This is because `DD_CLUSTER_CHECKS_ENABLED` env variable is missing.  Should we instead always set `DD_CLUSTER_CHECKS_ENABLED: "true"` in [cluster-agent-deployment.yaml](https://github.com/DataDog/helm-charts/blob/44a983ca66a14e5189738e2a8f85751dd629b9d4/charts/datadog/templates/cluster-agent-deployment.yaml#L159)?

I expect...
* daemonset: datadog (no cluster checks)
* deployment: datadog-cluster-agent (assign cluster checks to worker/runner pods)
* deployment: datadog-clusterchecks (run cluster checks assigned from cluster-agent)

...from this config:
```
datadog:
  clusterChecks:
    # disable cluster checks for daemonset, but not for clusterchecks deployment, right!?!
    enabled: false
  kubeStateMetricsEnabled: false
  kubeStateMetricsCore:
    enabled: true
    useClusterCheckRunners: true
clusterAgent:
  enabled: true
agents:
  enabled: true
clusterChecksRunner:
  enabled: true
```

ahmed-mez:
Hi @danemiller 

The logic behind will check whether `clusterChecksRunner` is enabled or not before enabling cluster checks in the daemonset, see [here](https://github.com/DataDog/helm-charts/blob/d5e1f4370442bdc5e457468ac7ff0ff943f528d5/charts/datadog/templates/_container-agent.yaml#L81-L89).

To achieve what you want you need to keep `datadog.clusterChecks.enabled=true` and `clusterChecksRunner.enabled=true` and the templating logic will do the rest.

ahmed-mez:
yes.

ahmed-mez:
`datadog.clusterChecks.enabled=true` and `clusterChecksRunner.enabled=true` together would disable cluster checks in the daemonset

danemiller:
oh I see.  indeed I was confused. this is working for me now.  thank you for all your help and patience!

ahmed-mez:
Sure thing, glad it worked :D

PR #465, opened 2021-12-10
cluster-agent-deployment: always enable cluster checks

danemiller:
Not sure I understand the original thinking here.  Proposing this change so that cluster agent is always deployed in an enabled state.  Why would we deploy cluster-agent in a disabled state?

Related to issue https://github.com/DataDog/helm-charts/issues/464


ahmed-mez:
Hi @danemiller 

`Cluster Checks` is one one of multiple features in the Cluster Agent. If cluster checks are globally disabled there is no need to enable them in the cluster agent, it would result in an unnecessary overhead.

danemiller:
ok thanks. I'll continue my questions in the issue and close this pr.

PR #488, opened 2021-12-20
fix(datadog): environments variables quoted correctly

sebbrandt87:
fix(datadog): environments variables quoted correctly

Fix environment variables to be quoted with a loop and `quote` instead of
`toYaml`.

**Checklist**

- [x] Chart Version bumped
- [x] CHANGELOG.md has been updated

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.28.7
+
+* Fix environment variables to be quoted correct with a loop and `quote` instead of `toYaml`.
+
 ## 2.28.6
 
 * Update `PodDisruptionBudget` api version to get rid of `policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget` warning.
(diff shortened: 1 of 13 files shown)


vboulineau:
Hello @sebbrandt87,

Thanks for your contribution, however this implementation misses the handling of `valueFrom`, that's why it currently uses `toYaml`.
I think you can check for the presence of `valueFrom` and pass it through in this case.

sebbrandt87:
Hey @vboulineau,

I added the changes, which should be now sufficient again for `valueFrom`.
Thanks for the hint! 👍

vboulineau:
I'm thinking maybe the logic should go in a helper and then we indent the output (or the helper handles the indent). If you're willing to implement that, it'd nice otherwise we can merge as-is

sebbrandt87:
Gimme some more minutes. I should be able to do it. 😄

sebbrandt87:
@vboulineau @clamoriniere wdyt 😉

sebbrandt87:
I did again a rebase to match with latest main branch changes.

clamoriniere (approved):
Thanks @sebbrandt87 for this contribution 👏 

It is a great added value 👍

PR #494, opened 2021-12-23
Update PodDisruptionBudget api version to get rid of warning message

### LINKED ISSUE #436: Avoid deprecated API versions on YAML definitions for Kubernetes 1.21
**Describe what happened:**

When using Kubernetes 1.21 we get these warnings all the time:

```
policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
```

```
batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

This is because the APIs above have now migrated out of beta.

Expected is on older Kubernetes versions it uses the v1beta1 APIs, but on Kubernetes 1.21+ it uses the new stable versions to avoid the warnings messages.

It might be possible to this using the Helm function to check API versions available `Capabilities.APIVersions.Has`.
It might be worth trying this function to see if it will work?

Run the helm chart on 1.21 and look at the logs.

wangshu3000:

Update PodDisruptionBudget api version from `policy/v1beta1 `to `policy/v1`, to get rid of the warning message, the v1beta1 is deprecated in v1.21+, unavailable in v1.25+;

  - fixes #436

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.7.7
+
+* Update `PodDisruptionBudget` api version to get rid of `policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget` warning.
+
 ## 0.7.6
 
 * Nothing
--- a/charts/datadog-operator/Chart.yaml
+++ b/charts/datadog-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: datadog-operator
-version: 0.7.6
+version: 0.7.7
 appVersion: 0.7.2
 description: Datadog Operator
 keywords:
--- a/charts/datadog-operator/README.md
+++ b/charts/datadog-operator/README.md
@@ -1,6 +1,6 @@
 # Datadog Operator
 
-![Version: 0.7.6](https://img.shields.io/badge/Version-0.7.6-informational?style=flat-square) ![AppVersion: 0.7.2](https://img.shields.io/badge/AppVersion-0.7.2-informational?style=flat-square)
+![Version: 0.7.7](https://img.shields.io/badge/Version-0.7.7-informational?style=flat-square) ![AppVersion: 0.7.2](https://img.shields.io/badge/AppVersion-0.7.2-informational?style=flat-square)
 
 ## Values
 
--- a/charts/datadog-operator/templates/pod_disruption_budget.yaml
+++ b/charts/datadog-operator/templates/pod_disruption_budget.yaml
@@ -1,5 +1,9 @@
 {{- if gt .Values.replicaCount 1.0 -}}
+{{- if .Capabilities.APIVersions.Has "policy/v1" }}
+apiVersion: policy/v1
+{{- else }}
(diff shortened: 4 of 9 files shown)


clamoriniere (approved):
Hi @wangshu3000 

Thanks for this contribution, everything looks good

rslinckx:
We are using helm to install datadog into our cluster, but we use the helm template feature (part of a gitops setup). As such we are limited in what kind of capabilities the helm command will accurately report, as it does not query the live cluster.

There is an easy way to add capabilities via a command line flag, but there is currently no way to remove capabilities from the default list, which is built-in to helm based on their current k8s library version. 

In this case with the latest helm 3 release, the policy/v1 capability seems baked in, and I see no way to disable the capability. As such the helm chart will render a policy/v1 PodDisruptionBudget, which isn't available in my 1.19 cluster.

Would you be open to adding a more explicit flag to select the version of the disruptionbudget to be used in values.yaml ?

clamoriniere:
Hi @rslinckx 
You can set the kubernetes api-server version when running “helm template”

```
helm template --kube-version 1.19.0 -f values.yaml datadog/datadog
```

We are using this option to validate our template against several kubernetes versions see: https://github.com/DataDog/helm-charts/blob/f1820b74bc1ec647c544b26f6d69e1ef01ce9974/.github/kubeval.sh#L23

Please let us know if it solves your issue? 

Else we can discuss how we can provide an option in the values.yaml to for a specific version.

rslinckx:
We are actually running this with helm 3.7.2:

`helm template --api-versions apps/v1 --api-versions apiextensions.k8s.io/v1/CustomResourceDefinition --kube-version 1.19 "$chartName" . --namespace "$chartNamespace" --include-crds --no-hooks`

but it generates policy/v1 PodDisruptionBudget instead of the expected policy/v1beta1 as 1.19 does not have v1 yet.

I think this is due to helm having some kind of hard-coded k8s versions list based on the version they build against?

clamoriniere:
Hello again @rslinckx 
I was able to reproduce your issue too. And I found several open issue on other helm charts reporting the same issue with `policy/v1` (https://github.com/argoproj/argo-cd/issues/7291, https://github.com/cortexproject/cortex-helm-chart/pull/277, https://github.com/elastic/helm-charts/issues/1454)

I have created this PR https://github.com/DataDog/helm-charts/pull/498 to be more restrictive and fix the issue with `helm template`

rslinckx:
Thanks a lot !

mscanlon72:
I am using EKS v1.21 and I still see the warning messages using the latest charts (datadog-2.28.13). Any information on why that would be the case?

wangshu3000:
@mscanlon72  
Did you see the warning when running 
`kubectl get PodDisruptionBudget`
or saw the warning when deploying the chart?
I think you may still see a warning when you run kubectl get PodDisruptionBudget. 
Kubernetes 1.21 still support policy/v1beta1, it'll be removed after 1.25.

mscanlon72:
This is what I see when I run that command:

`No resources found in datadog namespace.`

mscanlon72:
Setting `createPodDisruptionBudget` to true creates the PDB, but the warnings still persist in the logs and are very frequent.

PR #496, opened 2021-12-23
Replace `status.hostIP` by `spec.nodeName` as default kubelet host to be used by the agent

sadok-f:
We have a couple of k8s clusters running on the 1.19 version (on AWS, provisioned via kops), Datadog agent running perfectly fine there. however we have a new cluster version 1.21 and the dd-agent are throwing a lot of errors because unable to connect to the kubelet server running on the host node.

After investigation, it turns out the read-only server running in the Kubelet was deactivated by default since 1.20, and I assume dd-agent switch the secure server instead if we pass the option `datadog.kubelet.tlsVerify=false` to the agent then it starts working fine.
Furthermore if we change the `datadog.kubelet.host` to use `{"valueFrom":{"fieldRef":{"fieldPath":"spec.nodeName"}}}` instead of `{"valueFrom":{"fieldRef":{"fieldPath":"status.hostIP"}}}` this also worked better and clean way than `tlsVerify=false`.

Here is the description of the new configs loaded to kubelet where `readOnlyPort` was deactivated by default: 
https://v1-20.docs.kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

I hope this is clear enough for the issue we faced and the fix we found.
Thanks for the great effort here.

#### Checklist
- [x] Chart Version bumped


sadok-f:
I did a rebase from upstream main since my branch was 2 weeks old, probably that's why it includes all those commits, but the file changed should be clean to review.
thanks.

sadok-f:
I'll create a new PR after resolving the conflicts and bump the version.

PR #497, opened 2021-12-24
[synthetics-pl] upgrade chart version

KevinYeramian:

- Upgrade chart PL version to 1.18.0

### DIFF
--- a/charts/synthetics-private-location/Chart.yaml
+++ b/charts/synthetics-private-location/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: synthetics-private-location
 version: 0.10.0
-appVersion: 1.17.0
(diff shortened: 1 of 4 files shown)


clamoriniere (approved):

clamoriniere (changes requested):
You need to run the script to update the doc

m-rousse (approved):
Ok from synthetics PoV

charts/synthetics-private-location/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: synthetics-private-location
 version: 0.10.0
```
clamoriniere, line 3: You need to bump this version too

clamoriniere (approved):

PR #501, opened 2021-12-27
Update example mysql confs

aweaver89:
Update config examples for the mysql integration to match the current values accepted by the integration

Updates the examples of the mysql integration to correspond to the current values. For example replace the server parameter with the current host that is in the integration example. Also removed the "" around the server and password. This will hopefully help avoid customers using these in their setups and having issues.

#### Checklist

- [x] Chart Version bumped
- [x] CHANGELOG.md has been updated

### DIFF
--- a/charts/datadog/templates/cluster-agent-scc.yaml
+++ b/charts/datadog/templates/cluster-agent-scc.yaml
@@ -17,8 +17,6 @@ allowHostIPC: false
 allowHostPID: false
 allowPrivilegeEscalation: false
 allowPrivilegedContainer: false
-allowedCapabilities: null
-defaultAddCapabilities: null
 fsGroup:
   type: MustRunAs
 readOnlyRootFilesystem: false
--- a/charts/datadog/templates/cluster-agent-scc.yaml
+++ b/charts/datadog/templates/cluster-agent-scc.yaml
@@ -17,6 +17,8 @@ allowHostIPC: false
 allowHostPID: false
 allowPrivilegeEscalation: false
 allowPrivilegedContainer: false
+allowedCapabilities: null
+defaultAddCapabilities: null
 fsGroup:
   type: MustRunAs
 readOnlyRootFilesystem: false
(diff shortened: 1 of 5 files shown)


clamoriniere (commented):
Hi @aweaver89 
thanks for opening this Pull request.

Could you please follow the pull request check list so we can validate and merge it. 🙇

aweaver89:
@clamoriniere - I've removed some of the other checks because they don't seem to be relevant since this is more of a documentation update. Do I need to update the changelog for these minor changes? The version bump seems to be mandatory for the CI tests, but this is really minor imho.

clamoriniere:
The CI tooling is not able to differentiate fix,feature against documentation Pull Request. So unfortunately even for documentation we need to follow the same PR checklist.

aweaver89:
@clamoriniere - Everything passed this time 🥳

charts/datadog/templates/cluster-agent-scc.yaml
```diff
@@ -17,6 +17,8 @@ allowHostIPC: false
 allowHostPID: false
 allowPrivilegeEscalation: false
 allowPrivilegedContainer: false
+allowedCapabilities: null
+defaultAddCapabilities: null
```
clamoriniere, line 21: This 2 lines should not be added back

clamoriniere (changes requested):
I think you have a merged when you have rebased your pr

charts/datadog/templates/cluster-agent-scc.yaml
aweaver89: ahhhh I thought that you had added those. It was a bit weird with your merge earlier today ;)

aweaver89: @clamoriniere - fixed this now 👍

clamoriniere (approved):

clamoriniere (approved):

PR #502, opened 2021-12-27
[synthetics] Upgrade chart PL version to 1.18.1

KevinYeramian:
What this PR does / why we need it:
Upgrade chart PL version to 1.18.1

### DIFF
--- a/charts/synthetics-private-location/Chart.yaml
+++ b/charts/synthetics-private-location/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: synthetics-private-location
-version: 0.11.0
-appVersion: 1.18.0
+version: 0.12.0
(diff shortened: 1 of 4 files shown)


charts/synthetics-private-location/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: synthetics-private-location
-version: 0.11.0
-appVersion: 1.18.0
+version: 0.12.0
```
clamoriniere, line 3: IMO `0.11.1` should be use in that case.

clamoriniere (commented):
small comment about the version used. otherwise looks good

clamoriniere (approved):

m-rousse (approved):

PR #509, opened 2022-01-06
Do not declare the volumes for `/etc/*-release` if there is no `system-probe`

L3n41c:

Do not declare the volumes for `/etc/*-release` if there is no `system-probe`.
Only the `system-probe` container mounts them.

#### Checklist
- [X] Chart Version bumped


clamoriniere (approved):
thanks for the quick fix

PR #514, opened 2022-01-12
Process Agent: Replace deprecated arg "-config" with supported arg "--config"

gaffneyd4:

This PR addresses the following warning log line from the process agent:
```
datadog-thxjm process-agent WARNING: `-config` argument is deprecated and will be removed in a future version. Please use `--config` instead.
```

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-process-agent.yaml
+++ b/charts/datadog/templates/_container-process-agent.yaml
@@ -3,7 +3,7 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["process-agent", "-config={{ template "datadog.confPath" . }}/datadog.yaml"]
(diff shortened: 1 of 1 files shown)


clamoriniere:
Hi @gaffneyd4 

Thank you for opening this pull request, and sorry that this warning is display in the process-agent logs.

But we can make this change yet for chart backward compatibility. Datadog chart v2.0.0 should be compatible with all agent  v6 and v7. But the new process-agent `--config` argument was introduced in agent `v7.31.0` so it is to early to consider removing the chart compatibility with agent older than `7.31.0`.

Our plan to migrate to `--config` is:
* in the next major helm chart version (v3.0.0), we will introduce `--config` option for the process-agent, and define the minimum compatible agent version (possibility v7.31.0).
* in the next agent major version `v8.0.0` we will remove the process-agent `-config` arg

Does that sound good to you?

gaffneyd4:
@clamoriniere that sounds fine! Was just trying to catch some low hanging fruit, but totally understand why the change can't happen yet! This should be good to close.

PR #541, opened 2022-02-09
Replace status.hostIP by spec.nodeName as default kubelet host to be used by the agent

sadok-f:
We have a couple of k8s clusters running on the 1.19 version (on AWS, provisioned via kops), Datadog agent running perfectly fine there. however we have a new cluster version 1.21 and the dd-agent are throwing a lot of errors because unable to connect to the kubelet server running on the host node.

After investigation, it turns out the read-only server running in the Kubelet was deactivated by default since 1.20, and I assume dd-agent switch the secure server instead if we pass the option `datadog.kubelet.tlsVerify=false` to the agent then it starts working fine.
Furthermore if we change the `datadog.kubelet.host` to use `{"valueFrom":{"fieldRef":{"fieldPath":"spec.nodeName"}}}` instead of `{"valueFrom":{"fieldRef":{"fieldPath":"status.hostIP"}}}` this also worked better and clean way than `tlsVerify=false`.

Here is the description of the new configs loaded to kubelet where `readOnlyPort` was deactivated by default: 
https://v1-20.docs.kubernetes.io/docs/reference/config-api/kubelet-config.v1beta1/

I hope this is clear enough for the issue we faced and the fix we found.
Thanks for the great effort here.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.30.6
+
+* Update the default value of `datadog.kubelet.host` to use `spec.nodeName` instead of `status.hostIP`
+
 ## 2.30.5
 
 * Add a new note to recommand to the Cluster Agent in HA mode when the `admission-controller` or the `metrics provider` are enabled.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.30.5
+version: 2.30.6
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.30.5](https://img.shields.io/badge/Version-2.30.5-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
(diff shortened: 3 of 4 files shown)


L3n41c (changes requested):
Hello @sadok-f,

Thank you very much for your valuable feedback.

The reason why the default value for setting `kubelet.host` is to use IP instead of hostname is because using a hostname implies that the agent must do an extra DNS resolution.
In other words, switching to hostname brings into the picture a new component that might fail.
Using IP as it is done today makes the agent more tolerant to DNS disruptions.

The fact that your error can be fixed by:
* either switching from IP (`status.hostIP`) to hostname (`spec.nodeName`) to contact the kubelet or
* setting `datadog.kubelet.tlsVerify` to `false`
makes me think that the error is coming from the fact that the kubelet certificate is valid only for the node hostname and not for its IP.

This is something that is highly dependent on the setup. On GKE for ex., the kubelet certificates are valid both for the node hostname and for its IP.

So, another way to fix the kubelet connectivity issues reported by the agent would be to fix the kubelet certificates to make them valid for their IP.

So, in the end, the default values that have been chosen in the Helm chart are:
* to favor reliability by using IP instead of hostnames and make the agent less sensible to DNS failures;
* to favor security by validating TLS certificates.

Fortunately, those are only default values and, on setups where the kubelet certificates are not valid for node IP, the users have the choice between:
* degrading the reliability by setting `kubelet.host` to `{"valueFrom":{"fieldRef":{"fieldPath":"spec.nodeName"}}}` or
* degrading the security by setting `datadog.kubelet.tlsVerify` to `false`.

But as far as the default value for `kubelet.host` is concerned, I’d rather keep the current `status.hostIP` value to be as robust as possible by default.

PR #542, opened 2022-02-10
[datadog] add disableClusterNameTagKey to values.yaml

programmer04:

This PR adds `disableClusterNameTagKey` to be able easily to use the feature introduced in https://github.com/DataDog/datadog-agent/pull/5500 to get rid of redundant tags.

I think that ideally it should be set to `true` by default for Helm to not use this way for new installations. On the other hand, people may rely on this `cluster_name` tag in their Datadog configuration so updating the Helm Chart version shouldn't break their setup. Hence I decided to have `false` as a default value.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,10 +1,14 @@
 # Datadog changelog
 
+## 2.31.1
+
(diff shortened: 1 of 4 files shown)


programmer04:
Nobody reviewed this, and now it's stale. I'm closing this

PR #548, opened 2022-02-21
[datadog] Add support for Helm check

davidor:

Adds an option in the Datadog chart to enable the helm check (introduced in 7.35: https://github.com/DataDog/datadog-agent/pull/10949)

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_helm_check_config.yaml
+++ b/charts/datadog/templates/_helm_check_config.yaml
@@ -1,0 +1,9 @@
+{{- define "helmCheck-config" -}}
+helm.yaml: |-
+{{- if .Values.datadog.clusterChecks.enabled }}
+  cluster_check: true
+{{- end }}
+  init_config:
+  instances:
+    -
+{{- end -}}
--- a/charts/datadog/templates/cluster-agent-rbac.yaml
+++ b/charts/datadog/templates/cluster-agent-rbac.yaml
@@ -238,6 +238,17 @@ rules:
   resources: ["secrets"]
   verbs: ["get"]
 {{- end }}
+{{- if .Values.datadog.helmCheck.enabled}}
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+{{- end }}
 ---
 apiVersion: {{ template "rbac.apiVersion" . }}
 kind: ClusterRoleBinding
--- a/charts/datadog/templates/cluster-agent-rbac.yaml
+++ b/charts/datadog/templates/cluster-agent-rbac.yaml
@@ -238,17 +238,6 @@ rules:
   resources: ["secrets"]
   verbs: ["get"]
 {{- end }}
-{{- if .Values.datadog.helmCheck.enabled}}
-- apiGroups:
-  - ""
-  resources:
-  - secrets
-  - configmaps
-  verbs:
-  - get
-  - list
-  - watch
-{{- end }}
 ---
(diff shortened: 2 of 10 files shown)


charts/datadog/templates/_helm_check_config.yaml
```diff
@@ -0,0 +1,9 @@
+{{- define "helmCheck-config" -}}
+helm.yaml: |-
+{{- if .Values.datadog.clusterChecks.enabled }}
+  cluster_check: true
+{{- end }}
+  init_config:
+  instances:
```
clamoriniere, line 7: doest the "leader" option is not need

charts/datadog/templates/rbac.yaml
```diff
@@ -109,6 +109,17 @@ rules:
...
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  - configmaps
+  verbs:
+  - get
+  - list
+  - watch
+{{- end }}
```
clamoriniere, line 122: I think this is more for the "clusterCheckRunner"

charts/datadog/templates/cluster-agent-rbac.yaml
```diff
@@ -238,6 +238,17 @@ rules:
   resources: ["secrets"]
   verbs: ["get"]
 {{- end }}
+{{- if .Values.datadog.helmCheck.enabled}}
```
clamoriniere, line 241: IMO, it should only be added if clustercheck is not enabled

```suggestion
{{- if and .Values.datadog.helmCheck.enabled (not .Values.datadog.clusterChecks.enabled}}
```

L3n41c (commented):
I’m wondering if the extra permissions shouldn’t be for the CLC rather than for the node agent.
Also, it would be nice to add the new extra privileges either to only the DCA or only to the CLC.

Although the list of extra-privileges is quite small, it could be interesting to factorise the added section in a dedicated file similar to `kube-state-metrics-core-rbac.yaml` ?
That way, we would have an `rbac` file per core check.

charts/datadog/templates/_helm_check_config.yaml
davidor: I don't think so: https://github.com/DataDog/datadog-agent/pull/10949#discussion_r807058890
But maybe I'm missing something.

charts/datadog/templates/cluster-agent-rbac.yaml
davidor: Yes, that would be better :+1:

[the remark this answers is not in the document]
davidor: You're right. Same as https://github.com/DataDog/helm-charts/pull/548#discussion_r811098355

charts/datadog/templates/rbac.yaml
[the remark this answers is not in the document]
davidor: Instead of defining this on `agent-clusterchecks-rbac.yaml` I think the best solution might be to create a dedicated file similar to `kube-state-metrics-core-rbac.yaml` like you suggested.

davidor: Yes. Same as this comment: https://github.com/DataDog/helm-charts/pull/548#discussion_r811103112

davidor:
@L3n41c @clamoriniere I've added a new commit that extracts the helm check RBAC into its own file.

L3n41c (approved):

clamoriniere (approved):

PR #557, opened 2022-02-28
feat(datadog): hardcode serviceAccount&ConfigMap name on GKE Autopilot

clamoriniere:

From this version, on GKE Autopilot, only one "datadog" helm chart release is allow by Kubernetes namespace due to the following new constraints:
* On GKE autopilot, make the "Agent" Daemonset serviceaccountName immutable.
* On GKE autopilot, make the "Install Info" ConfigMap name immutable.
* Improve `Notes.txt` to better document GKE Autopilot limitation, fail the
  installation when an unsupported option has been used in the configuration.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 2.33.0
 
-***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by the Kubernetes namespace due to the following new constraints:
+***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by Kubernetes namespace due to the following new constraints:
 * On GKE Autopilot, hardcode the "Agent" DaemonSet serviceAccountName.
 * On GKE Autopilot, hardcode the "Install Info" ConfigMap name.
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -3,8 +3,8 @@
 ## 2.33.0
 
 ***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by the Kubernetes namespace due to the following new constraints:
-* On GKE Autopilot, make the "Agent" DaemonSet serviceAccountName immutable.
-* On GKE Autopilot, make the "Install Info" ConfigMap name immutable.
+* On GKE Autopilot, hardcode the "Agent" DaemonSet serviceAccountName.
+* On GKE Autopilot, hardcode the "Install Info" ConfigMap name.
 
 ## 2.32.5
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## 2.33.0
 
-***Warning:*** From this version, on GKE Autopilot, only one "datadog" helm chart release is allow by Kubernetes namespace due to the following new constraints:
-* On GKE autopilot, make the "Agent" Daemonset serviceaccountName immutable.
-* On GKE autopilot, make the "Install Info" ConfigMap name immutable.
+***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by the Kubernetes namespace due to the following new constraints:
+* On GKE Autopilot, make the "Agent" DaemonSet serviceAccountName immutable.
+* On GKE Autopilot, make the "Install Info" ConfigMap name immutable.
 
 ## 2.32.5
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Datadog changelog
 
+## 2.33.0
+
+***Warning:*** From this version, on GKE Autopilot, only one "datadog" helm chart release is allow by Kubernetes namespace due to the following new constraints:
+* On GKE autopilot, make the "Agent" Daemonset serviceaccountName immutable.
+* On GKE autopilot, make the "Install Info" ConfigMap name immutable.
+
 ## 2.32.5
 
 * Fix process detection, by adding `kill` syscall with signal `0` to system-probe seccomp profile.
--- a/charts/datadog/templates/NOTES.txt
+++ b/charts/datadog/templates/NOTES.txt
@@ -211,6 +211,21 @@ To enable it please set clusterAgent.enabled to 'true'.
 {{- end }}
 
 {{- if .Values.providers.gke.autopilot}}
+
+#######################################################################################################
+####               WARNING: only one datadog chart release allows by namespace on GKE autopilot    ####
+#######################################################################################################
+
+{{- if not .Values.agents.rbac.create }}
+
+####################################################################################
+####               WARNING: agents.rbac.create must be true on GKE autopilot    ####
+####################################################################################
+
+{{- fail "On GKE autopilot environment, the option 'agents.rbac.create' must be set to 'true'" }}
+
+{{- end }}
+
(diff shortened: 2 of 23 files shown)


charts/datadog/templates/NOTES.txt
```diff
@@ -219,11 +234,24 @@ To enable it please set clusterAgent.enabled to 'true'.
...
+
+{{- end }}
+
+{{- if .Values.datadog.securityAgent.compliance.enabled }}
+
+#############################################################################################################
+####               WARNING: Cloud Security Posture Management (CSPM) is not supported on GKE Autopilot   ####
+#############################################################################################################
+
+{{- fail "On GKE autopilot environment, Cloud Security Posture Management (CSPM) is not supported. The option 'datadog.securityAgent.compliance.enabled' must be set to 'false'" }}
```
cswatt, line 254:
```suggestion
{{- fail "On GKE autopilot environments, Cloud Security Posture Management (CSPM) is not supported. The option 'datadog.securityAgent.compliance.enabled' must be set to 'false'" }}
```

charts/datadog/templates/NOTES.txt
```diff
@@ -243,6 +271,16 @@ To enable it please set clusterAgent.enabled to 'true'.
 
 {{- end }}
 
+{{- if .Values.datadog.networkMonitoring.enabled }}
+
+#######################################################################################
+####   WARNING: Network Performance Monitoring is not supported on GKE Autopilot   ####
+#######################################################################################
+
+{{- fail "On GKE autopilot environment, Network Performance Monitoring is not supported. The option 'datadog.networkMonitoring.enabled' must be set to 'false'" }}
```
cswatt, line 280:
```suggestion
{{- fail "On GKE Autopilot environments, Network Performance Monitoring is not supported. The option 'datadog.networkMonitoring.enabled' must be set to 'false'" }}
```

cswatt (changes requested):

cswatt (approved):

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Datadog changelog
 
+## 2.31.0
+
+***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by the Kubernetes namespace due to the following new constraints:
+* On GKE Autopilot, make the "Agent" DaemonSet serviceAccountName immutable.
+* On GKE Autopilot, make the "Install Info" ConfigMap name immutable.
```
L3n41c, line 7: “immutable” could mean that the service account name and the config map name cannot be changed once the Helm chart has been installed but that it could still be chosen with different values for several Helm releases.

```suggestion
* On GKE Autopilot, hardcode the "Agent" DaemonSet serviceAccountName.
* On GKE Autopilot, hardcode the "Install Info" ConfigMap name.
```

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Datadog changelog
 
+## 2.31.0
+
+***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by the Kubernetes namespace due to the following new constraints:
```
L3n41c, line 5: Would it be possible to have the hardcoded-named service account be created by a first Helm release and re-used by a second Helm release that would be created with:
```yaml
agents:
  rbac:
    create: false
    serviceAccountName: datadog-agent
```

Technically speaking, is it more correct to say that only one Helm release on GKE auto-pilot can leave `agents.rbac.create` set to its default `true` value ?
I ask this, in the hypothesis where we wouldn’t enforce this restriction with the `fail` statement in the `NOTES.txt`.

charts/datadog/templates/NOTES.txt
```diff
@@ -211,19 +211,47 @@ To enable it please set clusterAgent.enabled to 'true'.
...
+{{- end }}
+
 {{- if eq (include "system-probe-feature" .) "true" }}
 
-##################################################################################
-####               WARNING: System Probe is not supported on GKE Autopilot    ####
-##################################################################################
+#####################################################################
+####   WARNING: System Probe is not supported on GKE Autopilot   ####
+#####################################################################
```
L3n41c, line 233: Is there a specific reason for which:
* If we try to enable the CWS or CSPM on GKE autopilot, it fails whereas
* If we try to enable a `system-probe` feature, it is only a warning.

charts/datadog/templates/_helpers.tpl
```diff
@@ -435,6 +435,26 @@ gke-autopilot
 {{- end -}}
 {{- end -}}
 
+Return the container runtime socket
```
L3n41c, line 438:
```suggestion
Return the service account name
```

charts/datadog/templates/helm-check-rbac.yaml
```diff
@@ -31,7 +31,7 @@ subjects:
     {{- if and .Values.datadog.clusterChecks.enabled .Values.clusterChecksRunner.enabled }}
     name: {{ if .Values.clusterChecksRunner.rbac.create }}{{ template "datadog.fullname" . }}-cluster-checks{{ else }}"{{ .Values.clusterChecksRunner.rbac.serviceAccountName }}"{{ end }}
     {{- else }}
-    name: {{ if .Values.clusterAgent.rbac.create }}{{ template "datadog.fullname" . }}-cluster-agent{{ else }}"{{ .Values.clusterAgent.rbac.serviceAccountName }}"{{ end }}
+    name: {{ include "agents.serviceAccountName" . }}
```
L3n41c, line 34: This was supposed to be service account used by the cluster agent. But here, it is replaced by the service account used by the agents.

L3n41c (commented):
Mostly nits or questions.
The biggest point to double-check is `charts/datadog/templates/helm-check-rbac.yaml` where there is a change replacing the service account of the DCA by the one of the agent.

charts/datadog/templates/helm-check-rbac.yaml
clamoriniere: right thanks for catching this

charts/datadog/templates/NOTES.txt
clamoriniere: it is because, the feature after is disabled silently. I can also put a fail here.

charts/datadog/templates/helm-check-rbac.yaml
clamoriniere:
```suggestion
    name: {{ if .Values.clusterAgent.rbac.create }}{{ template "datadog.fullname" . }}-cluster-agent{{ else }}"{{ .Values.clusterAgent.rbac.serviceAccountName }}"{{ end }}
```

charts/datadog/CHANGELOG.md
CharlyF: Also
```suggestion
***Warning:*** From this version onwards, on GKE Autopilot, only one "datadog" Helm chart release is allowed by Kubernetes namespace due to the following new constraints:
```

clamoriniere: I will keep the restriction as it is because, the serviceAccount in one think but we also have the issue with the install-info configmap that need to be unique.
So I think it is better to say for now `one release per namespace`

charts/datadog/templates/_helpers.tpl
```diff
@@ -435,6 +435,26 @@ gke-autopilot
 {{- end -}}
 {{- end -}}
 
+Return the service account name
+*/}}
```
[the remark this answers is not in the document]
clamoriniere: I forgot to add `{{/*` before it

L3n41c (approved):

clamoriniere:
@CharlyF 
the PR is ready to merge

CharlyF (approved):

PR #562, opened 2022-03-04
FIX Capabilities.APIVersions check on datadog-crds chart

anthonyhaussman:
Actual templates check `.Capabilities.APIVersions` has the "apiextensions.k8s.io/v1/CustomResourceDefinition" resource present.

Via a `helm template` command with a recent version of Kubernetes, the template always returns the `apiextensions.k8s.io/v1beta1` apiVersion.

Only checking if the "apiextensions.k8s.io/v1" version is present seems to be enough to set the correct apiVersion.
Tested on Kubernetes `v1.22.7` with helm `v3.8.0`.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,14 +12,14 @@ jobs:
       charts: ${{ steps.list-changed.outputs.changed }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           fetch-depth: 0
       - name: Set up Helm
         uses: azure/setup-helm@v2.0
         with:
           version: v3.6.3
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v3
         with:
           python-version: 3.7
       - name: Set up chart-testing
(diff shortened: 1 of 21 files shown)


clamoriniere:
hi @anthonyhaussman 
thanks for opening this pull request.

quick question before digging into the code change. Does it still work with older helm version?

anthonyhaussman:
> hi @anthonyhaussman thanks for opening this pull request.
> 
> quick question before digging into the code change. Does it still work with older helm version?

Hi @clamoriniere :wave: 

So I have tested with helm `v2.16.7` / Kubernetes `v1.22.7`.
It still returns me the `apiextensions.k8s.io/v1beta1` apiVersion with the fix.

Tested with the current chat and it returns me  `apiextensions.k8s.io/v1beta1`  too.

Unfortunately, old helm versions do not change the behavior. :confused:

anthonyhaussman:
At reading the Helm Version Support Policy (https://helm.sh/docs/topics/version_skew/), `2.16.x` helm version supports Kubernetes `1.16.x - 1.15.x`.

vboulineau:
@anthonyhaussman You cannot bump `datadog-crd` version in other charts yet as it requires an update to `Chart.lock` which needs to be done once the version is published (two steps required)

anthonyhaussman:
> @anthonyhaussman You cannot bump `datadog-crds` version in other charts yet as it requires an update to `Chart.lock` which needs to be done once the version is published (two steps required)

Alright, I revert the commit.

anthonyhaussman:
@vboulineau and it's done :slightly_smiling_face:

vboulineau (approved):

PR #577, opened 2022-03-22
Enforce use of `root` user for the node agent

L3n41c:

Enforce use of `root` user for the node agent.
This is currently useless and it won’t change anything.
The goal is to be ready for when the agent docker image will have a non-`root` user by default.

The corresponding `datadog-operator` PR is DataDog/datadog-operator#456.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.30.18
+
+* Enforce use of `root` user for the node agent.
+
 ## 2.30.17
 
 * Add `datadog.helmCheck.collectEvents` to enable event collection in the Helm check.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.30.17
+version: 2.30.18
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 2.30.17](https://img.shields.io/badge/Version-2.30.17-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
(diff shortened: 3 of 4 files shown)


jeremiahbowen:
@L3n41c and @clamoriniere
This causes an error on our agent when we deploy this chart:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x28418f7]

goroutine 63 [running]:
github.com/DataDog/datadog-agent/pkg/workloadmeta.(*store).handleEvents(0xc00089a280, 0xc000827000, 0x8c, 0x8c)
	/omnibus/src/datadog-agent/src/github.com/DataDog/datadog-agent/pkg/workloadmeta/store.go:397 +0xb7
github.com/DataDog/datadog-agent/pkg/workloadmeta.(*store).Start.func1(0xc00089a280, 0x550bbe0, 0xc000128010)
	/omnibus/src/datadog-agent/src/github.com/DataDog/datadog-agent/pkg/workloadmeta/store.go:121 +0x185
created by github.com/DataDog/datadog-agent/pkg/workloadmeta.(*store).Start
	/omnibus/src/datadog-agent/src/github.com/DataDog/datadog-agent/pkg/workloadmeta/store.go:114 +0x57
```
What's harder is that if you unset the securityContext: `datadog.securityContext: {}` helm3 does a three way merge and doesn't remove the change. What gets more interesting is that even setting the version in helm `--version 2.30.5` doesn't roll back the change... crimany this was difficult to track down. Basically you have to either edit the daemonset and remove it or delete `--cascade=orphan` and redeploy.

The way I found this update was doing a diff on the yaml output of a new and old pod after pinning the chart version...

L3n41c:
Hello @jeremiahbowen,

Thanks for reporting this issue.
Can you share the version of the agent that you are using as well as a minimal `values.yaml` that reproduces your problem so that we can investigate it deeper?
You can also try to upgrade to `7.34.0` to check if it helps.

The only thing that this PR does is enforcing the use of the `root` user for running the node agents. But without it, it was already the default behaviour for all kubernetes distributions but OpenShift. So, on most setups, this PR should be a no-op.

PR #578, opened 2022-03-23
add support for topologySpreadConstraints in datadog chart

jhohertz:

topologySpreadConstraints are available since 1.19 and allow expression of topology spreading, that is useful to support

No issue seen for topologySpreadConstraints in search of open issues.

I've only done this to the datadog chart, as that's what I am using. If you like the pattern I could apply it to other areas, but I cannot commit to testing those components. Additionally I've added support for the agents daemonset, because it appears to be a supported configuration in kubernetes, but I don't have a use or see where it fits with most uses of daemonsets, so I have not tested using it there.

#### Checklist
- [x] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -1194,6 +1199,11 @@ agents:
   ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
   affinity: {}
 
+  # agents.topologySpreadConstraints -- Allow the DaemonSet to schedule using pod topology spreading
+  ## By default, no constraints are set, allowing cluster defaults to be used for scheduling
+  ## Ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-topology-spread-constraints/
+  topologySpreadConstraints: []
```
vboulineau, line 1205: Is there any real use case for the DaemonSet to use `topologySpreadConstraints`?

jhohertz: Honestly, I can't conceive of a use case for it, but the spec says the field is applicable to DaemonSets, so I included it per the initial note in the PR. I can take it back out if you'd prefer too. Not sure what lint-docs wants, but I'll look into that too.

vboulineau: You need to run `.github/helm-docs.sh` after bumping the chart version.
I think it's indeed better to remove the option, we try to avoid exposing options that could produce unexpected behaviour or could be misleading.

vboulineau:
> @vboulineau I was looking for `topologySpreadConstraints`, ended up here.
> 
> if @jhohertz is not available to resume this PR, and given my never having contributed to this repo, is there anything else that'd need to be done? there's `.github/helm-docs.sh`, there's removal of the option for the `DaemonSet`, anything else? I am quite keen to get that feature, so I could potentially have a go at the PR

No that's pretty much it, nothing else to do!

vboulineau:
Closing in favor of #1020

PR #584, opened 2022-03-28
Fix system-probe integration

vxcodes:
~The problem we were having: The datadog agent, process-agent, and system-probe containers were unable to communicate due to the readOnly value set to true in each for the sys-probe-dir directory. We noticed the issue by this message in our process-agent logs:~

```datadog-lvkft process-agent 2022-01-11 18:36:20 UTC | PROCESS | WARN | (pkg/process/checks/net.go:107 in 	 getConnections) | could not initialize system-probe connection: error setting up remote system probe util, socket path does not exist: stat /var/run/sysprobe/sysprobe.sock: no such file or directory (will only log every 10 minutes)```

~We opened up a ticket with datadog support and they recommended that we update the readOnly value to true for the sys-probe-dir in the agent, process-agent, and system-probe containers and set the DD_SYSTEM_PROBE_EXTERNAL variable such that the process-agent container has access to the external system-probe.~

~After implementing these changes the error message went away and we started getting data out of the system-probe for the universal service monitoring beta, so we believe that the error was contributing to issues with that particular feature before we made this change.~

### Edit
Upon further investigation, we discovered that the issue with the system-probe was due to an accidental setting of `DD_SYSPROBE_SOCKET=/var/run/s6/sysprobe.sock` in the `system-probe` container. Once removing that environment variable override and allowing the default value to replace it (`DD_SYSPROBE_SOCKET=/var/run/sysprobe/sysprobe.sock`), our problem was solved.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-process-agent.yaml
+++ b/charts/datadog/templates/_container-process-agent.yaml
@@ -39,6 +39,10 @@
       value: {{ .Values.agents.containers.processAgent.logLevel | default .Values.datadog.logLevel | quote }}
     - name: DD_SYSTEM_PROBE_ENABLED
       value: {{ .Values.datadog.networkMonitoring.enabled | quote }}
+    {{- if eq (include "should-enable-system-probe" .) "true" }}
+    - name: DD_SYSTEM_PROBE_EXTERNAL
+      value: "true"
+    {{- end }}
     {{- if .Values.datadog.networkMonitoring.enabled }}
     - name: DD_SYSTEM_PROBE_NETWORK_ENABLED
       value: {{ .Values.datadog.networkMonitoring.enabled | quote }}
--- a/charts/datadog/templates/_container-process-agent.yaml
+++ b/charts/datadog/templates/_container-process-agent.yaml
@@ -85,7 +85,7 @@
     {{- if eq (include "should-enable-system-probe" .) "true" }}
     - name: sysprobe-socket-dir
       mountPath: /var/run/sysprobe
-      readOnly: true
+      readOnly: false
     - name: sysprobe-config
       mountPath: /etc/datadog-agent/system-probe.yaml
       subPath: system-probe.yaml
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Datadog changelog
 
+## 2.30.20
+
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: charts/datadog/templates/_container-process-agent.yaml [hunk 1]
```diff
@@ -39,6 +39,10 @@
       value: {{ .Values.agents.containers.processAgent.logLevel | default .Values.datadog.logLevel | quote }}
     - name: DD_SYSTEM_PROBE_ENABLED
       value: {{ .Values.datadog.networkMonitoring.enabled | quote }}
+    {{- if eq (include "should-enable-system-probe" .) "true" }}
+    - name: DD_SYSTEM_PROBE_EXTERNAL
```
brycekahle, line 43: The existence of this env var is for a specific scenario where:
1. The default agent container and *entrypoint* is used, where multiple processes are run in the same container
2. The system-probe is run as a separate container, since it requires higher privileges. This env var instructs the "main" container to not run system-probe (more specifically, system-probe to exit on startup), since it will exist outside of that container.

In the helm chart, the `command` is being specified for each container AND each process has its own container. Thus the default entrypoint is not being used, and there is no danger of a `system-probe` running in the `process-agent` container.

In summary, I don't think this change is necessary for the helm chart. If you think our public documentation is insufficient in some regard, please let us know and we can work on getting that updated/clarified.

vxcodes:
Thanks @L3n41c and @brycekahle for explaining! After looking at the Kubernetes documentation it looks like the default value for readOnly on a volumeMount is false, so specifying it for the system-probe container is not required. Additionally I understand that the agent and process-agent containers do not need the system probe socket directory to be mounted with readOnly: false, so that change is not required as well. Finally I understand that the DD_SYSTEM_PROBE_EXTERNAL env var is only required in a “single agent container + system probe container” configuration (not how this Helm chart is configured), and thus that change is not required as well. I am therefore closing this pull request.

vxcodes:
Thanks @L3n41c and @brycekahle for explaining! After looking at the Kubernetes documentation it looks like the default value for readOnly on a volumeMount is false, so specifying it for the system-probe container is not required. Additionally I understand that the agent and process-agent containers do not need the system probe socket directory to be mounted with readOnly: false, so that change is not required as well. Finally I understand that the DD_SYSTEM_PROBE_EXTERNAL env var is only required in a “single agent container + system probe container” configuration (not how this Helm chart is configured), and thus that change is not required as well. I am therefore closing this pull request.

PR #593, opened 2022-04-07
Remove use of `concat` to restore Helm2 compatibility

vboulineau:
  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -113,7 +113,12 @@ spec:
         {{- end }}
 {{- if or .Values.datadog.envFrom .Values.clusterAgent.envFrom }}
         envFrom:
-{{ concat .Values.datadog.envFrom .Values.clusterAgent.envFrom | toYaml | indent 10 }}
+{{- if .Values.datadog.envFrom }}
+{{ .Values.datadog.envFrom | toYaml | indent 10 }}
+{{- end }}
+{{- if .Values.clusterAgent.envFrom }}
+{{ .Values.clusterAgent.envFrom | toYaml | indent 10 }}
+{{- end }}
 {{- end }}
         env:
           - name: DD_HEALTH_PORT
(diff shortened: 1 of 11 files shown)


charts/datadog/values.yaml
```diff
@@ -2,6 +2,9 @@
 ## See Datadog helm documentation to learn more:
 ## https://docs.datadoghq.com/agent/kubernetes/helm/
 
+## FOR AN EFFORTLESS UPGRADE PATH, DO NOT COPY THIS FILE AS YOUR OWN values.yaml.
+## ONLY SET THE VALUES YOU WANT TO OVERRIDE IN YOUR values.yaml.
```
clamoriniere, line 6: 👍

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -113,7 +113,12 @@ spec:
         {{- end }}
 {{- if or .Values.datadog.envFrom .Values.clusterAgent.envFrom }}
         envFrom:
-{{ concat .Values.datadog.envFrom .Values.clusterAgent.envFrom | toYaml | indent 10 }}
+{{- if .Values.datadog.envFrom }}
```
clamoriniere, line 116: An helper function can be made to avoid this repetitive code

vboulineau: I'm not sure it will help that much. We can only pass one parameter to the helper. In this case it needs to the deepest `envFrom`, so we cannot handle the `.Values.datadog.envFrom` and we cannot handle the `| indent` either.

vboulineau: I'm not sure it will help that much. We can only pass one parameter to the helper. In this case it needs to the deepest `envFrom`, so we cannot handle the `.Values.datadog.envFrom` and we cannot handle the `| indent` either.

clamoriniere: you can pass several argument by creating a temporary `dict`. but maybe it is overkill

PR #617, opened 2022-05-06
initial attempt at adding obs pipeline to datadog chart

kennonkwok:
WIP from @silversupreme and @kennonkwok pairing session

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -1506,3 +1506,128 @@ providers:
       ## IMDS metadata endpoint is not accesible to containers. This flag mounts the host's
       ## `/var/lib/cloud/data/instance-id` and uses that for Agent's hostname instead.
       useHostnameFromFile: false
+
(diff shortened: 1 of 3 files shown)


charts/datadog/values.yaml
```diff
@@ -1506,3 +1506,128 @@ providers:
       ## IMDS metadata endpoint is not accesible to containers. This flag mounts the host's
       ## `/var/lib/cloud/data/instance-id` and uses that for Agent's hostname instead.
       useHostnameFromFile: false
+
+observabilityPipelines:
```
clamoriniere, line 1510: IMO we should not put the full values file in from vector into the datadog chart file. It will be very difficult to maintains.

You can look at what we did for the kube-state-metrics chart dependencies

silversupreme: This is just overriding a few options from the full values file, just the stuff that's needed to properly wire up to the Agent.

PR #619, opened 2022-05-09
[WIP] Add support for OTLP ingest

### LINKED ISSUE #529: Helm chart doesn't support the experimental object
https://docs.datadoghq.com/tracing/setup_overview/open_standards/#otlp-ingest-in-datadog-agent Describes the steps to configure the agent to ingest OLTP.

**Describe what happened:**
Helm chart successfully deploys, but does not pass through the variables passed.

To allow configuration of the experimental config block as part of the helm chart.

`helm repo add datadog https://helm.datadoghq.com  `

`helm install -n logging ddagent --set datadog.apiKey=$datadog_apikey datadog/datadog --set targetSystem=linux --set datadog.dogstatsd.useHostPort=true --set experimental.otlp.receiver.protocols.http.endpoint=0.0.0.0:4317 --create-namespace`

Kubernetes v1.21.7

mx-psi:

Adds support for configuring OTLP ingest. For example, to enable OTLP/gRPC:

```
datadog:
  otlp:
    receiver:
      protocols:
        grpc: 
          endpoint: "0.0.0.0:4317"
```

and to enable OTLP/HTTP:

```
datadog:
  otlp:
    receiver:
      protocols:
        http: 
          endpoint: "0.0.0.0:4318"
```

This is a work in progress!

  - fixes #529

#### Checklist
- [ ] Chart Version bumped


charts/datadog/requirements.lock
```diff
@@ -5,5 +5,5 @@ dependencies:
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
   version: 2.13.2
-digest: sha256:ebed249fa850dd365390e2e43fc82b8d47cedfbc3831d6c81039a7c216474d6b
-generated: "2022-03-09T12:28:39.941176978+01:00"
+digest: sha256:6757563c3d9ab1f942022ad081c7957cf8585415557ebe09c0a6a75f4189375d
+generated: "2022-05-10T11:50:18.477598+02:00"
```
mx-psi, line 9: I think this comes from when I ran `.github/kubeval.sh` locally, should I revert that change?

urseberry (approved):

urseberry (approved):

charts/datadog/values.yaml
```diff
@@ -308,6 +308,20 @@ datadog:
...
+  #  receiver:
+  #    protocols:
+  #      grpc:
+  #        # datadog.otlp.receiver.procotols.grpc.endpoint -- OTLP/gRPC endpoint
+  #        ## Set explicitly to enable OTLP/gRPC for all supported datatypes.
+  #        # endpoint: "0.0.0.0:4317"
+  #      http:
+  #        # datadog.otlp.receiver.procotols.http.endpoint -- OTLP/HTTP endpoint
+  #        ## Set explicitly to enable OTLP/HTTP for all supported datatypes.
+  #        # endpoint: "0.0.0.0:4318"
```
clamoriniere, line 323: IMO we should try to improve the user experience, by providing an easier configuration:
* a boolean to enable it.
* a default configuration.

```suggestion
  otlp: 
    enabled: false
    receiver:
      protocols:
        grpc:
          # datadog.otlp.receiver.procotols.grpc.endpoint -- OTLP/gRPC endpoint
          ## Set explicitly to enable OTLP/gRPC for all supported datatypes.
          # endpoint: "0.0.0.0:4317"
        http:
          # datadog.otlp.receiver.procotols.http.endpoint -- OTLP/HTTP endpoint
          ## Set explicitly to enable OTLP/HTTP for all supported datatypes.
          # endpoint: "0.0.0.0:4318"
```

mx-psi: I guess you would want something like this instead?

```
otlp:
  receiver:
    protocols:
      grpc:
        enabled: false
       http:
         enabled: false
```

Most people would use only one of the protocols for sending telemetry data. This makes sense, the way this works on every other OpenTelemetry product (Datadog or other companies) is to have:

```
otlp:
  receiver:
    protocols:
      grpc:
```

be the "use the default settings for gRPC", but I don't think Helm supports this (or at least, I haven't been able to make it work in any way), so enabled seems like the next best thing.

clamoriniere: IMO if it makes sens we can enable both (grpc, http) by default. And if a customer request provide a way to disable it.


```yaml
otlp:
  enabled: false
  receiver:
    protocols:
      grpc:
        enable: true
       http:
         enable: true
```

or if `endpoint` is empty disable it

```yaml
otlp:
  enabled: false
  receiver:
    protocols:
      grpc:
        endpoint: ""
      http:
        endpoint: ""
```

From our experience, we should provide a configuration as simple as possible.

mx-psi: @clamoriniere I made it work in the same way it works in the Datadog Agent YAML and the OpenTelemetry Collector OTLP receiver configuration. That is, for example:

```
datadog:
  otlp:
     receiver:
       protocols:
          grpc:
```

Will enable the OTLP/gRPC endpoint. I feel like this is simpler (people don't need to set the endpoint) and consistent with what people coming from other vendors or migrating from local Datadog Agent YAML would expect.

Do you feel this is acceptable?

charts/datadog/values.yaml
```diff
@@ -308,6 +308,22 @@ datadog:
     # datadog.apm.hostSocketPath -- Host path to the trace-agent socket
     hostSocketPath: /var/run/datadog/
 
+  # datadog.otlp -- OTLP ingest related configuration
+  ## Set a protocol section to enable its default endpoint.
+  otlp: {}
```
clamoriniere, line 313: I'm still in favour of defining the full configuration structure with default value.

```yaml
datadog:
  otlp:
    enabled: false
    receiver:
      grpc:
        enabled: true
        port: 4317
      http:
        enabled: true
        port: 4318
```

With this approach, the user will only have to set `datadog.otlp.enabled: true` to have it working. which is convenient for user that use the `--set` option with the `helm install` command. (ex: `helm install foo --set datadog.otlp.enabled=true datadog/datadog`

mx-psi: Changed to have an `enabled` option for each protocol.

CharlyF (changes requested):
Thank you very much for your patience, sorry it took so long.
I looked at [a few other charts](https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L30-L35) and no one really uses the `enabled` as we do. 

Let's go with this default:
```
datadog:
  otlp:
    receiver:
      protocols:
        grpc:
         enabled: false
         endpoint: "0.0.0.0:4317"
        http:
          enabled: false
          endpoint: "0.0.0.0:4318"
```

However, it would be really great if we could take advantage of the helpers.
Especially looking at this bit of logic https://github.com/DataDog/helm-charts/pull/619/files#diff-0973c8baa0f902471c709f7be2b0682aa0b89d92368289528c1d4b42118980a3R16-R32
We do not want that in the template of the agent.

Could you move the logic to the _helper.tpl file, similar to this?
https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_helpers.tpl#L56-L64
and update the templates to take advantage of a `{{- template "otlp-receiver-enabled" . }}` or something?

Thanks a lot!

mx-psi:
@CharlyF not sure if my latest change is what you were looking for?

CharlyF (approved):

pkoraca:
Thanks for the MR!

For consistency reasons, shouldn't OTLP ports also be in the [Service](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/agent-services.yaml#L80)?

From what I understood, there are 2 options to connect to the agent:
- via agent `Service`
- via `hostPort` and then exporting `status.hostIP` as env variable to the app. 

Would that make sense?

mx-psi:
@pkoraca Apologies for the late reply. I think your suggestions make sense although my familiarity with this chart is limited. I have added this to our backlog to eventually take a look, thanks for bringing it up.

PR #622, opened 2022-05-13
[synthetics] Update liveness and readiness probes in PL chart

ruizb:

The liveness probe mechanism used by the Synthetics Private Location chart is deprecated. We want to use the HTTP liveness probe instead.

Also, we are adding the HTTP readiness probe, with the possibility to enable/disable both probes using a new value: `enableStatusProbes`.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-private-location/templates/deployment.yaml
+++ b/charts/synthetics-private-location/templates/deployment.yaml
@@ -33,16 +33,22 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.enableStatusProbes }}
           livenessProbe:
-            exec:
-              command:
-                - /bin/sh
-                - -c
-                - '[ $(expr $(cat /tmp/liveness.date) + 300000) -gt $(date +%s%3N) ]'
+            initialDelaySeconds: 30
+            periodSeconds: 60
+            timeoutSeconds: 10
+            httpGet:
+              path: /liveness
+              port: 8080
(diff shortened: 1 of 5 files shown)


jakepruitt (approved):
👍  Looks good to me!

charts/synthetics-private-location/values.yaml
```diff
@@ -104,3 +104,7 @@ hostAliases: []
 #  - ip: "10.0.0.1"
 #    hostnames:
 #    - "host.domain.com"
+
+# enableStatusProbes -- Enable both liveness and readiness probes
+enableStatusProbes: false
```
ahmed-mez, line 109: Does this parameter depend on minimal private location image version to work? if so, let's mention in in the doc string + in the changelog.

charts/synthetics-private-location/templates/deployment.yaml
```diff
@@ -33,16 +33,22 @@ spec:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.enableStatusProbes }}
```
ahmed-mez, line 36: This mean we don't put a liveness probe by default anymore, is this the expected behavior? Or should we do an if/else to keep the old liveness probe when `enableStatusProbes` is disabled?

ruizb: Yes, it is the expected behavior 👍  We don't want to keep the old liveness probe as a fallback because it is deprecated, and it will be removed in the coming weeks/months.

ahmed-mez (approved):

celenechang (approved):

PR #163, opened 2021-02-16
[datadog-operator] Added support for podAnnotations and podLabels

DVN237294:
#### What this PR does / why we need it:
This PR allows for custom annotations and labels to be added to the DataDog operator pods
#### Which issue this PR fixes
No issue was created. I stumbled into this problem trying to add DataDog Agent Autodiscovery annotations to my pods.
#### Special notes for your reviewer:
None.
#### Checklist
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #72, opened 2020-10-28
[charts/datadog] Support additional labels

ahmed-mez:
#### What this PR does / why we need it:

* Support adding labels to the Agent pods and daemonset via `agents.additionalLabels`.
* Support adding labels to the Cluster Agent pods and deployment via `clusterAgent.additionalLabels`.
* Support adding labels to the Cluster Checks Runner pods and deployment via `clusterChecksRunner.additionalLabels`.

#### Which issue this PR fixes
  - fixes #43 

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

PR #645, opened 2022-06-02
allow aritrary pod labels to be set

resnostyle:
Signed-off-by: Bryan Pearson <bpearson@morningconsult.com>

We want to be able to set arbitrary pod labels for kube resources.

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.34.3
+
+* If set allow arbitrary podlabels on pods.
+
 ## 2.34.2
 
 * Default Cluster Agent image to `1.20.0`.
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,8 +1,22 @@
 # Datadog changelog
 
-## 2.34.3
+
+## 2.34.6
 
 * Allow optional arbitrary podlabels on pods by setting the `podLabels` configuration.
+
+## 2.34.5
+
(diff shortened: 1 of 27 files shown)


charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.34.3
+
+* If set allow arbitrary podlabels on pods.
```
kayayarai, line 5:
```suggestion
* If set, allow arbitrary podlabels on pods.
```
Something feels missing. If what is set? If a config option is set (in which case say "If X option is set, allow ...")? Or if arbitrary pod labels are set (in which case say "If arbitrary podlabels are set, allow them on pods" or "Allow optional arbitrary podlabels on pods")?

kayayarai (approved):

charts/datadog/CHANGELOG.md
resnostyle: I've made the requested changes to the changelog.

badouralix (changes requested):
Hi @resnostyle, thank you for your contribution !

Pod labels do not rely on `synthetics-private-location.labels` but rather on `synthetics-private-location.selectorLabels`

https://github.com/DataDog/helm-charts/blob/fbea56da9841c516ef9205d0f68070fe32215dca/charts/synthetics-private-location/templates/deployment.yaml#L12-L19

I suggest you follow a similar patch that was created for the `datadog-operator` chart https://github.com/DataDog/helm-charts/pull/163

Or even better, use `additionalLabels` and add it on all resources, including pods, deployments, configmaps, etc. similarly to https://github.com/DataDog/helm-charts/pull/72

PR #664, opened 2022-06-16
ci: update schema location for chart validation

wyardley:
Despite the note about checking for available schemas, many of the schemas being tested about do not exist at
https://kubernetesjsonschema.dev/, which is no longer updated; this could result in a false sense of confidence (because of `--ignore-missing-schemas`, I believe this will fail soft)

Update to later 1.18.x and 1.22.x variants in the matrix, add 1.24.x, and switch to the schemas at https://github.com/yannh/kubernetes-json-schema (used by `kubeconform`).

While this may seem a little strange, I haven't found a more updated / stable source to use thus far.

#### Checklist

### DIFF
--- a/.github/kubeval.sh
+++ b/.github/kubeval.sh
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 KUBEVAL_VERSION="v0.16.1"
-SCHEMA_LOCATION="https://kubernetesjsonschema.dev/"
+SCHEMA_LOCATION="https://raw.githubusercontent.com/yannh/kubernetes-json-schema/master"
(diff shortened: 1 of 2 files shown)


wyardley:
@clamoriniere - you mind taking a look at this or asking someone else with context on this to take a look? Happy to adjust the specific versions or make other changes as needed.

vboulineau (approved):

wyardley:
Thanks @vboulineau 
btw, didn't see if you triggered it, but hope you were able to test it works Ok? Please do tag me / let me know if this causes any issues now that it's updated!

PR #665, opened 2022-06-17
[synthetics-private-location] add ability to template config name

iorubs:

Ability to supply a template for the configSecret.
* Reference secrets that were created with a template without having to maintain multiple values files.
* Easier to integrate with external secret providers.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-private-location/CHANGELOG.md
+++ b/charts/synthetics-private-location/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+### 0.14.2
+
+* Added ability to template the config secret name.
+
 ### 0.14.1
 
 * Update private location image version to `1.24.0`.
--- a/charts/synthetics-private-location/CHANGELOG.md
+++ b/charts/synthetics-private-location/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 0.14.2
 
-* Added ability to template the config secret name.
+* Add ability to template the ConfigMap/Secret name.
 
 ### 0.14.1
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.35.6
+
+* Fix `include` in clusterchecks deployment template.
+
 ## 2.35.5
 
 * Allow cross-DCA communication in DCA `NetworkPolicy` and `CiliumNetworkPolicy`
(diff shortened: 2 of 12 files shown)


charts/synthetics-private-location/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+### 0.14.2
+
+* Added ability to template the config secret name.
```
badouralix, line 5:
```suggestion
* Add ability to template the config secret name.
```

badouralix (changes requested):
Hi @iorubs, thanks for your contribution ! Overall lgtm, just added one nitpick

Would to be able to make a similar change for `configConfigMap` a few lines above ?

charts/synthetics-private-location/CHANGELOG.md
iorubs: Done.

iorubs:
> Hi @iorubs, thanks for your contribution ! Overall lgtm, just added one nitpick
> 
> Would to be able to make a similar change for `configConfigMap` a few lines above ?

Done.

badouralix (approved):

iorubs:
@badouralix one of the checks is timing out, do you have any recommendation on what I should do to fix/rerun it?

badouralix:
So the chart installation is failing due to the following error, which sort of makes sense since the `synthetics-private-location` chart does not provide the configmap itself and expects it to be created elsewhere

```text
MountVolume.SetUp failed for volume "worker-config" : configmap "config-map-name" not found
```

With the current test setup, creating the configmap/secret before installing the chart is a bit tricky

@iorubs let's remove the new files under `ci/` for now
We will add them back in another pull request once we can handle the configmap/secret creation

iorubs:
> So the chart installation is failing due to the following error, which sort of makes sense since the `synthetics-private-location` chart does not provide the configmap itself and expects it to be created elsewhere
> 
> ```
> MountVolume.SetUp failed for volume "worker-config" : configmap "config-map-name" not found
> ```
> 
> With the current test setup, creating the configmap/secret before installing the chart is a bit tricky
> 
> @iorubs let's remove the new files under `ci/` for now We will add them back in another pull request once we can handle the configmap/secret creation

That makes sense, thanks.

Done.

badouralix (approved):

L3n41c (approved):

PR #671, opened 2022-06-27
Bump datadog default agent image to 7.37.1

jalaziz:

Bumps default datadog agent image version to 7.37.1.

Cluster agent 1.21.0 is not yet released, so that would have to be updated separately.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.36.0
+
+* Default Datadog Agent image to `7.37.1`.
+
 ## 2.35.6
 
 * Fix `include` in clusterchecks deployment template.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.35.6
+version: 2.36.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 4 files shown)


vboulineau:
We've identified a bug in `7.37.0` and will release `7.37.1` shortly, not merging the change until we have it released.

guizmaii:
IIRC, the old "experimental" configurations for the OpenTelemetry listener were supposed to be removed from this `7.37.0` version. Is it the case? Because, IIRC, the new non-experimental configs haven't been merged yet (See https://github.com/DataDog/helm-charts/pull/619)

mx-psi:
> cc @mx-psi Do we need to merge your PR before upgrading to `7.37.1` default?

@vboulineau There is nothing specific in the Helm chart right now about OTLP ingest, so I don't think my PR blocks this PR. The instructions we have on the public docs about this right now (https://docs.datadoghq.com/tracing/setup_overview/open_standards/otlp_ingest_in_the_agent/?tabs=kuberneteshelmvaluesyaml) will work both before and after this PR.

The only difference this PR will make is that the old, experimental way of setting OTLP will stop working, but this has been announced on the changelog and the logs since 7.35 https://github.com/DataDog/datadog-agent/blob/main/CHANGELOG.rst#upgrade-notes-1

jalaziz:
> We've identified a bug in `7.37.0` and will release `7.37.1` shortly, not merging the change until we have it released.

@vboulineau I've updated this PR for 7.37.1

jalaziz:
Looks like DCA 1.21 is preparing for a final release? I can include it in this PR if you want to wait for that as well.

jalaziz:
Looks like a different PR was merge.

PR #679, opened 2022-07-07
Add "tolerations" to kube-state-metrics

keigohtr:
I have added `tolerations` field to kube-state-metrics in values.yaml because it's missing.
In our env, all nodes has taints, so we need to specify some tolerations to all pods.

`tolerations` was already supported in "kube-state-metrics" of v.2.13.2. -> https://github.com/prometheus-community/helm-charts/blob/kube-state-metrics-2.13.2/charts/kube-state-metrics/values.yaml#L101

N/A

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 2.36.2
+
+* Add `tolerations` field to kube-state-metrics in values.yaml.
+
 ## 2.36.1
 
 * Add `datadog.otlp` section to configure OTLP ingest.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.36.1
+version: 2.36.2
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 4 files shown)


clamoriniere (commented):
Hi @keigohtr 

Thanks for opening this Pull Request.

Technically, the change looks good. but I'm not sure it is necessary, because even if the option is not present in the "datadog" `values.yaml` file, it can be set in your `values.yaml` override. In fact any options set under `kube-state-metrics` will be provided to the "kube-state-metrics" chart dependency instantiation.

Do you have a strong needs of adding this option in our chart? 

Regards
Cedric

hestonhoffman (approved):

keigohtr:
Hi @clamoriniere 

Thanks for the review.

> Do you have a strong needs of adding this option in our chart?

Short answer is Yes. I believe it would be nice to have this change for all of us. 

> because even if the option is not present in the "datadog" `values.yaml` file, it can be set in your `values.yaml` override.

Good point. 
Actually I have read this repository and have realized the fact you mentioned above. What I did were

1. checked `datadog/values.yaml` and confirmed the necessary parameters (e.g. `tolerations`).
2. run the helm chart with my `values.yaml`.
3. realized `kube-state-metrics` was `Pending` status.
4. checked this repository and realized this chart calls `kube-state-metrics` chart with overriding parameters. 
5. added `tolerations` to `kube-state-metrics` section.

If there were `tolerations` in `kube-state-metrics` section of `datadog/values.yaml` like `datadog-agent` and other resources's section, I think it would be nice for developers like me who want to set `tolerations`.

keigohtr:
No response. can anyone help?

keigohtr:
I will close this PR because no one responded.

PR #688, opened 2022-07-15
[datadog] Fix usage of deprecated command flags in the process-agent

davidor:

Starting from agent 7.36, we should use `--cfgpath` instead of `--config` (Ref: https://github.com/DataDog/datadog-agent/pull/11328).

Also, we were using `-config` instead of `--config`. `-config` was deprecated in 7.31. (Ref: https://github.com/DataDog/datadog-agent/pull/8964)

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-process-agent.yaml
+++ b/charts/datadog/templates/_container-process-agent.yaml
@@ -3,10 +3,10 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["process-agent", "-config={{ template "datadog.confPath" . }}/datadog.yaml"]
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/_container-process-agent.yaml
```diff
@@ -100,3 +100,17 @@
...
+{{- $length := len (split "." $version ) -}}
+{{- if and (gt $length 1) (not (semverCompare "^6.36.0 || ^7.36.0" $version)) -}}
+--config
+{{- else -}}
+--cfgpath
+{{- end -}}
+{{- else -}}
+--config
+{{- end -}}
+{{- end -}}
```
ahmed-mez, line 116: we usually put these helpers in https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_helpers.tpl

ahmed-mez (approved):

charts/datadog/templates/_container-process-agent.yaml
davidor: Moved :+1:

PR #706, opened 2022-08-02
DD_APM_ENABLED parametrization

leszek-niewiadomski:
DD_APM_ENABLED variable was hardcoded to false.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -92,7 +92,7 @@
     {{- include "containers-cluster-agent-env" . | nindent 4 }}
     {{- end }}
     - name: DD_APM_ENABLED
-      value: "false"
+      value: {{  (default false (or .Values.datadog.apm.enabled .Values.datadog.apmEnabled)) | quote }}
     - name: DD_LOGS_ENABLED
       value: {{  (default false (or .Values.datadog.logs.enabled .Values.datadog.logsEnabled)) | quote}}
     - name: DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 2.37.0
+version: 2.37.1
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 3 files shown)


kayayarai (commented):
Documentation team review not required

charts/datadog/templates/_container-agent.yaml
```diff
@@ -92,7 +92,7 @@
     {{- include "containers-cluster-agent-env" . | nindent 4 }}
     {{- end }}
     - name: DD_APM_ENABLED
-      value: "false"
+      value: {{  (default false (or .Values.datadog.apm.enabled .Values.datadog.apmEnabled)) | quote }}
```
clamoriniere, line 95: the DD_APM_ENABLED has no incident on the `agent` container. it should only impact the `trace-agent` container.

[Here](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_container-trace-agent.yaml#L38) you can see that the value is hardcoded to `true`. 

The current logic for enabling APM with the chart is:
* APM is enable by default with UDS (Unix Domain Socket) communication.
* To have the possibility to send APM trace with a TCP communication, the option `datadog.apm.portEnabled` need to be set to `true`.
   see: https://github.com/DataDog/helm-charts/tree/main/charts/datadog#enabling-apm-and-tracing

clamoriniere (commented):
Thanks for opening this Pull Request.

The proposed change should not have any impact on how APM is enable. I added some explanation about why in the review.

If you have any issue to enable APM, please contact our support https://help.datadoghq.com/ 
and provide the `values.yaml` file use to install the datadog agent with this chart. and the flare of on of the agent. 🙇

charts/datadog/templates/_container-agent.yaml
leszek-niewiadomski: @clamoriniere I have datadog.apm.portEnabled set to true which sets up DD_APM_ENABLED environment variable to true for trace-agent and on agent containers it's set to false. In such configuration the application throws an error `Cause: Errno:ECONNREFUSED Failed to open TCP connection to 10.0.54.163:8126`. If i will manually edit DD_APM_ENABLED env variable on the replica set and configure agent container with DD_APM_ENABLED = true value everything is working as expected. What else can cause this issue?

leszek-niewiadomski:
Issue is not valid. Everything is working as expected.

Issue #549, opened 2022-02-24
Still getting Autopilot GKE request denied errors with Datadog chart


kylegalbraith:
**Describe what happened:**
We are trying to deploy our Datadog chart (`2.28.11`) into our new Autopilot cluster which is on version `1.21.6-gke.1500`. But we are seeing the following errors being reported back out.

`hostPath volume pointerdir in container agent is accessed in write mode; disallowed in Autopilot.`
`hostPath volume runtimesocketdir used in container agent uses path /var/run/containerd which is not allowed in Autopilot`
`hostPath volume cloudinit-instance-id-file used in container agent uses path /var/lib/cloud/data/instance-id which is not allowed in Autopilot`

Plus many others revolving around `hostPath` in particular. 

We **do have** `providers.gke.autopilot` set to `true` which we thought would clear a lot of these errors. This is what I believe was suggested here https://github.com/DataDog/helm-charts/issues/182.

**Describe what you expected:**

Chart to provision in Autopilot without any errors.

**Steps to reproduce the issue:**

Here is the values we are using with the chart.

```
agents:
  containers:
    # Set resource limits for agent container
    agent:
      resources:
        limits:
          cpu: 256m
          memory: 768Mi
        requests:
          cpu: 50m
          memory: 512Mi

    # Set resource limits for process agent container
    processAgent:
      resources:
        limits:
          cpu: 256m
          memory: 512Mi
        requests:
          cpu: 50m
          memory: 128Mi

    # Set resource limits for system probe container
    systemProbe:
      resources:
        limits:
          cpu: 256m
          memory: 512Mi
        requests:
          cpu: 50m
          memory: 128Mi

    # Set resource limits for trace agent container
    traceAgent:
      resources:
        limits:
          cpu: 256m
          memory: 512Mi
        requests:
          cpu: 50m
          memory: 128Mi

  # Run the pods on all nodes, ignoring all taints
  tolerations:
    - operator: Exists

# Datadog cluster agent configuration
clusterAgent:
  # Enable cluster agent
  enabled: true

  # Enable metrics provider, which injects Datadog metrics into Kubernetes
  # to use with the horizontal pod autoscaler
  metricsProvider:
    enabled: true

  # Set resource limits and requests for the Datadog cluster agent
  resources:
    limits:
      cpu: 256m
      memory: 512Mi
    requests:
      cpu: 50m
      memory: 32Mi

  # A random token that secures communication between the cluster agents
[...]

PR #745, opened 2022-09-09
Fix os-release mount in GKE autopilot

CharlyF:

This is blocking a few customers - I believe it might be tied to https://github.com/DataDog/helm-charts/issues/549
We can revisit the logic around mounting the os-release file soon.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-host-release-volumemounts.yaml
+++ b/charts/datadog/templates/_container-host-release-volumemounts.yaml
@@ -4,7 +4,7 @@
   mountPath: /host{{ .Values.datadog.systemProbe.osReleasePath | default .Values.datadog.osReleasePath }}
   mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
   readOnly: true
-{{- else }}
+{{- else if not .Values.providers.gke.autopilot}}
 - name: os-release-file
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/_container-host-release-volumemounts.yaml
```diff
@@ -4,7 +4,7 @@
   mountPath: /host{{ .Values.datadog.systemProbe.osReleasePath | default .Values.datadog.osReleasePath }}
   mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
   readOnly: true
-{{- else }}
+{{- else if not .Values.providers.gke.autopilot}}
```
clamoriniere, line 7: I was thinking that `read-only` mount was "OK" on auto-pilot. 
Does the rule has changed?

clamoriniere (commented):
The changes look good, but I think it was possible to add as many host-path if they were in `readonly` mode which is the case for this volume.

charts/datadog/templates/_container-host-release-volumemounts.yaml
clamoriniere: the only volume that needed read/write was the volume that store the release.json used by the logs-agent.

clamoriniere (approved):

PR #768, opened 2022-09-30
Use seccompProfile under securityContext in k8s 1.19+ for system-probe

brycekahle:

Uses the preferred `securityContext.seccompProfile` instead of the annotations for kubernetes v1.19+

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_helpers.tpl
+++ b/charts/datadog/templates/_helpers.tpl
@@ -669,6 +669,19 @@ securityContext:
 {{- else }}
 securityContext:
 {{ toYaml .securityContext | indent 2 }}
+{{- if and .seccomp .kubeversion (semverCompare ">=1.19.0" .kubeversion) }}
+  seccompProfile:
+    {{- if hasPrefix "localhost/" .seccomp }}
+    type: Localhost
+    {{- else if eq "runtime/default" .seccomp }}
+    type: RuntimeDefault
+    {{- else }}
+    type: Unconfined
+    {{- end -}}
+    {{- if hasPrefix "localhost/" .seccomp }}
+    localhostProfile: {{ .seccomp }}
+    {{- end }}
+{{- end -}}
 {{- end -}}
 {{- end -}}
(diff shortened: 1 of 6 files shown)


jtappa (approved):

charts/datadog/templates/_helpers.tpl
```diff
@@ -669,6 +669,13 @@ securityContext:
 {{- else }}
 securityContext:
 {{ toYaml .securityContext | indent 2 }}
+{{- if and .seccomp .kubeversion (semverCompare ">=1.19.0" .kubeversion) }}
+  seccompProfile:
+    type: {{ if hasPrefix "localhost/" .seccomp }}Localhost{{ else if eq "runtime/default" .seccomp }}RuntimeDefault{{ else }}Unconfined{{ end }}
+    {{- if hasPrefix "localhost/" .seccomp }}
+    localhostProfile: {{ .seccomp }}
+    {{- end }}
```
clamoriniere, line 677: could you try to refactor the code to have the "if,else if,else" on different lignes

you can use `{{- ... -}}` to strip the lines.

charts/datadog/templates/daemonset.yaml
```diff
@@ -54,8 +54,10 @@ spec:
         {{- if .Values.agents.podSecurity.apparmor.enabled }}
         container.apparmor.security.beta.kubernetes.io/system-probe: {{ .Values.datadog.systemProbe.apparmor }}
         {{- end }}
+        {{- if semverCompare "<1.19.0" .Capabilities.KubeVersion.Version }}
```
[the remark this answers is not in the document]
brycekahle: It causes a deprecation warning now.

charts/datadog/templates/_helpers.tpl
brycekahle: Updated.

clamoriniere (approved):

PR #777, opened 2022-10-06
feat: add installInfoExistingConfigMap

hevcy:

Enable the ability to use a pre-existing Install Info ConfigMap.

This is something, admittedly very niche, that I ran into when trying to have separate Linux and Windows deployments on a cluster, it will not install as it creates duplicate ConfigMaps.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.1.10
+
+* Add `datadog.installInfoExistingConfigMap` option.
+
 ## 3.1.9
 
 * Add `faccessat` to system-probe seccomp profile.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.1.9
+version: 3.1.10
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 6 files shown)


clamoriniere:
Hi @hevcy 

thanks for opening this contribution. 

Do you say that when you have two chart installations in the same Namespace, the 2 ConfigMap are conflicting? the function `agents-install-info-configmap-name` should create configmap containing the helm chart installation name: https://github.com/DataDog/helm-charts/blob/58b8218341a45e008b2b3e0274415f218ea5c93d/charts/datadog/templates/_helpers.tpl#L479-L485

Could you explain a bit more which information you would like to store in each configmap? do you want to use the same configmap?

cswatt (approved):

hevcy:
Hi! I'm using my own helm chart that's installing datadog as a dependency. There are two instances of the dependency, one for Linux and one for Windows, but I've set a different alias for each, such as:

`- name: datadog
  alias: "datadog"
  version: "~3.1.9"
  repository: "https://helm.datadoghq.com"
- name: datadog
  alias: "datadog-windows"
  version: "~3.1.9"
  repository: "https://helm.datadoghq.com"`

This means that when I try installing the chart, it immediately fails because the configmap exists twice. In a similar vein, for the datadog-windows chart, I have set `datadog.apiKeyExistingSecret` so that it can use the one created from the first install.

Hopefully that makes sense!

clamoriniere:
Hi again @hevcy 

thanks it totally makes sense now. 

I'm wondering it we could take another approach. What I have in mind is using a parameter "fullnameOverride" that already exist to override the `values.fullname` field.

the values.yaml file of parent chart should look like:

```yaml
datadog:
  fullnameOverride: datadog
  // ...

datadog-windows:
  fullnameOverride: datadog-windows
  targetSystem: windows
  existingClusterAgent:
    join: true
    serviceName: "<EXISTING_DCA_SERVICE_NAME>" # from the first Datadog Helm chart
    tokenSecretName: "<EXISTING_DCA_SECRET_NAME>" # from the first Datadog Helm chart
  // ...
```

Does it makes sense?

hevcy:
Hi @clamoriniere, I think I had gone too far down this rabbit hole and completely missed that setting, It has indeed resolved my issue, thanks! I'll close this off.

clamoriniere:
Hi @hevcy  

Awesome news! I'm happy to see that we found a solution.

Thanks for your clear issue description, it was an efficient discussion 👍

PR #778, opened 2022-10-07
Remove DD_HOSTNAME from agent-clusterchecks

jerryc8:
Removed `DD_HOSTNAME` so that it won't supersede hostname from other sources.  More specifically, the presence of this environment variable supersedes other sources where the hostname can come from, such as imds (in case of aws).  This is a problem because the incorrect value provided by `spec.nodeName` can cause the wrong hostname to be sent to the backend, causing confusion.

These removed lines came from [this first commit](https://github.com/DataDog/helm-charts/commit/c88ada62#diff-a8ae734b61ca9075fd28fabe0f8a2af3d7ae18f99e3b0f765296e89d7aadab69R96) which added the datadog chart.  The commit is over two years old and these 4 lines are probably now obsolete given how datadog has evolved over the years.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Datadog changelog
 
+## 3.1.10
+* Removed `DD_HOSTNAME` so that it won't supersede hostname from other sources.
+
 ## 3.1.9
 
 * Add `faccessat` to system-probe seccomp profile.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.1.9
+version: 3.1.10
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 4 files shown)


cep21:
To clarify the bug, this can cause datadog to *think* it has more hosts running in the cluster, than it actually has.  This causes customers to spend more money (because datadog gets confused and thinks you have extra hosts).  In our case, the bill was close to 10-20% higher.

kayayarai (approved):

jerryc8:
Hi @kayayarai thank you for the approval, can you please also approve the `pr-validated` workflow so that this PR becomes mergeable?  Thanks!

kayayarai:
Please note that my approval was strictly the documentation. Someone from container-integrations should also have a look

vboulineau:
Hey @jerryc8,

We don't expect `DD_HOSTNAME` to have influence on your billing. Normally a Cluster Checks does not send host metadata and should not trigger billing either.

I'd encourage you to submit a support ticket so we can investigate the underlying issue.

cep21:
@vboulineau we submitted a support ticket and had an in person zoom call with a datadog engineer.  The result of the call, was that from the engineers perspective this was likely the cause of our double billing.  It is request 894011

vboulineau:
Hello @cep21, I followed the ticket and we can confirm that this is not the root cause of your billing issue.
I'm thus closing this PR as it could cause issues in some scenarios.

PR #794, opened 2022-10-26
charts/datadog: Use downward API to provide K8S CPU Limit to trace-agent

ajgajg1134:
This PR exposes any configured cpu limit on the trace-agent to the container so it can use it to better inform rate limiting. Currently in K8s environments the trace-agent only relies on the container runtime / k8s for cpu throttling. This means that the trace-agent can become overwhelmed with traffic but not report this to the backend leading to customers being unaware their trace-agents need more resources (or less traffic). Exposing the k8s limit to the trace-agent will allow it to set an internal "CPU Watchdog" based on this value which is then used to rate limit (read: drop) traffic when it starts to be overwhelmed, thereby indicating to the backend (via tags) that the trace-agent has insufficient resources.

The previous default value of `0` is removed so that the auto-lookup can take place. Setting a value of `0` explicitly disables the cpu watchdog entirely.

[Here's the PR ](https://github.com/DataDog/datadog-agent/pull/14014) where the trace-agent will make use of this new variable

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/datadog-yaml-configmap.yaml
+++ b/charts/datadog/templates/datadog-yaml-configmap.yaml
@@ -30,7 +30,6 @@ data:
       enabled: true
       apm_non_local_traffic: true
       max_memory: 0
-      max_cpu_percent: 0
 
     {{- $version := (.Values.agents.image.tag | toString | trimSuffix "-jmx") }}
     {{- $length := len (split "." $version ) -}} 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.2.0
+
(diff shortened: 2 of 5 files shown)


charts/datadog/templates/datadog-yaml-configmap.yaml
```diff
@@ -30,7 +30,6 @@ data:
       enabled: true
       apm_non_local_traffic: true
       max_memory: 0
-      max_cpu_percent: 0
```
katiehockman, line 33: Why remove the default?

katiehockman (approved):

charts/datadog/templates/datadog-yaml-configmap.yaml
katiehockman: Ah I see that you explain this a bit in the other PR. Just worth a call out in this PR description too.

buraizu (approved):

PR #798, opened 2022-11-04
Adding option to store v2alpha1 version of the DatadogAgent

CharlyF:
#### What this PR does / why we need it:

PR in the CRDs before PR in the operator to support using DDA in v2alpha1 directly.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has been updated
- [x] Variables are documented in the `README.md`

PR #799, opened 2022-11-04
Update datadog operator chart v2alpha1

CharlyF:

- Following up on https://github.com/DataDog/helm-charts/pull/798
- Add reference to updated CRD that allows storing/serving v2alpha1
- Add ability to deactivate Conversion Webhook and add safeguard not to use for versions of the Operator that don't support it.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/values.yaml
+++ b/charts/datadog-operator/values.yaml
@@ -110,8 +110,10 @@ datadog-crds:
     datadogAgents:
       version: "v1alpha1"
 
-migration:
-  webhook:
+# webhook -- configure webhook servers
+webhook:
+  ## Use the conversion webhook server when multiple versions of objects managed by the Datadog Operator are served.
+  conversion:
     enabled: false
 
 # podAnnotations -- Allows setting additional annotations for Datadog Operator PODs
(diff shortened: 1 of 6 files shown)


kayayarai (approved):
👍 for docs

charts/datadog-operator/values.yaml
```diff
@@ -106,6 +106,13 @@ datadog-crds:
     datadogMetrics: true
     # datadog-crds.crds.datadogMonitors -- Set to true to deploy the DatadogMonitors CRD
     datadogMonitors: true
+  migration:
+    datadogAgents:
+      version: "v1alpha1"
+
+migration:
+  webhook:
+    enabled: false
```
clamoriniere, line 115: maybe you can aggregate this migration in one block

```yaml
migration:
  datadogAgents:
    version: "v1alpha1"
  webhook:
    enabled: false 
```

also it can be done maybe in another way

```yaml
supportedDatadogAgentVersions:
  v1alpha1: true
  v2alpha1: true
```

CharlyF: I agree, as is, I would be better off aggregating the sections. However helm does not let users use `-` in variable names, so I can't have options under the `datadog-crds` option be used in the logic of the datadog-operator chart.
However, I realized that we will be adding more webhook backed features in the future so it makes sense to isolate that section in something more appropriate such as:

```yaml
webhook:
  conversion:
    enabled: false
```

khewonc (approved):

PR #808, opened 2022-11-10
Add configuration value for mounting custom BTFs

ISauve:

Add `datadog.systemProbe.btfPath` for mounting user-provided BTF files for Compile Once Run Everywhere (CO-RE) (see [#13962](https://github.com/DataDog/datadog-agent/pull/13962) and [#14096](https://github.com/DataDog/datadog-agent/pull/14096) for more context).

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -130,7 +130,7 @@
 {{- end }}
 {{- end }}
 {{- end }}
-{{- if and .Values.datadog.systemProbe.enableCORE .Values.datadog.systemProbe.btfPath }}
+{{- if .Values.datadog.systemProbe.btfPath }}
     - name: btf-path
       mountPath: {{ .Values.datadog.systemProbe.btfPath }}
       readOnly: true
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -130,25 +130,10 @@
 {{- end }}
 {{- end }}
 {{- end }}
-{{- if .Values.datadog.systemProbe.enableCORE }}
-{{- if .Values.datadog.systemProbe.btfPath }}
+{{- if and .Values.datadog.systemProbe.enableCORE .Values.datadog.systemProbe.btfPath }}
     - name: btf-path
       mountPath: {{ .Values.datadog.systemProbe.btfPath }}
       readOnly: true
-{{- else }}
-  - name: lib-modules
-    mountPath: /lib/modules
-    mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
-    readOnly: true
-  - name: usr-lib-modules
-    mountPath: /usr/lib/modules
-    mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
-    readOnly: true
-  - name: usr-lib-debug
-    mountPath: /usr/lib/debug
-    mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
-    readOnly: true
-{{- end }}
 {{- end }}
 {{- if .Values.agents.volumeMounts }}
 {{ toYaml .Values.agents.volumeMounts | indent 4 }}
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -78,7 +78,7 @@
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
 {{- end }}
-{{- if and (or (eq (include "runtime-compilation-enabled" .) "true") .Values.datadog.systemProbe.enableCORE) .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
+{{- if and (eq (include "runtime-compilation-enabled" .) "true") .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
     - name: modules
       mountPath: /lib/modules
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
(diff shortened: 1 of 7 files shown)


charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -81,7 +81,7 @@
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
 {{- end }}
-{{- if and (or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill) .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
+{{- if and (or (eq (include "runtime-compilation-enabled" .) "true") .Values.datadog.systemProbe.enableCORE) .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
```
brycekahle, line 84: Why does CO-RE need kernel headers? It shouldn't need this mount.

ISauve: While we primarily expect default kernel BTFs to be in `/sys/kernel/btf/vmlinux` (which already gets mounted), they can sometimes  be found in `/lib/modules`. It's a bit confusing to lump this in with the kernel header configurations though, so I could separate this into a separate statement for clarity if necessary.

ISauve: oh actually - I just double checked the default BTF locations and realized the default BTFs can be found in `/lib/modules` and `/usr/lib`, not `/usr/src`. So it doesn't make sense to lump these together - let me fix that.

ISauve: ([source](https://github.com/cilium/ebpf/blob/9da42c2fd67e8761067fcd9c27eb24b23acdb51c/btf/btf.go#L274-L319))

brycekahle: There is a much larger list of searched locations https://github.com/cilium/ebpf/blob/master/btf/btf.go#L301-L307
I don't think we can get away with mounting `/boot` though.

I'm tempted to not have any mounts for CO-RE because:
1. They can specify a custom path if necessary
2. It should be rare that the kernel BTF isn't in the default location and we don't have a bundled BTF file

Simplifying the security concerns and mounting requirements for non-helm setups is another benefit.

ISauve: Don't worry, I wasn't planning on mounting `/boot` :) Just `/lib/modules` and `/usr/lib/` (though I could make it `/usr/lib/modules` + `/usr/lib/debug` instead to minimize the mounts).

Is it a security concern if we're already mounting similar paths for kernel header downloading? I agree that this is probably a rare case, which means I don't think we should bother worrying about it for non-helm setups, but if we can add support for that case here, I don't see why we wouldn't.

brycekahle: So the only case where those files would be used, is if they had installed the package with the debug kernel **and it had BTF embedded** right? That seems sufficiently rare to not worry about, especially if there is a workaround.

jtappa (approved):

charts/datadog/values.yaml
```diff
@@ -520,6 +520,12 @@ datadog:
     # datadog.systemProbe.runtimeCompilationAssetDir -- Specify a directory for runtime compilation assets to live in
     runtimeCompilationAssetDir: /var/tmp/datadog-agent/system-probe
 
+    # datadog.systemProbe.enableCORE -- Enables Compile Once - Run Everywhere (CO-RE) for eBPF probes
+    enableCORE: true
```
clamoriniere, line 524: I see that it is enable by default. 
what happens if the host kernel version doesn't support CO-RE?
does the customer needs to change manually this Option to `false`?

do we have some logic in system-probe to fallback on the previous method?

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -125,6 +125,11 @@
 {{- end }}
 {{- end }}
 {{- end }}
+{{- if and .Values.datadog.systemProbe.enableCORE .Values.datadog.systemProbe.btfPath }}
```
clamoriniere, line 128: it ie enable by default, but. the btfPath is empty by default. so it means that it is disable by default.

can we just say: if btfPath is not empty we mount it?

charts/datadog/values.yaml
ISauve: If CO-RE fails, we automatically fall back to runtime compilation (or whatever the current default way is of loading a particular eBPF program). You can see an example here: https://github.com/DataDog/datadog-agent/pull/14096

ISauve: FWIW - we enable CO-RE by default in the agent code as well.

clamoriniere (commented):
Thanks for all the answers. 

please, rebase on `main` and bump again the chart version.
Also, small nits: could you reference the release-note or the PR that introduce the feature in the datadog-agent repo. It will allow users to better understand what this new option is about 🙇

charts/datadog/templates/_daemonset-volumes-linux.yaml
ISauve: Done :)

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -78,7 +78,7 @@
   name: debugfs
 - name: sysprobe-socket-dir
   emptyDir: {}
-{{- if and (or .Values.datadog.systemProbe.enableTCPQueueLength .Values.datadog.systemProbe.enableOOMKill) .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
+{{- if and (eq (include "runtime-compilation-enabled" .) "true") .Values.datadog.systemProbe.enableDefaultKernelHeadersPaths }}
```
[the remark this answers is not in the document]
ISauve: `.Values.datadog.systemProbe.enableRuntimeCompiler` should have been included here in the first place - this is just a small bug fix - which admittedly I should have put in a separate PR, I had just included it here because in my original set of changes I was already changing this line 😬  If you think it makes sense to address separately, I can cut a new PR for that

CharlyF (approved):

CharlyF:
the CI failure for K8s 1.18.20 can be ignored, we are fixing this on the side.

brycekahle (approved):

PR #809, opened 2022-11-10
Remove defaults for enableKernelHeaderDownload and enableRuntimeCompiler

ISauve:

Removes the default values for `datadog.systemProbe.enableKernelHeaderDownload` and `datadog.systemProbe.enableRuntimeCompiler`, and only set those values in the system-probe's configuration file if those fields have been manually set.

The system-probe enables runtime compilation & kernel header downloading by default when USM has been enabled, but only when those features have not been manually enabled/disabled. By setting default values here, the system-probe thinks those values have always been manually set, and so it fails to enable the features for USM customers.

This will affect the OOMKill and TCP Queue Length checks, as kernel header downloading is no longer going to be enabled by default; if that behavior is desired, those checks should enable kernel header downloading by default in their config settings.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.2.0](https://img.shields.io/badge/Version-3.2.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.2.1](https://img.shields.io/badge/Version-3.2.1-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 [Datadog](https://www.datadoghq.com/) is a hosted infrastructure monitoring platform. This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally depends on the [kube-state-metrics chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics). For more information about monitoring Kubernetes with Datadog, please refer to the [Datadog documentation website](https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/).
 
@@ -693,9 +693,7 @@ helm install <RELEASE_NAME> \
 | datadog.systemProbe.enableConntrack | bool | `true` | Enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data |
 | datadog.systemProbe.enableDefaultKernelHeadersPaths | bool | `true` | Enable mount of default paths where kernel headers are stored |
 | datadog.systemProbe.enableDefaultOsReleasePaths | bool | `true` | enable default os-release files mount |
-| datadog.systemProbe.enableKernelHeaderDownload | bool | `true` | Enable the downloading of kernel headers for runtime compilation of eBPF probes |
 | datadog.systemProbe.enableOOMKill | bool | `false` | Enable the OOM kill eBPF-based check |
-| datadog.systemProbe.enableRuntimeCompiler | bool | `false` | Enable the runtime compiler for eBPF probes |
 | datadog.systemProbe.enableTCPQueueLength | bool | `false` | Enable the TCP queue length eBPF-based check |
 | datadog.systemProbe.maxTrackedConnections | int | `131072` | the maximum number of tracked connections |
 | datadog.systemProbe.mountPackageManagementDirs | list | `[]` | Enables mounting of specific package management directories when runtime compilation is enabled |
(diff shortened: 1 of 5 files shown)


charts/datadog/values.yaml
```diff
@@ -503,10 +503,10 @@ datadog:
     enableOOMKill: false
 
     # datadog.systemProbe.enableRuntimeCompiler -- Enable the runtime compiler for eBPF probes
-    enableRuntimeCompiler: false
+    # enableRuntimeCompiler: false
```
brycekahle, line 506: I think you can do this:
```suggestion
    enableRuntimeCompiler:
```

brycekahle: But that might cause a change in your other template code. The benefit of having it defined, but unset, is that it shows up in documentation.

brycekahle: Maybe I'm wrong about the documentation though, I'm not 100% sure how that is processed.

charts/datadog/README.md
```diff
@@ -693,9 +693,7 @@ helm install <RELEASE_NAME> \
 | datadog.systemProbe.enableConntrack | bool | `true` | Enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data |
 | datadog.systemProbe.enableDefaultKernelHeadersPaths | bool | `true` | Enable mount of default paths where kernel headers are stored |
 | datadog.systemProbe.enableDefaultOsReleasePaths | bool | `true` | enable default os-release files mount |
-| datadog.systemProbe.enableKernelHeaderDownload | bool | `true` | Enable the downloading of kernel headers for runtime compilation of eBPF probes |
 | datadog.systemProbe.enableOOMKill | bool | `false` | Enable the OOM kill eBPF-based check |
-| datadog.systemProbe.enableRuntimeCompiler | bool | `false` | Enable the runtime compiler for eBPF probes |
```
ISauve, line 698: Ideally, I would like to find a way to leave these config options in the README in order to provide visibility that they exist, but I'm not sure how to do that when they don't exist by default.

Note that having no value set (i.e. having `enableRuntimeCompiler: <empty>`) in `values.yaml` doesn't work, as there is no way AFAIK for the configmap to differentiate between a nil value and a manually set value of `false`. We need to be able to tell the difference, because if it's nil we don't want to write anything to the config map, but if it's `false` we do.

charts/datadog/values.yaml
ISauve: Unfortunately this doesn't work for our use case - see my note in the README.

jtappa (approved):

charts/datadog/README.md
ISauve: @brycekahle I found  a helm template function which allows us to check directly whether or not the value has been set, so we can leave these options in our `values` file and still know whether or not to write them to the configmap, which solves this problem

charts/datadog/README.md
```diff
@@ -693,9 +693,9 @@ helm install <RELEASE_NAME> \
 | datadog.systemProbe.enableConntrack | bool | `true` | Enable the system-probe agent to connect to the netlink/conntrack subsystem to add NAT information to connection data |
 | datadog.systemProbe.enableDefaultKernelHeadersPaths | bool | `true` | Enable mount of default paths where kernel headers are stored |
 | datadog.systemProbe.enableDefaultOsReleasePaths | bool | `true` | enable default os-release files mount |
-| datadog.systemProbe.enableKernelHeaderDownload | bool | `true` | Enable the downloading of kernel headers for runtime compilation of eBPF probes |
+| datadog.systemProbe.enableKernelHeaderDownload | string | `nil` | Enable the downloading of kernel headers for runtime compilation of eBPF probes |
```
brycekahle, line 696: Not ideal that this lists the type as `string`, but not sure what we can do about it.

brycekahle: @clamoriniere any advice on how to approach this?

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -32,8 +32,12 @@ data:
       collect_dns_stats: {{ $.Values.datadog.systemProbe.collectDNSStats }}
       max_tracked_connections: {{ $.Values.datadog.systemProbe.maxTrackedConnections }}
       conntrack_max_state_size: {{ $.Values.datadog.systemProbe.conntrackMaxStateSize }}
+      {{- if eq  (include "enableRuntimeCompiler-is-set" .) "true" }}
       enable_runtime_compiler: {{ $.Values.datadog.systemProbe.enableRuntimeCompiler }}
+      {{- end }}
+      {{- if eq  (include "enableKernelHeaderDownload-is-set" .) "true" }}
       enable_kernel_header_download: {{ $.Values.datadog.systemProbe.enableKernelHeaderDownload }}
```
clamoriniere, line 39: instead of defaulting `enableKernelHeaderDownload` to `nil` can we  keep `true` as default value but do something like:
```suggestion
      {{- if not $.Values.datadog.systemProbe.enableKernelHeaderDownload }}
      enable_kernel_header_download: false
      {{- end }}
```

ISauve:
Closing in favour of #813

PR #811, opened 2022-11-11
[observability-pipelines-worker] Initial commit

spencergilbert:
Signed-off-by: Spencer Gilbert <spencer.gilbert@datadoghq.com>

Creating a new helm chart to deploy the observability pipelines worker binary in Kubernetes environments.

_n/a_

#### Checklist

- [x] Chart Version bumped


charts/observability-pipelines-worker/templates/daemonset.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if (eq .Values.role "Agent") -}}
```
spencergilbert, line 1: @lukesteensen @binarylogic there was discussion about only supporting the service/aggregator models. If we can confirm that there's quite a bit of templating that can get cut away.

charts/observability-pipelines-worker/templates/deployment.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if (eq .Values.role "Stateless-Aggregator") -}}
```
spencergilbert, line 1: @lukesteensen @binarylogic similar to the agent, if we wanted to force StatefulSets for durability reasons (though they can be created without persistent storage) we could drop this file and related templates.

charts/observability-pipelines-worker/templates/haproxy/deployment.yaml
```diff
@@ -0,0 +1,137 @@
+{{- if .Values.haproxy.enabled }}
```
spencergilbert, line 1: @lukesteensen @binarylogic this may be contentious but I'd recommend removing HAProxy from the chart. @silversupreme has mentioned to me that it has caused some problems in the past and that for testing/initial onboarding a single instance should be suitable.

charts/observability-pipelines-worker/templates/podmonitor.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if and .Values.podMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
```
spencergilbert, line 1: Not sure if we want to keep this or assume this will be monitored by a Datadog agent, or self-monitored.

charts/observability-pipelines-worker/templates/rbac.yaml
```diff
@@ -0,0 +1,45 @@
+{{- if and .Values.rbac.create (eq .Values.role "Agent") -}}
```
spencergilbert, line 1: If we're not shipping an agent option we don't need RBAC.

spencergilbert (commented):
Added initial thoughts and questions, and started replacing the Vector specific bits and naming.

spencergilbert:
@CharlyF, before I go too far into this - I was thinking that along with user facing changes to better match the other DD charts here, that I could rewrite some of the internals to better match how the other charts are written. The theory here being maintenance burden would be _lower_ if the content/templates are familiar/consistent.

What do you think?

charts/observability-pipelines-worker/templates/daemonset.yaml
lukesteensen: Yes, it should be safe to remove anything related to the agent-style model. OP for the foreseeable future will focus on the aggregator use case.

charts/observability-pipelines-worker/templates/deployment.yaml
lukesteensen: I'm not too sure about this one. AFAIK OP will support the normal range of disk-based Vector features, so I'm not sure if we want to differ here or not. Maybe @binarylogic will have thoughts.

charts/observability-pipelines-worker/templates/haproxy/deployment.yaml
lukesteensen: I don't have much context here, but anything that keeps things simpler seems like a good idea to me.

charts/observability-pipelines-worker/templates/podmonitor.yaml
lukesteensen: Again, not a ton of context on this one, but my assumption is that the only official/supported/recommended monitoring would be via the Datadog agent.

.github/CODEOWNERS
```diff
@@ -10,3 +10,4 @@ charts/datadog/templates/container-system-probe.yaml @DataDog/ebpf-platform @Dat
 charts/datadog/templates/system-probe-configmap.yaml @DataDog/ebpf-platform @DataDog/container-integrations
 charts/datadog/templates/system-probe-init.yaml      @DataDog/ebpf-platform @DataDog/container-integrations
 charts/synthetics-private-location/                  @Datadog/synthetics
+charts/observability-pipelines-worker                @spencergilbert @lukesteensen
```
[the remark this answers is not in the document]
clamoriniere: hi @spencergilbert
we can also define a team for you 2. it will be simpler after to maintains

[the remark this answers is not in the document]
spencergilbert: 👍 whatever the preference is from your side, I imagine I'd do most of the changes here - but it would be good for the observability pipelines time to have ownership.

Perhaps the better way forward is adding @DataDog/observability-pipelines as the owner, and then adding me/Luke to that team?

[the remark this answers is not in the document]
spencergilbert: Updated to the obs-pipeline team.

spencergilbert:
I _think_ I've updated all the places I wanted to for this first draft. The CI files could be improved but that's not user facing (unless a regression slips through).

buraizu:
Thanks, created DOCS-4450 for docs team review

spencergilbert:
cc @silversupreme @CharlyF

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+
+# podManagementPolicy -- Specify the
+# [podManagementPolicy](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies).
+podManagementPolicy: OrderedReady
+
+## TODO: Support Watermark Pod Autoscaler?
+autoscaling:
+  # autoscaling.enabled -- If **true**, create a
+  # [HorizontalPodAutoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/).
+  enabled: false
```
silversupreme, line 69: **🌶️ Spicy Opinion Hour 🌶️**

Any particular reason we wouldn't want to enable this by default? Pretty much anyone using this is gonna want to autoscale, I'm finding.

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+
+# securityContext -- Specify securityContext for containers.
+securityContext: {}
+
+# command -- Override default image command.
+command: []
+
+# args -- Override default image arguments.
+args:
+  - --config
```
silversupreme, line 125: Always wondered this- would it be reasonable to use `--config-dir` instead so it's easier for folks to mount additional config files?

silversupreme (approved):
6j679t

Looks good based on what I've seen from the original chart. A good foundation for us to work from!

charts/observability-pipelines-worker/values.yaml
spencergilbert: 😆 take it up with @lukesteensen! Currently OPW just takes a single config file, https://github.com/DataDog/observability-pipelines-worker/blob/master/src/main.rs#L11

charts/observability-pipelines-worker/values.yaml
spencergilbert: The only thing immediately coming to mind would be deploying to clusters where it's not supported. If it's an option that the majority of users are going to be enabling, I figure we can default it to `true` and gracefully handle the failure paths.

vboulineau:
Chart looks healthy and in line with the others, good work, only one comment.

.run/Docs.run.xml
```diff
@@ -0,0 +1,8 @@
+<component name="ProjectRunConfigurationManager">
```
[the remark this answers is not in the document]
spencergilbert: 😅 nope, I thought I had these globally gitignored. I'll fix

spencergilbert:
The most recent CI failure looks like a runner failure, should someone with permissions retry that or should I push an empty commit to trigger CI again?

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+fullnameOverride: ""
+
+# commonLabels -- Labels to apply to all resources.
+commonLabels: {}
+#  team_name: dev
+
+datadog:
+  # datadog.apiKey -- Specify your Datadog API key.
+  apiKey: <DD_API_KEY>
+  # datadog.apiKeyExistingSecret -- Specify a preexisting Secret which has your API key instead of creating a new one.
```
rtrieu, line 21:
```suggestion
  # datadog.apiKeyExistingSecret -- Specify a preexisting secret that has your API key instead of creating a new one.
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+  # The value must be stored under the `api-key`.
+  apiKeyExistingSecret: ""
+  # datadog.configkey -- Specify your Datadog Configuration key.
+  configKey: <DD_CONFIGURATION_KEY>
+  # datadog.configKeyExistingSecret -- Specify a preexisting Secret which has your Configuration key instead of
```
rtrieu, line 26:
```suggestion
  # datadog.configKeyExistingSecret -- Specify a preexisting secret that has your configuration key instead of
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+  ## Set to 'us3.datadoghq.com' to send data to the US3 site.
+  ## Set to 'us5.datadoghq.com' to send data to the US5 site.
+  ## Set to 'ddog-gov.com' to send data to the US1-FED site.
+  # datadog.site -- The [site](https://docs.datadoghq.com/getting_started/site/) of the Datadog intake to send data to.
+  site: datadoghq.com
+
+image:
+  # image.name -- Specify image name to use (relative to `image.repository`).
+  name: observability-pipelines-worker
+  # image.tag -- Specify image tag to use.
```
rtrieu, line 40:
```suggestion
  # image.tag -- Specify the image tag to use.
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+  tag: latest
+  # image.digest -- (string) Specify image digest to use, takes precedence over `image.tag`.
```
rtrieu, line 42:
```suggestion
  # image.digest -- (string) Specify the image digest to use; takes precedence over `image.tag`.
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+  # autoscaling.behavior -- Configure separate scale-up and scale-down behaviors.
+  behavior: {}
+    # scaleDown:
+    #   stabilizationWindowSeconds: 300
+
+podDisruptionBudget:
+  # podDisruptionBudget.enabled -- If **true**, create a
+  # [PodDisruptionBudget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/).
+  enabled: false
+  # podDisruptionBudget.minAvailable -- Specify the number of Pods that must still be available after an eviction.
```
rtrieu, line 87:
```suggestion
  # podDisruptionBudget.minAvailable -- Specify the number of pods that must still be available after an eviction.
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+  create: true
+  # serviceAccount.name -- Specify a preexisting ServiceAccount to use if `serviceAccount.create` is **false**.
+  name: default
+  # serviceAccount.annotations -- Annotations to add to the ServiceAccount, if `serviceAccount.create` is **true**.
+  annotations: {}
+
+# podAnnotations -- Set annotations on Pods.
+podAnnotations: {}
+
+# podLabels -- Set labels on Pods.
```
rtrieu, line 103:
```suggestion
# podLabels -- Set labels on pods.
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -0,0 +1,340 @@
...
+podLabels: {}
+
+# podPriorityClassName -- Set the
+# [priorityClassName](https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#priorityclass).
+podPriorityClassName: ""
+
+# podHostNetwork -- Enable the hostNetwork option on Pods.
```
rtrieu, line 110:
```suggestion
# podHostNetwork -- Enable the hostNetwork option on pods.
```

rtrieu (changes requested):

charts/observability-pipelines-worker/README.md
```diff
@@ -0,0 +1,160 @@
...
+To install the chart with the release name `<RELEASE_NAME>` run:
+
+```bash
+helm install --name <RELEASE_NAME> \
+  --set datadog.apiKey=<DD_API_KEY> \
+  --set datadog.configKey=<DD_CONFIGURATION_KEY> \
+  datadog/observability-pipelines-worker
+```
+
+By default, this Chart creates Secrets for your Observability Pipelines API and Configuration keys. However, you can use
```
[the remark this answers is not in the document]
rtrieu: Hiya @spencergilbert, made some minor suggestions for consistency. Otherwise, looks good!

spencergilbert:
👍 Thanks @rtrieu, I'll have this updated by tomorrow, appreciate the review!

charts/observability-pipelines-worker/values.yaml
spencergilbert: Following up on this @silversupreme - this was a "v0.1.0" intentional limitation we'll come back to down the line.

spencergilbert:
All your feedback should be addressed with the latest commit @rtrieu - I think some of the same changes need/should be applied to the `datadog` chart as well, I believe I pulled more than a few of the lines directly from there 👍

rtrieu (approved):

charts/observability-pipelines-worker/values.yaml
lukesteensen: Sorry, thought I'd already responded to this @silversupreme! But yeah, we're intentionally scoping down the wide range of config loading options from Vector, at least for now. I think `--config-dir` specifically is one of the more complex ones, but we're happy to add support if we really think it'll be valuable to users.

spencergilbert:
OK, with @lukesteensen's publishing of the `0.1.0` image this PR should be good to review without further changes (aside from possible PR feedback).

cc @vboulineau @clamoriniere @CharlyF

CharlyF (approved):

PR #813, opened 2022-11-16
Remove runtime compilation helm config options

ISauve:

Removes `enableKernelHeaderDownload` and `enableRuntimeCompiler` from the helm chart values. These settings should instead be configured via the `DD_ENABLE_KERNEL_HEADER_DOWNLOAD` and `DD_ENABLE_RUNTIME_COMPILER` environment variables.

Runtime compilation and kernel header downloading are being slowly released to customers. At the current stage of this release, we want these features to be turned off by default for all customers except USM customers. This is achieved by having the system-probe turn on runtime compilation & kernel header downloading when customers have USM enabled, but **only if they haven't explicitly configured these features themselves** (that is, only if these configuration values are not set - see [here](https://github.com/DataDog/datadog-agent/blob/6a3d555a334cfd2dacd7e17c7f9614a2fd22ed28/pkg/network/config/config.go#L324-L332)). We do this so that USM customers have a way to turn off runtime compilation & kernel header downloading if they need to.

When we set default values for runtime compilation & kernel header downloading in the helm charts, the system-probe sees that these configuration values have been set, and so it never enables runtime compilation and kernel header downloading for USM customers. Thus, having default values set in the helm chart is interfering with our rollout of these features.

For the purposes of knowing when to mount runtime compilation-related files, we can simply assume runtime compilation is enabled when either the OOM Kill/TCP Queue length checks or service monitoring (USM) is enabled.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Datadog changelog
 
-## 3.4.1
+## 3.5.0
 
 * Remove runtime compilation-related config values `enableKernelHeaderDownload` and `enableRuntimeCompiler` in the system-probe.
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.4.1
+
+* Remove runtime compilation-related config values `enableKernelHeaderDownload` and `enableRuntimeCompiler` in the system-probe.
+
 ## 3.4.0
 
 * Add `datadog.systemProbe.btfPath` for mounting user-provided BTF files (see datadog-agent PRs #13962 and #14096 for more context).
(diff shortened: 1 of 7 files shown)


buraizu (approved):

CharlyF (changes requested):
We don't drop fields in minor bumps unfortunately, could you add a deprecation notice instead https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/NOTES.txt#L132?
We will drop them in the next major version (4.0.0).

CharlyF:
Issue pertains to Kubernetes dropping support of PSP in 1.25. We need to fix this separately 
```
>>> helm install datadog-ji03ar6ktl charts/datadog --namespace datadog-ji03ar6ktl --wait --values charts/datadog/ci/disable-apparmor-values.yaml --timeout 300s
Error: INSTALLATION FAILED: unable to build kubernetes objects from release manifest: resource mapping not found for name: "datadog-ji03ar6ktl" namespace: "" from "": no matches for kind "PodSecurityPolicy" in version "policy/v1beta1"
```

brycekahle (approved):

PR #814, opened 2022-11-17
feat: add DD_EXTERNAL_METRICS_PROVIDER_MAX_AGE and DD_EXTERNAL_METRIC…

bellondr:
…S_PROVIDER_BUCKET_SIZE for cluster agent

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -156,6 +156,10 @@ spec:
                 key: app-key
           - name: DD_EXTERNAL_METRICS_PROVIDER_ENABLED
             value: {{ .Values.clusterAgent.metricsProvider.enabled | quote }}
+          - name: DD_EXTERNAL_METRICS_PROVIDER_MAX_AGE
+            value: {{ .Values.clusterAgent.metricsProvider.maxAge | quote }}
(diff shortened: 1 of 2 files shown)


karl18:
Any updates about this PR ? Thanks ! I believe I'm stuck a bit until it's merged for increasing the max_age & bucket_size ?

CharlyF:
Hello folks - Very sorry for the delay here, have you tried using the `.Values.clusterAgent.env`?
My concern with having these default values set in the helm chart is that if the defaults need to change in the binary that will cause a configuration discrepancy that could yield a long tail of friction for users (that need to update multiple versions to solve one issue).
[The defaults](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config.go#L964) in the cluster agent are:
- max_age: 120s
- bucket_size: 300s
 
I am wondering if you might be getting the logs because the values are too sparse? It is usually recommended to use metrics that are submitted more than once a minute given the reconciliation loop of the HPA controller. 

Here would be an example of how set these with the current chart:
```
clusterAgent:
  env:
  - name: DD_EXTERNAL_METRICS_PROVIDER_MAX_AGE
    value: "900"
  - name: DD_EXTERNAL_METRICS_PROVIDER_BUCKET_SIZE
    value: "900"
```

fanny-jiang:
Bumping @bellondr @karl18; did Charly's response help?

Issue #827, opened 2022-12-04
apiKeyExistingSecret does not take precedence. 


Arvindh-Mojo:
Problem: 
Trying to figure out why despite the fact that i'm using the `apiKeyExistingSecret` it is still not taking precedence over "apiKey". 

my secrets are in AWS secret manager and I have successfully setup and verified that the csi driver and aws provider secret store for eks is working fine. `SecretProviderClass` is used to fetch and load the secrets. 

```
 2022-12-04 06:14:21 UTC | CORE | ERROR | (pkg/forwarder/transaction/transaction.go:344 in internalProcess) | API Key invalid, dropping transaction for https://7-32-4-app.agent. │
│ datadoghq.com/api/v1/series?api_key=<DATADOG_API_KEY>
```

It seems like the value `<DATADOG_API_KEY>` is coming it as a value from your `values.yaml` should it be commented like `#<DATADOG_API_KEY` ??

https://github.com/DataDog/helm-charts/blob/caa16f0f9d3faba9f1a35e6e3a1cb8d1fd0a4741/charts/datadog/values.yaml#L33

This is why I see helm chart also creating datadog-agent & cluster-agent secret, which shouldn't be happening in my case. 

Expected behavior:
Since I'm using existingapikeys, helm chart should not create a secret but rather use the one that already exists. 

values.yaml

```
datadog:
  apiKeyExistingSecret: datadog-api-key
  appKeyExistingSecret: datadog-api-key
  clusterName: shared-services
  tags:
    - "environment": "dev"
    - "cluster": "my-eks-cluster"
  registry: "public.ecr.aws/datadog"

  agents:
    rbac:
      create: false
      serviceAccountName: datadog-sa
    volumeMounts:
      - name: "aws-secrets-inline"
        mountPath: "/mnt/secret-store"
        readOnly: true
    env:
      - name: datadog-agent
        value:
          secretKeyRef:
            name: api-key
            key: api-key
    volumes:
      - name: "aws-secrets-inline"
        csi:
          driver: "secrets-store.csi.k8s.io"
          readOnly: true
          volumeAttributes:
            secretProviderClass: datadog-api-credentials
  clusterAgent:
    rbac:
      create: false
      serviceAccountName: datadog-sa
    volumeMounts:
      - name: "aws-secrets-inline"
        mountPath: "/mnt/secret-store"
        readOnly: true
    env:
      - name: datadog-cluster
        value:
          secretKeyRef:
            name: api-key
            key: api-key
[...]

PR #828, opened 2022-12-04
Comment Value for apiKey

svarvindh:
apiKey value is getting defaulted to <DATADOG_API_TOKEN> as its value.

apiKey value is coming in as <DATADOG_API_KEY>, because it is not commented and helm thinks its an actual value. Check my issue https://github.com/DataDog/helm-charts/issues/827

  - fixes # https://github.com/DataDog/helm-charts/issues/827

the value should be commented. Seems to be typo.

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.5.1
+
+* Removing default value placeholder for the API Key in the values.yaml.
+
 ## 3.5.0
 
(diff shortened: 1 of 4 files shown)


CharlyF:
@svarvindh & @Arvindh-Mojo  hey folks - Thank you very much for opening this up!
Do you mind bumping the [chart bugfix version](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/Chart.yaml#L3)  (and running `.github/helm-docs.sh`)?

PR #834, opened 2022-12-08
add DD_TAGS to clusterchecks

### LINKED ISSUE #831: metrics collected with kubeStateMetricsCore doesn't apply global tags when useClusterCheckRunners enabled
**Describe what happened:**

Global Tags not applied to metrics collected with kubeStateMetricsCore when useClusterCheckRunners is enabled.

Global Tags defined using the `datadog.tags` helm value would be applied to metrics collected using `kubeStateMetricsCore`

Set the following helm values:

```
datadog.tags: "team: my-team"
clusterChecksRunner.enabled: true
datadog.kubeStateMetricsCore.useClusterCheckRunners: true
```

After the chart is applied, the `datadog-clusterchecks` pods do not have the `DD_TAGS` environment variable set in their yaml. I presume this is why the tags are not being applied to metrics collected by these pods.

`DD_TAGS` is set on the node and cluster agent pods.

- GKE

marksmithson:

Adds DD_TAGS to clusterChecks pods. Ensures global tags are attached to metrics collected by this pod

  - fixes #831

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/agent-clusterchecks-deployment.yaml
+++ b/charts/datadog/templates/agent-clusterchecks-deployment.yaml
@@ -147,6 +147,10 @@ spec:
           - name: DD_LOG_LEVEL
             value: {{ .Values.datadog.logLevel | quote }}
           {{- end }}
+          {{- if .Values.datadog.tags }}
+          - name: DD_TAGS
+            value: {{ tpl (.Values.datadog.tags | join " " | quote) . }}
+          {{- end }}
           - name: DD_EXTRA_CONFIG_PROVIDERS
             value: "clusterchecks"
           - name: DD_HEALTH_PORT


jtappa (approved):
👍🏻 from docs team for changelog

Scoobed:
Is this correct based on this text -- https://docs.datadoghq.com/containers/cluster_agent/clusterchecks/?tab=helm#example-http_check-on-a-kubernetes-service

Note: With cluster checks, the metrics reported by the Agent are not linked to a given hostname because they are meant to be cluster-centric metrics and not necessarily host-based metrics. As a result, these metrics do not inherit any host-level tags associated with that host, such as those inherited from a cloud provider or added by the Agent’s DD_TAGS environment variable. To add tags to cluster check metrics, use the DD_CLUSTER_CHECKS_EXTRA_TAGS environment variable.

Which I don't see DD_CLUSTER_CHECKS_EXTRA_TAGS on the Helm Chart and not sure if DD_TAGs will be enough

rafaelromcar-parabol:
Actually, using the current version of the chart 3.35.1, the variable DD_TAGS is already present in the Cluster Agents, but as it is said in the [documentation](https://docs.datadoghq.com/containers/cluster_agent/clusterchecks/?tab=helm#set-up-cluster-check-dispatching) this is not valid. It is required to use `DD_CLUSTER_CHECKS_EXTRA_TAGS`.

We need a new variable for tags, dedicated only to the Cluster Agent. Something like
```
clusterAgent:
  tags: []
```

Or 
```
datadog:
  tags: []
  clusterAgentTags: []
```

The Datadog team should decide which one they prefer.

iamsaso:
I added `DD_CLUSTER_CHECKS_EXTRA_TAGS` but the tags are not being populated, and it looks like this is not working as expected

PR #878, opened 2023-01-11
[CWS] Adapt the chart to reflect CWS activity dumps parameters

spikat:

This will sync the available activity dumps parameters and their default values between the agent config file and the helm charts

  - fixes #

#### Checklist
- [ ] Chart Version bumped


alai97 (approved):
Looks good for docs!

brycekahle (approved):

CharlyF (changes requested):
This seems like a breaking change, so we can't merge as is.
Can you add a deprecation notice for the options you want to remove similar to here https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/NOTES.txt#L130-L157.
In the description of the PR you mention aligning this with the default in the agent, are cgroupdDumpTimeout and tracedCgroupsCount set to 30 and 5 respectively in the agent's/sys-probe config? (if so, is there a version dependency).

With regards to the options you want to remove, what is the experience going to be like for users that are using them at the moment?

spikat:
Hi @CharlyF, thanks for your review.

> This seems like a breaking change, so we can't merge as is. Can you add a deprecation notice for the options you want to remove similar to here https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/NOTES.txt#L130-L157. 

I've added the missing deprecation notices [here](https://github.com/DataDog/helm-charts/pull/878/commits/c187c22b1558707749bd45d69eafb23e74a7351a), tell me if we are good for this.

> In the description of the PR you mention aligning this with the default in the agent, are cgroupdDumpTimeout and tracedCgroupsCount set to 30 and 5 respectively in the agent's/sys-probe config? (if so, is there a version dependency).

Yes, [they are](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config.go#L1176-L1178).

> With regards to the options you want to remove, what is the experience going to be like for users that are using them at the moment?

First of all, for some context, all of these 4 config fields are related to a feature in private beta and it's not GA yet.

The config `datadog.securityAgent.runtime.activityDump.cgroupWaitListSize` does not exist anymore. If a user will launch the agent with this field configured, it will generate a warning log message like:
`
2023-01-18 11:40:34 CET | SYS-PROBE | WARN | (pkg/util/log/log.go:618 in func1) | Unknown key in config file: runtime_security_config.activity_dump.cgroup_wait_list_size
`
For the second one, `datadog.securityAgent.runtime.activityDump.pathMerge.enabled`: it still exists for now but that may change on future. So, for now if an user launch the agent with this config, it will be taken into account.

More globally:
There is a lot of config fields on CWS, and we don't want to expose them all, either from the helm charts or from the [config template](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml).
The configuration fields which we want to expose/document shouldn't change a lot, but yes they absolutely should be aligned between the helm charts and the config template.

spikat:
Gentle bump on this @CharlyF

PR #879, opened 2023-01-11
Allow env vars to be defined in a dict

### LINKED ISSUE #850: [helm-charts/charts/datadog] Support for generic environment variables defined as a map
- When using Helm values files in a hierarchical manner, any instances of lists are overwritten by values files further down the hierarchy
- This means that we have to duplicate environment variables in multiple places when we want to add a new one
- This behaviour is discussed [here](https://github.com/elastic/helm-charts/issues/53) for another non-Datadog helm chart

Since `.env` and `.envFrom` are already defined, I suggest creating an `.envMap` value which will allow the following:

```yaml
datadog:
  envMap:
    DD_ENVIRONMENT_VARIABLE: "VALUE"
```

Then this would be templated as below:

```
  env:
    {{- range $key, $value := .Values.datadog.envMap }}
    - name: {{ $key }}
      value: {{ $value | quote }}
    {{- end }}
```

Since this is an extension to env vars, after some analysis I found that the following would need to be created in total

```
datadog.envMap
agents.containers.agent.envMap
agents.containers.processAgent.envMap
agents.containers.securityAgent.envMap
agents.containers.systemProbe.envMap
agents.containers.traceAgent.envMap
clusterAgent.envMap
clusterChecksRunner.envMap
```

Additionally, this will allow us to prevent having to constantly add new values to the helm chart when future functionality is released, users can add the config via the helm values:

```yaml
agents:
  containers:
    agent:
      envMap:
        DD_SOME_FUTURE_ENV_VAR: VALUE
```

Note: this enhancement was originally discussed in:
- #767

jdheyburn:

* Add the following configurations which allow environment variables to be defined in a dictionary:
  * `agents.containers.agent.envDict`
  * `agents.containers.processAgent.envDict`
  * `agents.containers.securityAgent.envDict`
  * `agents.containers.systemProbe.envDict`
  * `agents.containers.traceAgent.envDict`
  * `clusterAgent.envDict`
  * `clusterChecksRunner.envDict`
  * `datadog.envDict`

  - fixes #850

Context in linked issue

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_helpers.tpl
+++ b/charts/datadog/templates/_helpers.tpl
@@ -656,6 +656,16 @@ Returns env vars correctly quoted and valueFrom respected
 {{- end -}}
 
 {{/*
+Returns env vars correctly quoted defined in a dict
+*/}}
+{{- define "additional-env-dict-entries" -}}
+{{- range $key, $value := . }}
+- name: {{ $key }}
+  value: {{ $value | quote }}
+{{- end }}
+{{- end -}}
+
+{{/*
 Return the appropriate apiVersion for PodDisruptionBudget policy APIs.
 */}}
 {{- define "policy.poddisruptionbudget.apiVersion" -}}
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Datadog changelog
 
+## 3.7.3
+
+* Add the following configurations which allow environment variables to be defined in a dictionary:
+  * `agents.containers.agent.envDict`
+  * `agents.containers.processAgent.envDict`
+  * `agents.containers.securityAgent.envDict`
+  * `agents.containers.systemProbe.envDict`
+  * `agents.containers.traceAgent.envDict`
+  * `clusterAgent.envDict`
+  * `clusterChecksRunner.envDict`
+  * `datadog.envDict`
+
 ## 3.7.2
 
 * Rename dogstatsd port on the Agent Service to match the name of the dogstatsd port in the Agent pod (`dogstatsd -> dogstatsdport`).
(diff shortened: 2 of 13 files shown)


alai97 (approved):
Looks good for docs!

charts/datadog/templates/_helpers.tpl
```diff
@@ -655,6 +655,16 @@ Returns env vars correctly quoted and valueFrom respected
 {{- end -}}
 {{- end -}}
 
+{{/*
+Returns env vars correctly quoted defined in a dict
+*/}}
+{{- define "additional-env-dict-entries" -}}
+{{- range $key, $value := . }}
+- name: {{ $key }}
+  value: {{ $value | quote }}
```
clamoriniere, line 664: what about `valueFrom` support? do you think you can also support it?

jdheyburn: Thanks for the suggestion @clamoriniere! I've added that functionality by using below:

```
{{- if kindIs "map" $value }}
{{ toYaml $value | indent 2 }}
{{- else }}
  value: {{ $value | quote }}
{{- end }}
```

Then you can use either of:

```yaml
envDict:
  MY_STRING_VAR: myValue
  MY_VAL_FROM_VAR:
    valueFrom:
      secretKeyRef:
        key: myKey
        name: mysecret
```

clamoriniere (commented):
hi @jdheyburn 

the change looks good.
could you add a test in https://github.com/DataDog/helm-charts/tree/main/charts/datadog/ci that try to add in one of the `envDict` a value and valueFrom 🙇 

then, I think we will be able to merge it.

jdheyburn:
@clamoriniere Thank you so much for your help pushing this through 😃

PR #891, opened 2023-01-23
[process-agent][helm-chart] Enable process discovery check by default

just-chillin:
This PR enables the process-discovery check in the process agent by default when using helm. Initally this setting was turned off by default because the minimum helm version was less than the version that the setting was introduced in. Now that the minimum version is 7.34, we can safely enable this setting by default.

  - fixes #

Original PR: https://github.com/DataDog/datadog-agent/pull/9082

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.8.1
+version: 3.8.2
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.8.2
+version: 3.9.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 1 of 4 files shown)


wiyu (changes requested):
You should add a [changelog](https://github.com/DataDog/helm-charts/blob/efab9fd0c856a5bf40b99c9c82f204cc1474c153/charts/datadog/CHANGELOG.md?plain=1#L625) entry and update this [README](https://github.com/DataDog/helm-charts/blob/efab9fd0c856a5bf40b99c9c82f204cc1474c153/charts/datadog/README.md?plain=1#L664).

wiyu (approved):

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.8.1
+version: 3.8.2
```
clamoriniere, line 3: I would suggest to bump the minor digit, since it changes a default behaviour

just-chillin: Makes sense. Fixed!

clamoriniere (approved):

CharlyF (approved):
🚢

PR #901, opened 2023-02-02
Adding logic to not mount auth-token volume in autopilot

CharlyF:

Fix templates for Autopilot.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -29,11 +29,9 @@
   resources:
 {{ toYaml .Values.agents.containers.systemProbe.resources | indent 4 }}
   volumeMounts:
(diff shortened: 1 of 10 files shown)


charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -29,9 +29,11 @@
   resources:
 {{ toYaml .Values.agents.containers.systemProbe.resources | indent 4 }}
   volumeMounts:
+    {{- if (not .Values.providers.gke.autopilot) }}
```
clamoriniere, line 32: IMO this change is not needed since system-probe is disabled in autopilot

CharlyF: yes indeed

clamoriniere (approved):

PR #923, opened 2023-02-21
add a parameter to enable remote configuration

YoannGh:

This PR adds a parameter to enable the remote configuration service of the agent.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/system-probe-configmap.yaml
+++ b/charts/datadog/templates/system-probe-configmap.yaml
@@ -53,6 +53,8 @@ data:
         enabled: {{ $.Values.datadog.securityAgent.runtime.syscallMonitor.enabled }}
       network:
         enabled: {{ $.Values.datadog.securityAgent.runtime.network.enabled }}
+      remote_configuration:
+        enabled: {{ if and .Values.datadog.securityAgent.runtime.enabled .Values.datadog.remoteConfiguration.enabled -}} true {{else -}} false {{end}}
 {{- if .Values.datadog.securityAgent.runtime.activityDump.enabled }}
       activity_dump:
         enabled: true
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,19 +1,23 @@
 # Datadog changelog
 
-# 3.20.3
+## 3.21.0
+
+* Add `datadog.remoteConfiguration.enabled` parameter to enable remote configuration.
+
+## 3.20.3
 
 * Fix command script in linux init container to prevent blocking deployment in GKE Autopilot on Rapid release channel.
 * Only mount DogStatsD socket in non-Autopilot environments.
 
-# 3.20.2
+## 3.20.2
 
 * Fix R/W volume mounts for CRI on Windows
 
-# 3.20.1
+## 3.20.1
 
 * Fix command args in linux init container to prevent blocking deployment in GKE Autopilot.
 
-# 3.20.0
+## 3.20.0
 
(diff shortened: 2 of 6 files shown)


lebauce (approved):

brycekahle (approved):

charts/datadog/templates/_components-common-env.yaml
```diff
@@ -60,4 +60,8 @@
 - name: DD_EXCLUDE_PAUSE_CONTAINER
   value: "false"
 {{- end }}
+{{- if $.Values.datadog.securityAgent.runtime.remoteConfiguration.enabled }}
+- name: DD_REMOTE_CONFIGURATION_ENABLED
+  value: "true"
+{{- end }}
```
clamoriniere, line 66: is it needed in every container?

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -53,6 +53,8 @@ data:
         enabled: {{ $.Values.datadog.securityAgent.runtime.syscallMonitor.enabled }}
       network:
         enabled: {{ $.Values.datadog.securityAgent.runtime.network.enabled }}
+      remote_configuration:
+        enabled: {{ $.Values.datadog.securityAgent.runtime.remoteConfiguration.enabled }}
```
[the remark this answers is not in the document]
YoannGh: The config option in this file and `DD_REMOTE_CONFIGURATION_ENABLED` are unrelated. This option is only used to tell the system-probe container to receive policy updates through remote config.
`DD_REMOTE_CONFIGURATION_ENABLED` isn't used directly because we don't want system-probe to use remote config by default, even if it's enabled at the core-agent level.

charts/datadog/templates/_components-common-env.yaml
YoannGh: It is required by the core agent container in order to turn on remote configuration. For other containers it provides an easy way to test whether remote config is enabled at the core-agent level.

clamoriniere: since remove config will be used by different products. It can be good to create a function in `_helper.tpl` that will be use to know if the envvar needs to be added, something like 

```suggestion
{{- if eq (include "remote-configuation-enabled" .) "true" }}
- name: DD_REMOTE_CONFIGURATION_ENABLED
  value: "true"
{{- end }}
```

YoannGh: Thanks, I added the `should-enable-remote-configuration` function in `_helper.tpl`. Other products that want to enable remote configuration will just have to add an OR condition to this function.

CharlyF (approved):
Looks good - Just note that there is a `.` missing in the README.md, can you fix it for the CI to be 🟢

laurentdelosieresfactorial:
Hello,

After enabling the remote configuration using the latest Helm template, the Datadog agent is complaining about the absence of the "remote_configuration.refresh_interval" parameter. See log below.

`2023-03-27 11:21:21 UTC | CORE | WARN | (pkg/config/viper.go:209 in GetDuration) | failed to get configuration value for key "remote_configuration.refresh_interval": unable to cast <nil> of type <nil> to Duration`

Thank you in advance,
L.

YoannGh:
Hello, [this datadog-agent pull request](https://github.com/DataDog/datadog-agent/pull/15549) should fix this issue. It will be part of the next agent release (version 7.44). Thank you for reporting this.

PR #929, opened 2023-03-01
Add automountServiceAccountToken option

fanny-jiang:

Add `automountServiceAccountToken` to the agent, cluster-agent and clustercheck runner chart templates. By default, the option is set to True. This new option allows users to configure whether the created RBAC ServiceAccount token should be mounted.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/agent-clusterchecks-rbac.yaml
+++ b/charts/datadog/templates/agent-clusterchecks-rbac.yaml
@@ -16,6 +16,7 @@ subjects:
 ---
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.clusterChecksRunner.rbac.automountServiceAccountToken }}
 metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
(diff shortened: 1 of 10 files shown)


charts/datadog/templates/agent-clusterchecks-rbac.yaml
```diff
@@ -16,6 +16,7 @@ subjects:
 ---
 apiVersion: v1
 kind: ServiceAccount
+automountServiceAccountToken: {{ .Values.clusterChecksRunner.rbac.automountServiceAccountToken | default "true" }}
```
clamoriniere, line 19: I'm wondering if `"true"` is correct for the default value. maybe `true` is enough

```suggestion
automountServiceAccountToken: {{ .Values.clusterChecksRunner.rbac.automountServiceAccountToken | default true }}
```

if if doesn't work you will need to do something like

```yaml
{{- if not .Values.clusterChecksRunner.rbac.automountServiceAccountToken }}
automountServiceAccountToken: false
{{- end }}
```

fanny-jiang: Thanks! Fixed

CharlyF (approved):

clamoriniere (approved):

PR #932, opened 2023-03-03
Update volumeMounts with `readOnly` fields

fanny-jiang:
Add `readOnly` option to all mounted volumes to improve security of the datadog-agent installation. Makes sure volumes are `readOnly: true` where possible and exceptions are noted with a reason. 

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -55,8 +55,10 @@
     - name: sysprobe-config
       mountPath: /etc/datadog-agent/system-probe.yaml
       subPath: system-probe.yaml
+      readOnly: true
     - name: sysprobe-socket-dir
       mountPath: /var/run/sysprobe
+      readOnly: false
     - name: procdir
       mountPath: /host/proc
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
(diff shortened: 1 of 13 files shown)


clamoriniere (approved):

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -103,9 +109,10 @@
     - name: runtime-compiler-output-dir
       mountPath: {{ .Values.datadog.systemProbe.runtimeCompilationAssetDir }}/build
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
+      readOnly: true
```
L3n41c, line 112: According to the name of the volume, this seems to be the path where `system-probe` writes the result of the eBPF programs compilation.
That’s why I’d be surprised if `system-probe` doesn’t fail if it cannot write there.

L3n41c (approved):

vladbeha109:
Hi,

After these changes the init-volume container cannot be initialised, it fails with the "Access denied" error. As a result, the next steps fail as well.
Once we unmarked "Read Only", it starts to work.
image

L3n41c:
Hello @vladbeha109,
Thank you for your feedback and sorry for the bug we introduced.
It should hopefully be fixed by #946.

PR #944, opened 2023-03-10
Auto-configure admissionController.configMode based on apm.socketEnabled|portEnabled

fanny-jiang:

Auto-configure `clusterAgent.admissionController.configMode` based on `datadog.apm.socketEnabled|portEnabled` for a better user experience. 

* If `datadog.apm.socketEnabled: true`, then `socket` config mode is used
* If `datadog.apm.portEnabled: true`, then `service` config mode is used

Default config mode is `socket` if `clusterAgent.admissionController.configMode` is not set.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -178,13 +178,17 @@ spec:
             value: {{ .Values.clusterAgent.admissionController.mutateUnlabelled | quote }}
           - name: DD_ADMISSION_CONTROLLER_SERVICE_NAME
             value: {{ template "datadog.fullname" . }}-cluster-agent-admission-controller
-          {{- if .Values.clusterAgent.admissionController.configMode }}
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_MODE
-            value: {{ .Values.clusterAgent.admissionController.configMode | quote }}
-          {{- if eq .Values.clusterAgent.admissionController.configMode "service" }}
+            {{- if or (eq (default .Values.datadog.apm.portEnabled false) true) (eq (default .Values.datadog.apm.enabled false) true) }}
+            value: service
+            {{- else if (eq (default .Values.datadog.apm.socketEnabled true) true) }}
+            value: socket
+            {{- else }}
+            value: {{ .Values.clusterAgent.admissionController.configMode | default "socket" }}
+            {{- end }}
+          {{- if or (eq (default .Values.clusterAgent.admissionController.configMode "socket") "service") (eq (default .Values.datadog.apm.portEnabled false) true) (eq (default .Values.datadog.apm.enabled false) true)}}
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_LOCAL_SERVICE_NAME
             value: {{ template "localService.name" . }}
-          {{- end }}
           {{- end }}
           {{- if .Values.providers.aks.enabled }}
           - name: DD_ADMISSION_CONTROLLER_ADD_AKS_SELECTORS
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -179,12 +179,12 @@ spec:
           - name: DD_ADMISSION_CONTROLLER_SERVICE_NAME
             value: {{ template "datadog.fullname" . }}-cluster-agent-admission-controller
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_MODE
-            {{- if or (eq (default .Values.datadog.apm.portEnabled false) true) (eq (default .Values.datadog.apm.enabled false) true) }}
+            {{- if .Values.clusterAgent.admissionController.configMode }}
+            value: {{ .Values.clusterAgent.admissionController.configMode | default "socket" }}
+            {{- else if or (eq (default .Values.datadog.apm.portEnabled false) true) (eq (default .Values.datadog.apm.enabled false) true) }}
             value: service
             {{- else if (eq (default .Values.datadog.apm.socketEnabled true) true) }}
             value: socket
-            {{- else }}
-            value: {{ .Values.clusterAgent.admissionController.configMode | default "socket" }}
             {{- end }}
           {{- if or (eq (default .Values.clusterAgent.admissionController.configMode "socket") "service") (eq (default .Values.datadog.apm.portEnabled false) true) (eq (default .Values.datadog.apm.enabled false) true)}}
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_LOCAL_SERVICE_NAME
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+# 3.20.0
+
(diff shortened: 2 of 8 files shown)


ahmed-mez (commented):
Thanks for working on this! One question about service vs hostip

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -178,14 +178,20 @@ spec:
...
           - name: DD_ADMISSION_CONTROLLER_SERVICE_NAME
             value: {{ template "datadog.fullname" . }}-cluster-agent-admission-controller
-          {{- if .Values.clusterAgent.admissionController.configMode }}
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_MODE
-            value: {{ .Values.clusterAgent.admissionController.configMode | quote }}
-          {{- if eq .Values.clusterAgent.admissionController.configMode "service" }}
+            {{- if .Values.clusterAgent.admissionController.configMode }}
+            value: {{ .Values.clusterAgent.admissionController.configMode }}
+            {{- else if or (eq .Values.datadog.apm.portEnabled true) (eq .Values.datadog.apm.enabled true) }}
+            value: service
```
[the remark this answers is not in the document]
fanny-jiang: I'll need your help with this one! The Jira cards for this requested change mentioned this order of priority and I interpreted the config modes as: 

```
* Local Service Traffic if available  # service config mode
* TCP if hostPort Enable                  # service config mode
* else UDS                                         # socket config mode
```

Now that I'm looking at the Jira cards again I think it should be:

```
* Local Service Traffic if available  # service config mode if agents.localService.forceLocalServiceEnabled: true OR K8s 1.22+
* TCP if hostPort Enable                  # hostip config mode
* else UDS                                         # socket config mode
```

^ Is that correct? Let me know what you think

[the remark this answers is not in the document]
ahmed-mez: Thanks for sharing more context!
Yea the second ordering makes more sense! Although I'm not 100% sure we should prioritize service over hostip. Historically hostip was the only mode and doesn't depend on dns nor on the k8s version. Users who want service can always configure it explicitly, that said, I'd follow this logic:
```
if configMode is not set
  if socket enabled, configMode == socket
  else if port enabled, configMode == hostip
```
The order matters, we want to chose the socket mode if both socket and port are enabled.

Any thoughts @yoannrobin ? Since you're familiar with this topic.

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -178,14 +178,24 @@ spec:
...
+            {{- else }}
+            value: hostip
+            {{- end }}
+            {{- else if eq (include "trace-agent-use-uds" .) "true" }}
+            value: socket
+            {{- else }}
+            value: socket
+            {{- end }}
+          {{- if or (eq .Values.clusterAgent.admissionController.configMode "service") (and (eq (include "trace-agent-use-tcp-port" .) "true") (eq (include "enable-service-internal-traffic-policy" .) "true")) }}
```
ahmed-mez, line 195: This condition is long and difficult to read, it's completely fine to remove it and always set `DD_ADMISSION_CONTROLLER_INJECT_CONFIG_LOCAL_SERVICE_NAME` regardless of the mode.

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -178,14 +178,24 @@ spec:
...
-          {{- if eq .Values.clusterAgent.admissionController.configMode "service" }}
+            {{- if .Values.clusterAgent.admissionController.configMode }}
+            value: {{ .Values.clusterAgent.admissionController.configMode }}
+            {{- else if eq (include "trace-agent-use-tcp-port" .) "true" }}
+            {{- if eq (include "enable-service-internal-traffic-policy" .) "true" }}
+            value: service
+            {{- else }}
+            value: hostip
+            {{- end }}
+            {{- else if eq (include "trace-agent-use-uds" .) "true" }}
```
ahmed-mez, line 190: This is not working as intended, especially if both `socketEnabled` and `portEnabled` are `true`, we end up with the mode set to `service` or `hostip`, while the desired value is `socket`. We want to always use mode `socket` if `socketEnabled` is `true`.

Let me clarify a bit [my suggestion in the other thread](https://github.com/DataDog/helm-charts/pull/944/files?diff=unified&w=0#r1132566243), I suggest we simplify the logic and not set `service` automatically at all.
- If `configMode` is set by the user, just uses its value
- If `configMode` is not set
  - if `trace-agent-use-uds`, set config mode to `socket`
  - else if `trace-agent-use-tcp-port`, set config mode to `hostip`.
  - else do nothing and keep it not set (i.e fallback to the default value in cluster agent config)

It's a simple and a predictable logic, _safer_ as we don't depend auto set the mode to `service` (no DNS dependency). Also it's import to check `trace-agent-use-uds` before `trace-agent-use-tcp-port` to ensure `socket` wins if both socket and port are enabled.

fanny-jiang: Thanks, that makes sense!

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -178,14 +178,18 @@ spec:
...
+            {{- if .Values.clusterAgent.admissionController.configMode }}
+            value: {{ .Values.clusterAgent.admissionController.configMode }}
+            {{- else if eq (include "trace-agent-use-uds" .) "true" }}
+            value: socket
+            {{- else if eq (include "trace-agent-use-tcp-port" .) "true" }}
+            value: hostip
+            {{- else }}
             value: {{ .Values.clusterAgent.admissionController.configMode | quote }}
-          {{- if eq .Values.clusterAgent.admissionController.configMode "service" }}
+            {{- end }}
```
ahmed-mez, line 190: Can we add a 2 additional test cases: one with both socket and port enabled and another one with apm disabled. I suspect this code path might panic since configMode can be nil.

ahmed-mez (commented):
One final suggestion! LGTM otherwise! (need approval from code owners)

charts/datadog/templates/cluster-agent-deployment.yaml
fanny-jiang: I added the test test cases to `/ci` and they both enable `socket`. I added them as conditions too just to be sure.

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -178,14 +178,16 @@ spec:
...
           - name: DD_ADMISSION_CONTROLLER_INJECT_CONFIG_MODE
-            value: {{ .Values.clusterAgent.admissionController.configMode | quote }}
-          {{- if eq .Values.clusterAgent.admissionController.configMode "service" }}
+            {{- if .Values.clusterAgent.admissionController.configMode }}
+            value: {{ .Values.clusterAgent.admissionController.configMode }}
+            {{- else if eq (include "trace-agent-use-tcp-port" .) "true" }}
+            value: hostip
+            {{- else if eq (include "trace-agent-use-uds" .) "true" }}
+            value: socket
+            {{- end }}
```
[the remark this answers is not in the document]
fanny-jiang: It now defaults to the default values.yaml value or `""`

ahmed-mez:
@fanny-jiang Thanks! I'm not code owner I can't approve, you need 👀 from the container teams.

khewonc (approved):

PR #952, opened 2023-03-17
[observability pipelines worker] 1.2.0-rc.0 release

001wwang:

Helm Chart changes required by OPW 1.2.0

* ~closes [OPB-704](https://datadoghq.atlassian.net/browse/OPB-704)~
* closes [OPB-751](https://datadoghq.atlassian.net/browse/OPB-751) 

Drafted pending questions and OPW 1.2.0 release

#### Checklist
- [x] Chart Version bumped

[OPB-704]: https://datadoghq.atlassian.net/browse/OPB-704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OPB-751]: https://datadoghq.atlassian.net/browse/OPB-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


charts/observability-pipelines-worker/ci/all-values.yaml
```diff
@@ -1,3 +1,9 @@
+args:
+  - run
+  - --skip-key-validation
```
001wwang, line 3: Note, key validation work is still blocked on the OPW side -- if changes occur, this will need updating.

charts/observability-pipelines-worker/ci/dupe-ports-values.yaml
```diff
@@ -1,3 +1,8 @@
+args:
+  - run
+  - --skip-key-validation
+  - --pipelines-config
+  - "/etc/observability-pipelines-worker/pipelines.yaml"
```
001wwang, line 5: @spencergilbert let me know if this is correct/if there's a way to avoid this duplicated override.

charts/observability-pipelines-worker/templates/configmap.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     checksum/config: {{ tpl (toYaml .Values.config) . | sha256sum }}
 data:
   {{- if .Values.config }}
-  config.yaml: | {{ tpl (toYaml .Values.config) . | nindent 4 }}
+  pipelines.yaml: | {{ tpl (toYaml .Values.config) . | nindent 4 }}
   {{- else }}
 {{ fail "A `config` must be specified, see https://docs.datadoghq.com/observability_pipelines/reference for all options" }}
```
spencergilbert, line 13: The filename has become `pipelines.yaml`, do you still want the user to configure it's contents via a key `config`?

charts/observability-pipelines-worker/Chart.yaml
```diff
@@ -13,7 +13,7 @@ icon: https://datadog-live.imgix.net/img/dd_logo_70x75.png
 maintainers:
   - name: Datadog
     email: support@datadoghq.com
-appVersion: "1"
+appVersion: "1.2.0"
```
spencergilbert, line 16: Can technically leave this as-is if you want to change things in one less place. The Datadog Agent also just puts the major version here (it's not functionally used, just documentation)

charts/observability-pipelines-worker/templates/configmap.yaml
```diff
@@ -8,7 +8,7 @@ metadata:
     checksum/config: {{ tpl (toYaml .Values.config) . | sha256sum }}
 data:
   {{- if .Values.config }}
-  config.yaml: | {{ tpl (toYaml .Values.config) . | nindent 4 }}
+  pipelines.yaml: | {{ tpl (toYaml .Values.config) . | nindent 4 }}
```
001wwang, line 11: I'd like to get more familiar with the wiring up of the config. Is the following correct?

- Users must provide a config value
- The config is turned into a `ConfigMap` (i.e. here)
- The `ConfigMap` is made available in a volume at `/etc/observability-pipelines-worker/`
- OPW can then read from `/etc/observability-pipelines-worker/pipelines.yaml`

And if users already have their own config file at a different location, are they meant to override the `args` to use it?

charts/observability-pipelines-worker/ci/dupe-ports-values.yaml
spencergilbert: I _think_ this is right, without needing to change the shared CI workflow. We can also look to remove some of the cases, they may not all be needed with the more streamlined OPW experience.

charts/observability-pipelines-worker/templates/configmap.yaml
001wwang: Yes, good timing, I was about to comment here. I think `config` is too general, especially with bootstrap involved. Is there a good way to refactor this other than find/replace? I think `pipelinesConfig` (open to suggestions) may be better.

charts/observability-pipelines-worker/Chart.yaml
spencergilbert: Less updating/they're more strict on semver I'd assume. Vector used this `appVersion` to populate the `image.tag` - but I followed the existing DD chart's pattern so the setup would be more familiar for anyone reviewing PRs from the org.

001wwang:
@spencergilbert One large remaining piece is handling `bootstrap.yaml` (contents pending https://github.com/DataDog/observability-pipelines-worker/pull/66). We will be adding a default file to the image itself, but what does wiring (if any) look like for the Helm Chart (i.e. making it easy for users to set the environment variables, etc.)?

charts/observability-pipelines-worker/Chart.yaml
spencergilbert: TL;DR is y'alls choice between specificity/accuracy and desire to change one less line per release 😆

charts/observability-pipelines-worker/templates/configmap.yaml
spencergilbert: Yeah, that name seems fine to me? I'm terrible with names though. Find and replace should do it.

We'll probably need to make a couple more changes to account for the `bootstrap.yaml` however.

charts/observability-pipelines-worker/templates/configmap.yaml
spencergilbert: 1. User provides contents to the `config` key in the `values.yaml`.
2. `configmap.yaml` will take those contents try to parse them as YAML.
3. This [volume](https://github.com/DataDog/helm-charts/blob/0e9cc065f09c94cdc54ccdfc3e2f66a9e22620d7/charts/observability-pipelines-worker/templates/_pod.tpl#L117-L121) is created using the ConfigMap from 2.
4. This [volumeMount](https://github.com/DataDog/helm-charts/blob/0e9cc065f09c94cdc54ccdfc3e2f66a9e22620d7/charts/observability-pipelines-worker/templates/_pod.tpl#L84-L86) places the contents of the volume at `/etc/observability-pipelines-worker/` - the configuration we provided creates each key under `data` as a file there.
5. The provided file is accessible at `/etc/observability-pipelines-worker/pipelines.yaml` during runtime.

At this point the only file path the config will be placed is the above, and the only method of providing a config is through the `config` key. Do we think users actually need this supported? Off the top of my head I can't think of anything that would require that.

spencergilbert:
> @spencergilbert One large remaining piece is handling `bootstrap.yaml` (contents pending [DataDog/observability-pipelines-worker#66](https://github.com/DataDog/observability-pipelines-worker/pull/66)). We will be adding a default file to the image itself, but what does wiring (if any) look like for the Helm Chart (i.e. making it easy for users to set the environment variables, etc.)?

We have a couple of environment variables that can be specified directly [here](https://github.com/DataDog/helm-charts/blob/0e9cc065f09c94cdc54ccdfc3e2f66a9e22620d7/charts/observability-pipelines-worker/templates/_pod.tpl#L45-L56). We could provide a similar `config|pipelinesConfig` key to override the file, but if all of the options can currently be set via variables I feel like that's a more straightforward.

charts/observability-pipelines-worker/Chart.yaml
001wwang: Ok, in the spirit of following existing practices and since I don't have much expertise here, I'm going to leave this as is.

001wwang:
> We have a couple of environment variables that can be specified directly [here](https://github.com/DataDog/helm-charts/blob/0e9cc065f09c94cdc54ccdfc3e2f66a9e22620d7/charts/observability-pipelines-worker/templates/_pod.tpl#L45-L56). We could provide a similar config|pipelinesConfig key to override the file, but if all of the options can currently be set via variables I feel like that's a more straightforward.

Yes, environment variables is fine; we'll default bootstrap pre-filled with env vars. As mentioned elsewhere, I'm starting to see we should encourage users to interact with this setup via `values`.

Other than what's already here, we're missing `DD_ENDPOINT` and `DD_OP_DATA_DIR`. The `data_dir` volume that's created right now seems wired to the pipeline config value. I think we need to change this -- maybe have both since it's possible they could configure one in bootstrap config and another in the pipeline config (not sure why they would, but it's possible)?

001wwang:
I likely won't have a chance to update this personally before my out-of-office, so leaving a status update for now. This is awaiting the following changes at least

- Rename `config` in `values` to `pipelineConfig` throughout
- Ensure all environment variables in the default `bootstrap.yaml` can be easily set (`DD_OP_DATA_DIR`, `DD_ENDPOINT` remaining)
- Any updates related to https://datadoghq.atlassian.net/browse/OPB-750

charts/observability-pipelines-worker/ci/all-values.yaml
001wwang: Changes have been merged on the OPW side.

charts/observability-pipelines-worker/templates/configmap.yaml
001wwang: I've updated the naming to `pipelineConfig` in [31a5536](https://github.com/DataDog/helm-charts/pull/952/commits/31a55360b1dcf5d876ff1087c50c54bcca9950e1) and beyond.

001wwang:
Hey @spencergilbert, I've brought this up to date with the following changes
- Updating `config` to `pipelineConfig`
- Adding values to set `endpoint` and `data_dir` for the bootstrap file
- Updating test configs, README, etc.

Let me know how that all looks. 

Thinking more, I'm wondering if we should offer a way to override the default bootstrap and allow a user to pass one optionally via a `bootstrapConfig` value. Do you think this would be a valid use case or would setting the environment variables that feed into the default bootstrap be sufficient?

charts/observability-pipelines-worker/templates/_pod.tpl
```diff
@@ -80,7 +84,7 @@ containers:
 {{- end }}
     volumeMounts:
       - name: data
-        mountPath: "{{ .Values.config.data_dir | default "/var/lib/observability-pipelines-worker" }}"
+        mountPath: "{{ .Values.datadog.data_dir | default "/var/lib/observability-pipelines-worker" }}"
```
spencergilbert, line 87: Curious how OPW handles things if a `data_dir` is specified in both the pipelines config and the bootstrap/opw side.

charts/observability-pipelines-worker/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+# 1.2.0
+
+* Update `run` subcommand to use new `--pipeline-config` argument
+* Rename `config` to `pipelineConfig` in values
+* Add `endpoint` and `data_dir` values
```
spencergilbert, line 7:
```suggestion
* Add `datadog.endpoint` and `datadog.data_dir` values
```

spencergilbert (commented):
> Let me know how that all looks.

Changes look good, small nits for technically correctness.
 
> Thinking more, I'm wondering if we should offer a way to override the default bootstrap and allow a user to pass one optionally via a `bootstrapConfig` value. Do you think this would be a valid use case or would setting the environment variables that feed into the default bootstrap be sufficient?

As I understand it, outside of Kubernetes, the user would override the values _primarily_ through the environment variables. I'd suggest leaving it just as environment variables unless someone requests it particularly - I can't think of anything that a file would solve that the env vars don't.

Keep it simple until it's needed, imho

charts/observability-pipelines-worker/templates/_pod.tpl
001wwang: As of https://github.com/DataDog/observability-pipelines-worker/pull/81, if `data_dir` is set in the pipeline config, we'll raise a validation error and OPW will fail to start e.g.
```
Configuration is invalid.
-------------------------
x Pipeline configuration must not contain data_dir setting. This should be set in the bootstrap configuration.
```

spencergilbert: Ah, that sounds familiar - just wanted to make sure that's covered.

spencergilbert:
RE: default `pipeline.yaml`, I think you'll just need to add a conditional around the `config` volume and volumeMount to not render those if the `pipelineConfig` is empty. Happy to help you get that done if you need to pair.

001wwang:
For posterity, @spencergilbert and I met to discuss updates required for using a default `pipeline.yaml`. I've pushed a few updates to handle basic cases where `.Values.pipelineConfig` is empty, and @spencergilbert will look into changes related to service ports.

001wwang:
Noting I'll make another update to take https://datadoghq.atlassian.net/browse/OPB-780 into account. In short, I think we'd like to change
- `configKey` to `pipelineID`
- `endpoint` to `dd_url`

Note, `DD_OP_CONFIG_KEY` and `DD_ENDPOINT` are still supported aliases for the newer `DD_OP_PIPELINE_ID` and `DD_URL`. I think we can similarly leave `configKey` and `endpoint` available here if backwards compatibility is needed.

Since we've already been going with env var based bootstrap configuration in this PR, I don't think anything else needs updating.

cc: @spencergilbert for context.

charts/observability-pipelines-worker/README.md
```diff
@@ -20,15 +20,15 @@ To install the chart with the release name `<RELEASE_NAME>` run:
...
 helm install --name <RELEASE_NAME> \
   --set datadog.apiKey=<DD_API_KEY> \
-  --set datadog.configKey=<DD_OP_CONFIG_KEY> \
+  --set datadog.pipelineID=<DD_OP_PIPELINE_ID> \
   datadog/observability-pipelines-worker
 ```
 
-By default, this chart creates secrets for your Observability Pipelines API and configuration keys. However, you can use
-manually created Secrets by setting the `datadog.apiKeyExistingSecret` and/or `datadog.appKeyExistingSecret` values
+By default, this chart creates secrets for your Observability Pipelines API key and pipeline ID. However, you can use
```
001wwang, line 27: @spencergilbert I made some updates in [48707e1](https://github.com/DataDog/helm-charts/pull/952/commits/48707e1d02ddbb51ac958161105f9a6bd2ad0280) and [30d5b0a](https://github.com/DataDog/helm-charts/pull/952/commits/30d5b0ad0c8c283bee223fda61379c252c671cab). Let me know how that looks.

I'm curious if we absolutely need secrets management for the config key/pipeline ID? It's essentially just a UUID identifier for a particular pipeline, so I'm not sure it's considered sensitive data. This is mostly for my own understanding, I think we're fine to leave as is for now.

001wwang:
Another heads up: we've decided to remove the default pipeline configuration, so I'll walk back some of my other changes and leave `pipelineConfig` as required.

charts/observability-pipelines-worker/README.md
spencergilbert: If it's just a UUID, it may not need to be a Secret. Could anything funky happen if an attacker reused the same ID?

I think I was under the impression that it was suppose to be a confidential value when I wrote this, if not it could just as easily be a ConfigMap.

charts/observability-pipelines-worker/README.md
```diff
@@ -20,15 +20,15 @@ To install the chart with the release name `<RELEASE_NAME>` run:
...
+manually created Secrets by setting the `datadog.apiKeyExistingSecret` and/or `datadog.pipelineIDExistingSecret` values
```
spencergilbert, line 28:
```suggestion
manually created Secrets by setting the `datadog.apiKeyExistingSecret` and/or `datadog.pipelineIdExistingSecret` values
```

If we're being super proper/pedantic it should be `Id`.

charts/observability-pipelines-worker/values.yaml
```diff
@@ -24,19 +24,28 @@ datadog:
...
+  pipelineIDExistingSecret: ""
   ## Set to 'datadoghq.com' to send data to the US1 site.
   ## Set to 'datadoghq.eu' to send data to the EU site.
   ## Set to 'us3.datadoghq.com' to send data to the US3 site.
   ## Set to 'us5.datadoghq.com' to send data to the US5 site.
   ## Set to 'ddog-gov.com' to send data to the US1-FED site.
   # datadog.site -- The [site](https://docs.datadoghq.com/getting_started/site/) of the Datadog intake to send data to.
   site: datadoghq.com
+  # datadog.ddURL -- The full [URL](https://docs.datadoghq.com/getting_started/site/) of the Datadog intake to send data to (e.g. https://api.datadoghq.com). This overrides the site setting.
+  ddURL: ""
```
spencergilbert, line 40:
```suggestion
  url: ""
```

I think this name looks better especially when written with the parent: `datadog.ddURL` vs `datadog.url`.

charts/observability-pipelines-worker/values.yaml
```diff
@@ -24,19 +24,28 @@ datadog:
   # datadog.configKeyExistingSecret -- Specify a preexisting Secret that has your configuration key instead of
   # creating a new one. The value must be stored under the `config-key`.
   configKeyExistingSecret: ""
+  # datadog.pipelineID -- Specify your Datadog Observability Pipelines pipeline ID
+  pipelineID:  # <DD_OP_PIPELINE_ID>
```
spencergilbert, line 28:
```suggestion
  pipelineId:  # <DD_OP_PIPELINE_ID>
```

Nit/pedantic

charts/observability-pipelines-worker/values.yaml
```diff
@@ -24,19 +24,28 @@ datadog:
...
+  # datadog.pipelineIDExistingSecret -- Specify a preexisting Secret that has your pipeline ID instead of
+  # creating a new one. The value must be stored under the `pipeline-id`.
+  pipelineIDExistingSecret: ""
```
spencergilbert, line 31:
```suggestion
  pipelineIdExistingSecret: ""
```

Nit/pedantic

charts/observability-pipelines-worker/templates/_helpers.tpl
```diff
@@ -47,6 +47,14 @@ Return the configuration key Secret name to be used based on provided values.
 {{- default $fullName .Values.datadog.configKeyExistingSecret | quote -}}
 {{- end -}}
 
+{{/*
+Return the pipeline ID Secret name to be used based on provided values.
+*/}}
+{{- define "opw.pipelineIDSecretName" -}}
+{{- $fullName := printf "%s-pipelineid" (include "opw.fullname" .) -}}
```
spencergilbert, line 54:
```suggestion
{{- define "opw.pipelineIdSecretName" -}}
{{- $fullName := printf "%s-pipeline-id" (include "opw.fullname" .) -}}
```

Nit/pedantic

charts/observability-pipelines-worker/templates/_helpers.tpl
```diff
@@ -47,6 +47,14 @@ Return the configuration key Secret name to be used based on provided values.
 {{- default $fullName .Values.datadog.configKeyExistingSecret | quote -}}
 {{- end -}}
 
+{{/*
+Return the pipeline ID Secret name to be used based on provided values.
```
spencergilbert, line 51: This replaces the `opw.configKeySecretName` partial above, correct? We should remove it if it's now unused.

charts/observability-pipelines-worker/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+# 1.2.0
+
+* Update `run` subcommand to use new `--pipeline-config` argument
+* Rename `config` to `pipelineConfig` in values
+* Add `datadog.pipeline_id` value to replace `datadog.config_key`
```
spencergilbert, line 7:
```suggestion
* Add `datadog.pipelineId` value to replace `datadog.configKey`
```

Or with the current styling of `pipelineID`, if you still prefer that. Probably also worth noting in the change log that `configKey` should still works for backwards compatibility?

spencergilbert (changes requested):

charts/observability-pipelines-worker/README.md
001wwang: 😢 Alas, I flip-flopped on this and landed on ID. Let's go for super proper 👍

001wwang: I'll bundle the renaming's into a single commit.

001wwang: Adjusted in [7a5e07c](https://github.com/DataDog/helm-charts/pull/952/commits/7a5e07c5b852e9393a9f49a75395fe08597c52dd)

charts/observability-pipelines-worker/values.yaml
001wwang: Adjusted in [7a5e07c](https://github.com/DataDog/helm-charts/pull/952/commits/7a5e07c5b852e9393a9f49a75395fe08597c52dd)

charts/observability-pipelines-worker/values.yaml
001wwang: Adjusted in [7a5e07c](https://github.com/DataDog/helm-charts/pull/952/commits/7a5e07c5b852e9393a9f49a75395fe08597c52dd)

charts/observability-pipelines-worker/templates/_helpers.tpl
001wwang: I think this is still necessary for backwards compatibility unless I'm misunderstanding something here.

charts/observability-pipelines-worker/CHANGELOG.md
001wwang: Sounds good, adjusted in [7a5e07c](https://github.com/DataDog/helm-charts/pull/952/commits/7a5e07c5b852e9393a9f49a75395fe08597c52dd)

charts/observability-pipelines-worker/values.yaml
001wwang: Yep, sounds good, updated in [cc13d58](https://github.com/DataDog/helm-charts/pull/952/commits/cc13d58f845d5c24df6b474fbad2d4925203a928)

charts/observability-pipelines-worker/README.md
001wwang: > Could anything funky happen if an attacker reused the same ID?

Off the top of my head, I don't think anything malicious can be done with solely the pipeline ID. 

> I think I was under the impression that it was suppose to be a confidential value when I wrote this, if not it could just as easily be a ConfigMap.

Right, I'm wondering if we could treat it like the other values under `datadog`. Perhaps we could simplify this in a follow up? It doesn't seem particularly urgent.

charts/observability-pipelines-worker/templates/_helpers.tpl
spencergilbert: Oops, yeah - I temporarily lost my mind and forgot about the backward compatibility here.

charts/observability-pipelines-worker/README.md
spencergilbert: Ah, yeah if it's determined there's not an issue exposing it we could transition it to similar to the `datadog.site` or `datadog.url` options.

I'd want to doublecheck the assumption before making that happen, for sure.

spencergilbert (approved):
The changes LGTM at this point

charts/observability-pipelines-worker/README.md
leebenson: > Off the top of my head, I don't think anything malicious can be done with solely the pipeline ID.

Just a drive-by...

Pipeline ID isn't intended to be sensitive. It's really just a resource ID, embedded in URLs and made public in UI without regard to obfuscation. Doesn't need to be a secret.

spencergilbert: Definitely sounds like something to change then, apologies for my misunderstanding!

001wwang: @spencergilbert, for simplicity's sake, how do we feel about
- Treat `pipelineId` similar to `datadog.___` options
- Remove the `configKey` secret entirely

as part of this release?

spencergilbert: No objections for the first. For the second, that may be considered a breaking change if users had been using the `existingConfigKey` options.

It would complicate the code a here a bit, but we could do `datadog.pipelinesId` or `datadog.configKey` goes directly in the Pod's `env` block (both creating the _new_ env var name), and `datadog.configKeyExistingSecret` still using a referenced Secret to populate it. Marking the two `configKey` options as deprecated and dropping next release?

001wwang: > No objections for the first.

Ok, I'll make some minor updates then for `pipelineId`.

> It would complicate the code a here a bit, but we could do datadog.pipelinesId or datadog.configKey goes directly in the Pod's env block (both creating the new env var name), and datadog.configKeyExistingSecret still using a referenced Secret to populate it. Marking the two configKey options as deprecated and dropping next release?

Sure, this seems reasonable. Would you be able to make the updates here?

To make sure I'm on the right page, we're planning to set `DD_OP_PIPELINE_ID` in the environment in order of precedence from 1) `datadog.pipelineId` 2) `datadog.configKey` 3) `datadog.configKeyExistingSecret`? And we can get rid of some current plumbing that creates a secret from `datadog.configKey`?

spencergilbert: Yeah, I can make some time tonight to do that. Your understanding sounds right, though I'll double check that there's no issues with `datadog.configKey` no longer creating a Secret - between versions.

spencergilbert: Pushed some changes in https://github.com/DataDog/helm-charts/pull/952/commits/34e8658df19b73bed271faa39d5b2c7036313c65, I left the secret as-is since I don't think Helm will natively remove the orphaned resources.

sghall:
@001wwang just wanted to note some feedback we received from @jszwedko in the [Feature Fest doc](https://docs.google.com/document/d/1aC5PNYvx0dc1xrdxTQyHUqNm6URgvnPAa8d0BifUdcE/edit#).  

> …when attempting to use datadog.pipelineId…it gives a warning that configKey isn’t set

Just want to make sure we have addressed that before merging this one 👍

sghall:
Also noting we need align on enabling RC.  Instructions PR up over here https://github.com/DataDog/web-ui/pull/89053

Instructions now look like...
Screenshot 2023-05-02 at 10 11 03 AM

001wwang:
Hey @spencergilbert, I pushed a few changes, fixing some things mentioned in :
- [efb3e02](https://github.com/DataDog/helm-charts/pull/952/commits/efb3e02b7ca4587252b1f65bfd6f932d8206ce53) Avoids the error message referencing `configKey` when `pipelineId` is set.
- [143ed0f](https://github.com/DataDog/helm-charts/pull/952/commits/143ed0f4aa80625d0b6b1abf08941dfaa26a6bd6), [9fc2d4a](https://github.com/DataDog/helm-charts/pull/952/commits/9fc2d4a1c0079168ddfe687ed1ce6608ac77f8d8) Add and wire up `datadog.remoteConfigurationEnabled`
- [aba6253](https://github.com/DataDog/helm-charts/pull/952/commits/aba6253344a5bca17ca3cc2a1c895d113c3c0089) Remove the explicit `--pipeline-config` in default args -- OPW should find the default mounted to `/etc/observability-pipelines-worker/pipeline.yaml` automatically. This avoids an error if remote config is enabled wherein `--pipeline-config` is then not allowed.
- [2ad4b3c](https://github.com/DataDog/helm-charts/pull/952/commits/2ad4b3c5cf20146814778afa5def91f78d3ca6d5) Updating this to be a release candidate rather than official 1.2.0.

Let me know how that looks. A few things specifically I'd appreciate a double check on:
- Conditionally rendering the `configmap` based on whether remote config is enabled
- Any other places `configKey` might accidentally be an issue despite `pipelineId` being set
- Any other updates necessary for release candidate
- Any updates we should make to any docs in this repo

charts/observability-pipelines-worker/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+# 1.2.0
```
spencergilbert, line 3:
```suggestion
# 1.2.0-rc.0
```

charts/observability-pipelines-worker/README.md.gotmpl
```diff
@@ -53,7 +53,7 @@ organization. This Secret is used in the manifest to deploy the Observability Pi
 export DATADOG_SECRET_NAME=datadog-secrets
 kubectl create secret generic $DATADOG_SECRET_NAME \
     --from-literal api-key="<DD_API_KEY>" \
-    --from-literal config-key="<DD_OP_CONFIG_KEY>"
+    --from-literal pipeline-id="<DD_OP_PIPELINE_ID>"
```
spencergilbert, line 56:
```suggestion
```

Since you're moving away from the pipelineId/configKey from being a secret you can probably drop this?

charts/observability-pipelines-worker/templates/_pod.tpl
```diff
@@ -47,13 +47,29 @@ containers:
           secretKeyRef:
             name: {{ template "opw.apiSecretName" . }}
             key: api-key
-      - name: DD_OP_CONFIG_KEY
+      - name: DD_OP_PIPELINE_ID
+        {{- with .Values.datadog.pipelineId }}
+        value: {{ . | quote }}
+        {{- else }}
```
spencergilbert, line 53:
```suggestion
      {{- with .Values.datadog.pipelineId }}
        value: {{ . | quote }}
      {{- else }}
```

Formatting nit, matches the indention on the lines below.

spencergilbert (changes requested):
I'd consider adding a new `ci/` file to test the behavior of the new RC bits since that's a pretty critical user path going forward.

spencergilbert (changes requested):

001wwang:
> I'd consider adding a new ci/ file to test the behavior of the new RC bits since that's a pretty critical user path going forward.

Yea, sounds good -- I think this should just involve a test config with `datadog.remoteConfigurationEnabled = true` and no pipeline config. 

On this note though, what exactly are the tests in `ci/` asserting? Is it just that the chart is installed successfully and all the k8s pieces are running?

spencergilbert:
> > I'd consider adding a new ci/ file to test the behavior of the new RC bits since that's a pretty critical user path going forward.
> 
> Yea, sounds good -- I think this should just involve a test config with `datadog.remoteConfigurationEnabled = true` and no pipeline config.
> 
> On this note though, what exactly are the tests in `ci/` asserting? Is it just that the chart is installed successfully and all the k8s pieces are running?

Yeah, just setting that flag to true. It's not _truly_ end-to-end, but it'll attempt to install/upgrade the chart using those values so if we have any invalid resources or say the pod still trying to mount the non-existing configmap it'll error. If the image itself crashes due to error (for whatever reason), that'll also fail the CI job.

spencergilbert (approved):

001wwang:
Note for posterity, I squashed all commits to satisfy a repo requirement for signed commits (my previous commits were not signed).

PR #959, opened 2023-03-21
[DCA flare] Injects additional useful env vars

tbavelier:
#### What this PR does :
* Injects some env variables in the DCA to determine : 
    * The Helm release name
    * The generated Daemonset name of the node Agent
    * The generated Deployment name of the cluster Agent

#### Motivation:
* These values will be used to retrieve the necessary resources from the API server to add to the DCA flare user configuration, as well as manifests of the node Agent/cluster Agent : https://datadoghq.atlassian.net/browse/CONT-3734

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-crds/CHANGELOG.md
+++ b/charts/datadog-crds/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 0.6.0
+
+* Support Certificate Manager.
+* Document conversion webhook configuration.
+
 ## 0.5.9
 
 * Updating DatadogMonitors CRD and DatadogAgents CRDs.


charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -267,6 +267,12 @@ spec:
             value: {{ .Values.datadog.prometheusScrape.version | quote }}
           {{- end }}
           {{- end }}
+          - name: DD_CHART_RELEASE_NAME
```
vboulineau, line 270: nit: `DD_` is used for Agent parameters, meaning that you will have log lines about unknown parameters due to that.
As these are not read from Agent configuration, I'd remove the `DD_`

PR #1053, opened 2023-05-23
Make `dnsPolicy` configurable for agents

### LINKED ISSUE #755: Datadog Fails DNS Lookups
Hi,

We have the [Datadog helm chart](https://github.com/DataDog/helm-charts/tree/main/charts/datadog) deployed in our AWS EKS running in an EC2 container. The dnsPolicy given to each Pod (via Deployment) is the default "ClusterFirst" value. We need to override this to "Default" so that it inherits the nameServer from the EC2. Otherwise Datadog fails to post any data to Datadog until we manually update it. However, there is not a value in Values.yaml that allows us to do this.

We are using a custom core_dns implementation and once Datadog uses nameServer 10.0.0.2 all networking is successful. We cannot yet explain why the default value 172.20.0.10 fails when it is also successful on another pod e.g. a basic Nginx. For some reason Datadog does not like this value and must use 10.0.0.2

Anyway, could a value please be added to [Values.yaml ](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml) allowing us to specify a dnsPolicy? Similar to dnsConfig. This'll also require at minimum change in [ClusterAgent](https://github.com/DataDog/helm-charts/blob/6d12d9c5614834761a8d1e6867a06406f5b0651a/charts/datadog/templates/cluster-agent-deployment.yaml#L83) and [Agent Daemonset](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/daemonset.yaml#L92).

Thanks

_I'm happy to make a PR for this. This resolution allows us to stop patching ClusterAgent and Agent after deployment, or manually editing the helm_charts and having to keep doing so for every update._

rpriyanshu9:

Makes `dnsPolicy` configurable for agents using `agents.dnsPolicy` variable in `values.yaml`.

  - fixes #755 

#### Checklist
- [x] Chart Version bumped


rpriyanshu9:
Hey! @brycekahle @clamoriniere 👋🏼 
Could you please take a look at this PR?

rpriyanshu9:
👋🏼  @truthbk @clamoriniere Could you please take a look?

PR #1058, opened 2023-05-24
upgrade(remoteConfig): Add top-level enablement parameter

BaptisteFoy:
Adds a top-level parameter to enable Remote Configuration both on the datadog agent and on the cluster agent. This top-level parameter will override the two existing parameters that enable Remote Configuration on the two agents separately.

This will avoid confusion in the scenarii where Remote Configuration is enabled for one but not the other.

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/_containers-common-env.yaml
```diff
@@ -9,7 +9,7 @@
     secretKeyRef:
       name: {{ template "datadog.apiSecretName" . }}
       key: api-key
-{{- if .Values.datadog.remoteConfiguration.enabled }}
+{{- if or (.Values.datadog.remoteConfiguration.enabled .Values.remoteConfiguration.enabled) }}
```
mellon85, line 12: can't we set a variable to all these conditions with an `or` and use that variable?
One for the controller, and one for the agent. So we have a common global definition to use in these places

BaptisteFoy: Not a big fan of what this makes us do with the string comparison, but done in [20303ce](https://github.com/DataDog/helm-charts/pull/1058/commits/20303ce608848ebe087e9f6b489f2ed3138384ed).

mellon85 (approved):

mellon85 (approved):

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.30.6
+version: 3.30.7
```
AliDatadog, line 3: I think we need to update the minor version when a parameter is added

charts/datadog/values.yaml
```diff
@@ -1935,3 +1936,9 @@ providers:
   aks:
     # providers.aks.enabled -- Activate all specifities related to AKS configuration. Required as currently we cannot auto-detect AKS.
     enabled: false
+
+remoteConfiguration:
+  # remoteConfiguration.enabled -- Set to true to enable remote configuration on the cluster agent (if set) and the datadog agent.
```
celenechang, line 1941:
```suggestion
  # remoteConfiguration.enabled -- Set to true to enable remote configuration on the Cluster Agent (if set) and the node agent.
```

charts/datadog/Chart.yaml
BaptisteFoy: Fixed in [e9615a0](https://github.com/DataDog/helm-charts/pull/1058/commits/e9615a0afc93846fa187050ec5c9ff4e837885e5)

charts/datadog/values.yaml
BaptisteFoy: Fixed in [bce36c7](https://github.com/DataDog/helm-charts/pull/1058/commits/bce36c7cb465befe6dd59927e6e6265fe91e5112)

celenechang (changes requested):
looks good, do you mind running the helm-docs.sh script again?

BaptisteFoy:
`helm-docs` ran in [77a642d](https://github.com/DataDog/helm-charts/pull/1058/commits/77a642d3e3c939f77d8374ed77038a16140a8f79)!

celenechang (approved):

PR #1095, opened 2023-06-26
Support custom FIPS proxy configuration using configMap

fanny-jiang:

Adds support for configuring the FIPS proxy sidecar (`/etc/datadog-fips-proxy/datadog-fips-proxy.cfg`) with a custom configMap. One new parameter added:

* `fips.customFipsConfig`

Example:

```
fips:
  customFipsConfig: |
    foobar
      foo bar baz
```

Default behavior: 
* If `fips.enabled` && `fips.customFipsConfig` => create ConfigMap using data configured with the `fips.customFipsConfig` parameter
* If `fips.enabled` && `fips.customFipsConfig` is _not_ configured => do not create ConfigMap

* Should there default ConfigMap data? If yes, what should the default values be? 
* Currently, the default behavior is to not create the ConfigMap if `fips.customFipsConfig` is empty 
* Changes have been tested manually in EKS

#### Checklist
- [X] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -1183,9 +1183,18 @@ fips:
...
+    # fips.image.repository -- Override default registry + image.name for the FIPS sidecar container.
     repository:
 
+  # fips.useConfigMap -- Configures a configmap to provide the FIPS configuration. Use this in combination with the `fips.customFipsConfig` parameter.
+  useConfigMap:  # false
+
+  # fips.customFipsConfig -- Specify custom contents for the FIPS proxy sidecar container config (/etc/datadog-fips-proxy/datadog-fips-proxy.cfg)
+
+  ## ref: https://docs.datadoghq.com/agent/guide/agent-fips-proxy
+  ## Note the `fips.useConfigMap` needs to be set to `true` for this parameter to be taken into account.
```
sgnn7, line 1195:
```suggestion
  # fips.customFipsConfig -- Specify custom contents for the FIPS proxy sidecar container config (/etc/datadog-fips-proxy/datadog-fips-proxy.cfg)
  ## ref: https://docs.datadoghq.com/agent/guide/agent-fips-proxy
  ## Note the `fips.useConfigMap` needs to be set to `true` for this parameter to be taken into account.
```

charts/datadog/values.yaml
```diff
@@ -1183,9 +1183,18 @@ fips:
     # fips.image.digest -- Define the FIPS sidecar image digest to use, takes precedence over `fips.image.tag` if specified.
     digest: ""
 
-    # agents.image.repository -- Override default registry + image.name for Agent
+    # fips.image.repository -- Override default registry + image.name for the FIPS sidecar container.
     repository:
 
+  # fips.useConfigMap -- Configures a configmap to provide the FIPS configuration. Use this in combination with the `fips.customFipsConfig` parameter.
+  useConfigMap:  # false
```
sgnn7, line 1190: Can we use a single config option for this instead of two separate ones? I would assume that having a config map would imply `useConfigMap`.

sgnn7: Given the fact that the users may easily fall out of compliance using these option, I think we also need this caveat noted explicitly in the comments around them.

charts/datadog/templates/NOTES.txt
```diff
@@ -475,4 +475,14 @@ To send OTLP data to the Agent use the Service created by specifying "http://{{
...
+{{- end }}
+
+{{- if and (eq  (include "should-enable-fips" .) "true") (and .Values.fips.useConfigMap (empty .Values.fips.customFipsConfig) )}}
+
+#################################################################
+####               WARNING: Configuration notice             ####
+#################################################################
+
+You are using the fips.useConfigMap parameter but did not configure fips.customFipsConfig.
+This configuration is unsupported and FIPS proxy side-car custom configMap configuration is disabled.
```
sgnn7, line 487:
```suggestion
This configuration is unsupported and FIPS proxy sidecar custom configMap configuration is disabled.
```

charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -211,6 +211,15 @@ spec:
           - name: config
             mountPath: {{ template "datadog.confPath" . }}
             readOnly: false # Need RW for config path
+{{- if eq (include "should-mount-fips-configmap" .) "true"}}            
+          - name: fips-config
+            mountPath: /etc/datadog-fips-proxy
+            readOnly: false # Need RW to mount to config path
+          - name: fips-proxy-cfg
+            mountPath: /etc/datadog-fips-proxy/datadog-fips-proxy.cfg
```
sgnn7, line 219: Do we need to worry about UIDs to be able to read the mount? [We don't run the fips-proxy container process as root](https://github.com/DataDog/fips-proxy/blob/main/Dockerfiles/amd64/Dockerfile#L110) so hopefully it won't be an issue.

charts/datadog/CHANGELOG.md
```diff
@@ -12,6 +16,7 @@
 
 * Fix comment for datadog.kubernetesEvents.collectedEventTypes in values.yaml.
 
+
```
sgnn7, line 19:
```suggestion
```

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.33.0
+
+* Add `fips.useConfigMap` and `fips.customFipsConfig` parameters to configure FIPS proxy side-car datadog-fips-proxy.cfg using configMap
```
sgnn7, line 5:
```suggestion
* Add `fips.useConfigMap` and `fips.customFipsConfig` parameters to allow configuring FIPS proxy sidecar `datadog-fips-proxy.cfg` using a ConfigMap.
```

sgnn7 (commented):
@fanny-jiang Left some comments. As to the defaults, I think removing the configmap in those cases makes sense so that the config file from the fips-proxy image is used.

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -27,10 +27,12 @@
   name: etc-lsb-release
 {{- end }}
 {{- end -}}
-{{- if eq (include "should-enable-fips" .) "true" }}
+{{- if eq (include "should-enable-fips" .) "true"}}
 - name: fips-config
+  emptyDir: {}
+- name: fips-proxy-cfg
   configMap:
-    name: {{ template "datadog.fullname" . }}-fips-config
```
fanny-jiang, line 33: I don't understand why this `configMap` mount was added [prior](https://github.com/DataDog/helm-charts/commit/d7d311157ffe9aeff6b34780063d29fe8a44d43c#diff-e5b9831252dddaaaed4fcb5f16bc06d3aa6871ec8ade6bb7ceb4ae4d09012aadR30-R34) to this PR since we're only now adding the fips proxy configMap object in this current PR. 

Should I retain this behavior to always include the configMap entry if fips is enabled? I'm not sure if there's any negative side effect of keeping this behavior. Or should I only include it if `fips.customFipsConfig` is configured?

charts/datadog/templates/agent-clusterchecks-deployment.yaml
fanny-jiang: I'm not too sure what you mean. Do you mean whether the fips-proxy container process would be able to read the config file from the mount if the mount is read-only?

charts/datadog/templates/_container-fips-proxy.yaml
```diff
@@ -16,6 +16,16 @@
...
     # TODO Add config to monitor journald
+{{- if eq (include "should-mount-fips-configmap" .) "true"}}  
+  volumeMounts:
+    - name: fips-config
+      mountPath: /etc/datadog-fips-proxy
+      readOnly: false # Need RW to mount to config path
+    - name: fips-proxy-cfg
+      mountPath: /etc/datadog-fips-proxy/datadog-fips-proxy.cfg
+      subPath: datadog-fips-proxy.cfg
+      readOnly: true
```
clamoriniere, line 27: if we need to mount the same volumes in several container, maybe it is is better to create a function helper? 

Also, do we need to consider window host? I think the current change will break windows deployment if `should-mount-fips-configmap == true`

What I suggest it to create file like https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_container-host-release-volumemounts.yaml
do manage the `volumeMounts` for fips configmap.

fanny-jiang: I think we'll be okay in a windows deployment because `"should-mount-fips-configmap"` includes the check for `"should-enable-fips"`, which returns `false` if the targetSystem is not linux: https://github.com/DataDog/helm-charts/pull/1095/files#diff-483dad0ace7ed469efb9d93ba2f3d3c2045bbc1cd317f49e7e0a122bdf9d6116R321-R329

Good idea to create a function helper.

charts/datadog/values.yaml
fanny-jiang: Great idea, updated: https://github.com/DataDog/helm-charts/pull/1095/files#diff-2fcec98016e714405bc936fd949012c2e72f23c20a1a94acf695a9cd3e29d75fR1185-R1192

charts/datadog/templates/agent-clusterchecks-deployment.yaml
sgnn7: The concern is mainly around the fips-proxy being able to read that file since UID/GID of fips-proxy process won't be root. It's more just a question if that will be a problem for the mount or not.

charts/datadog/templates/_daemonset-volumes-linux.yaml
sgnn7: I think we knew that the need to add this functionality was present but it didn't get implemented right then/there. It might have been useful for debugging too 🤷 . As for when we should include it always or not, since I don't know these workflows too well, I think you would have a better perspective on that decision.

sgnn7 (approved):
@fanny-jiang LGTM (as best I could review)!

charts/datadog/templates/agent-clusterchecks-deployment.yaml
fanny-jiang: Got it. I tested the changes manually in EKS and the `fips-proxy` container in the cluster-agent pod log many lines of output indicating that the FIPS mode is enabled and servers are `UP/READY`. I'm not sure if I should include redacted log lines so I'll leave them out.

charts/datadog/templates/_container-agent.yaml
```diff
@@ -242,6 +242,9 @@
     {{- if .Values.datadog.kubelet.hostCAPath }}
 {{ include "datadog.kubelet.volumeMount" . | indent 4 }}
     {{- end }}
+{{- if eq (include "should-mount-fips-configmap" .) "true" }}
+{{- include "linux-container-fips-proxy-cfg-volumemount" . | nindent 4 }}
+{{- end }}
```
clamoriniere, line 247: it should be in the `{{- if eq .Values.targetSystem "linux" }}` condition ([line 161](https://github.com/DataDog/helm-charts/pull/1095/files#diff-0973c8baa0f902471c709f7be2b0682aa0b89d92368289528c1d4b42118980a3L161))

clamoriniere (commented):
hi @fanny-jiang 

a small change needs to be done to be sure we don't break windows deployment.

clamoriniere (approved):
hi @fanny-jiang 

do you think we can add a new CI values.yaml to test this feature in https://github.com/DataDog/helm-charts/tree/main/charts/datadog/ci ?

PR #1102, opened 2023-06-29
Fix conversionWebhook.enabled parameter to set user-configured value

fanny-jiang:

Fix the datadog-operator deployment template to correctly set the user-configured `datadogCRDs.migration.datadogAgents.conversionWebhook.enabled` parameter. 

Previously, the `-webhookEnabled` container arg was not being set on the datadog-operator deployment.yaml when `datadogCRDs.migration.datadogAgents.conversionWebhook.enabled` was set to `true`. It was only set when the config was set to `false`. This bug was obscured by the `datadog-operator` binary which has the default behavior of `webhookEnabled: true`. 

When the default behavior in the datadog-operator is updated to `webhookEnabled: false` in a future version, users using the conversion webhook must upgrade their datadog-operator helm chart to the latest version in order to avoid any interruption in the conversion webhook.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.0.6
 
-* Fix conversionWebhook.enabled parameter to set user-configured value.
+* Fix conversionWebhook.enabled parameter to correctly set user-configured value when enabling the conversion webhook. 
 
 ## 1.0.5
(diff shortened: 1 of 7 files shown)


celenechang (approved):

levan-m (approved):
Thanks for adding tests!

PR #1110, opened 2023-07-06
clusterAgent.admissionController.webhookName to select the name of the mutating webhook and restrict the RBAC policy

AliDatadog:
Currently the Cluster-Agent ServiceAccount get the permission to create,update,delete any mutatingwebhookconfigurations.admissionregistration in the cluster to be able to manage the certificate rotation.

I would like that the cluster-agent only have the strict minimum of RBAC to applyThe principle of least privilege (PoLP) to the Datadog agent install. 

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -189,6 +189,8 @@ spec:
           {{- if .Values.clusterAgent.admissionController.enabled }}
           - name: DD_ADMISSION_CONTROLLER_ENABLED
             value: {{ .Values.clusterAgent.admissionController.enabled | quote }}
+          - name: DD_ADMISSION_CONTROLLER_WEBHOOK_NAME
+            value: {{ .Values.clusterAgent.admissionController.webhookName | quote }}
```
L3n41c, line 193: This parameter doesn’t seem defined in the `values.yaml` file: https://github.com/DataDog/helm-charts/blob/bb9de14bf5bf070f55cec896cd22cfd1e1fac83e/charts/datadog/values.yaml#L951-L978

AliDatadog: I forgot to save the file. It should be fixed now

zhuminyi (approved):

charts/datadog/templates/cluster-agent-deployment.yaml
clamoriniere: do we know it the `cluster-agent` is able to unregister previous webhook if it doesn't use it anymore? (exemple if the webhook is renamed due to this new setting)

PR #1124, opened 2023-07-24
Add process language detection settings in the publich helm chart

AliDatadog:
This PR adds parameters to configure process language detection for APM library injection.

- Since we only use this metadata for APM, this setting is currently under `datadog.apm.*`. However, I think we could have a different prefix if we have other use cases.
- `DD_PROCESS_CONFIG_LANGUAGE_DETECTION_ENABLED` is currently not used in the core agent but it is set only for `agent config`. Should we remove it ?
- `DD_WORKLOADMETA_REMOTE_PROCESS_COLLECTOR_ENABLED` is not used in the process agent.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.33.3](https://img.shields.io/badge/Version-3.33.3-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.34.0](https://img.shields.io/badge/Version-3.34.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 [Datadog](https://www.datadoghq.com/) is a hosted infrastructure monitoring platform. This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally depends on the [kube-state-metrics chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics). For more information about monitoring Kubernetes with Datadog, please refer to the [Datadog documentation website](https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/).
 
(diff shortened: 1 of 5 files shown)


charts/datadog/values.yaml
```diff
@@ -449,6 +449,15 @@ datadog:
     # datadog.apm.hostSocketPath -- Host path to the trace-agent socket
     hostSocketPath: /var/run/datadog/
 
+    process_language_detection:
+      # datadog.apm.process_language_detection.enabled -- Set to true to enable process language detection
+      # which gather information about the language associated with each process. This information
+      # is then used to inject the appropriate Datadog library specific to that language.
+      enabled: false
```
just-chillin, line 456: [question] Do we need both the environment variable and the config setting?

charts/datadog/README.md
```diff
@@ -592,6 +592,8 @@ helm install <RELEASE_NAME> \
 | datadog.apm.hostSocketPath | string | `"/var/run/datadog/"` | Host path to the trace-agent socket |
 | datadog.apm.port | int | `8126` | Override the trace Agent port |
 | datadog.apm.portEnabled | bool | `false` | Enable APM over TCP communication (port 8126 by default) |
+| datadog.apm.process_language_detection.enabled | bool | `false` | Set to true to enable process language detection which gather information about the language associated with each process. This information  is then used to inject the appropriate Datadog library specific to that language. |
```
just-chillin, line 595:
```suggestion
| datadog.apm.process_language_detection.enabled | bool | `false` | Set to true to enable process language detection which gathers information about the language associated with each process. This information is then used to inject the appropriate Datadog library specific to that language. |
```

charts/datadog/templates/_container-agent.yaml
```diff
@@ -100,6 +100,12 @@
     {{- end }}
     - name: DD_APM_ENABLED
       value: "false"
+    - name: DD_WORKLOADMETA_REMOTE_PROCESS_COLLECTOR_ENABLED # Enable the workloadmeta remote process collector only in the core agent
```
just-chillin, line 103: Once rc.5 is released it should be safe to pass this setting everywhere

just-chillin (approved):

PR #1129, opened 2023-07-26
[datadog] trace-agent: now uses cobra so --switch style cli

truthbk:
After merging this PR: https://github.com/DataDog/datadog-agent/pull/16387 in the datadog-agent repository, the CLI behavior for the `trace-agent` changed. Now using `--` two hyphens prepending CLI switches. This PR tries to address that.

  - fixes #

We shouldn't really merge this until the `7.47.x` agent has been released and the chart updated to use that as default value. These changes will affect 7.48.0+ and would break deployments for prior versions of the trace-agent. 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.34.0
+
+* After changes to the trace-agent CLI, arguments now require `--` double-hyphen.
+
 ## 3.33.3
 
 * Remove `datadog.dataStreamsMonitoring.enabled` parameter.
--- a/charts/datadog/templates/_container-trace-agent.yaml
+++ b/charts/datadog/templates/_container-trace-agent.yaml
@@ -3,10 +3,10 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["trace-agent", "-config={{ template "datadog.confPath" . }}/datadog.yaml"]
+  command: ["trace-agent", "--config={{ template "datadog.confPath" . }}/datadog.yaml"]
   {{- end -}}
   {{- if eq .Values.targetSystem "windows" }}
-  command: ["trace-agent", "-foreground", "-config={{ template "datadog.confPath" . }}/datadog.yaml"]
+  command: ["trace-agent", "--foreground", "--config={{ template "datadog.confPath" . }}/datadog.yaml"]
   {{- end -}}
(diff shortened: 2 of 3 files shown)


charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.34.0
```
truthbk, line 3: This is probably not the right version, it doesn't make sense to merge this until we set the default agent version to `7.47.0` which still works with the old single-hyphen arguments.

guyarb: I do suggest to add to the helm chart conditions based on the agent version

truthbk: You're absolutely right, made it backward compatible.

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -2,12 +2,32 @@
 - name: trace-agent
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
+  {{- $version := .Values.agents.image.tag | toString | trimSuffix "-jmx" -}}
+  {{- $length := len (split "." $version) -}}
+  {{- if and (eq $length 1) (eq $version "6") -}}
+  {{- $version = "6.48.0" -}}
```
truthbk, line 8: We could maybe default to `[6,7].46.0` for now (latest stable at the time of this writing), and update this once `7.48.x` is available in a few weeks. That would allow us to merge this immediately.

katiehockman (approved):

charts/datadog/templates/_container-trace-agent.yaml
guyarb: what about custom versions?
Can we have a config flag to override the check here?

guyarb: The concern - debug version or special version we provide to customers, or use internally

but that leads me into a realization that breaking the command line is problematic and can introduce friction with customers
Shouldn't we allow backward compatibility for a couple of version?

charts/datadog/templates/_container-trace-agent.yaml
```diff
@@ -2,12 +2,32 @@
...
+  {{- end -}}
+  {{- if and (eq $length 1) (eq $version "latest") -}}
+  {{- $version = "7.48.0" -}}
+  {{- end -}}
+  {{- if semverCompare "^6.48.0-0 || ^7.48.0-0" $version -}}
+  {{- if eq .Values.targetSystem "linux" }}
+  command: ["trace-agent", "--config={{ template "datadog.confPath" . }}/datadog.yaml"]
+  {{- end -}}
+  {{- if eq .Values.targetSystem "windows" }}
+  command: ["trace-agent", "--foreground", "--config={{ template "datadog.confPath" . }}/datadog.yaml"]
```
clamoriniere, line 21: changing the `trace-agent` container command will break the deployment in GKE autopilot

clamoriniere (commented):
can we implement in the agent a backward CLI compatibility solution to keep the support of `-` argument?

PR #1143, opened 2023-08-08
avoid creating the DD_PROVIDER_KIND env var two times

### LINKED ISSUE #1142: GKE Autopilot: orphaned datadog pods from deamon set
**Describe what happened:**
After a node get scaled up/down the deamon set try to deploy on that node. But in the meantime the nodes gets evicted and the result is an orphaned datadog pod.

The datadog deamonset should be able to remove or terminate the pod gracefully before the node gets evicted

After every node scale up/down on GKE Autopilot

GKE Autopilot Version (1.26.5-gke.1400)

valkum:

Fixes DaemonSets getting stuck in Autopilot when the same env var is defined multiple times.
Currently, `DD_PROVIDER_KIND` is defined twice. This is a spec violation but the validation currently is not enforced during creation. But this prevents the pod GC from removing pods because it seems at that time a different validation is used.
See https://github.com/kubernetes/kubernetes/issues/118261

  - fixes #1142

`provider-env` is currently included in both `containers-cluster-agent-env` and `containers-common-env`. Containers that use `containers-common-env` also use `containers-cluster-agent-env` most of the time. `containers-cluster-agent-env` is only used without `containers-common-env` in `charts/datadog/templates/agent-clusterchecks-deployment.yaml`

#### Checklist
- [x] Chart Version bumped
   _This fails for me locally for the operator which I did not touch._

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Datadog changelog
 
-## 3.33.9 
+## 3.33.10
+
+* Avoid creating the `DD_PROVIDER_KIND` environment variable twice for containers.
+
+## 3.33.9
 
 * Add `fips.customFipsConfig` parameter to allow configuring FIPS proxy sidecar `datadog-fips-proxy.cfg` using a ConfigMap.
 
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.33.9
+version: 3.33.10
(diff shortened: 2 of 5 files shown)


celenechang:
@valkum thanks for the contribution and apologies for the delay. Do you mind fixing the lint errors? Thanks in advance.

valkum:
Ah, sorry. I must have hit save in my editor, and the Markdown format seems to differ from my settings. Amended my commit with the README directly from `.github/helm-docs.sh`

celenechang (approved):

celenechang:
Thanks @valkum!

PR #1146, opened 2023-08-16
update extended daemonset chart for EDS v0.8

adel121:
Updates the extended daemonset chart to use EDS v0.8

#### Motivation
The CRDs of the extended daemonset have been modified from v0.7 to v0.8. As a result, the example demo illustrated in the README.md of the [extendeddaemonset repo](https://github.com/DataDog/extendeddaemonset) fails to create the defined resources because they contain properties not supported in v0.7. Updating this chart to use EDS v0.8 fixes the problem.

#### Changes Made:
- Modified the image tag in values.yaml from 0.7.0 to 0.8.0
- Updated the appVersion and version properties in Chart.yaml
- Updated the CRD definition files by running the `charts/extended-daemon-set/update-crds.sh` script
- Updated the README files by running the `.github/helm-docs.sh`
- Updated CHANGELOG.md
- Modified `update-crds.sh` to check for kubernetes version instead of support for v1/CustomResourceDefinition in case of `v1`

#### Notes for the reviewer:
The change made to `update-crds.sh` was needed because of the following reasons:
- As indicated [here](https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122), the version `apiextensions.k8s.io/v1` is supported on kubernetes versions >= `1.16.0`
- On the other hand, the property `x-kubernetes-map` was introduced in version `1.17.0` of kubernetes (similar issue [here](https://github.com/kedacore/charts/issues/241))

As a result, if we choose between `v1` or `v1beta1` based on support of the `.Capabilities.APIVersions` for `v1`, deployment will fail on version `1.16` because the EDS CRDs use `x-kubernetes-map` which is only supported at or after `1.17.0`. 

This is the reason why `update-crds.sh` was modified to check for the kubeversion with `.Capabilities.KubeVersion.GitVersion` instead.

#### Testing:
To test the changes work as expected:
- Create a local cluster using `kind`
- Install the local helm chart using `helm install -f values.yaml .` from the extended-daemon-set chart director
- Run illustrated in the [extendeddaemonset repo](https://github.com/DataDog/extendeddaemonset) README.md
- Result: resources were created as expected, the canary ERS was initialised and converted to active after 5 minutes.

The `Lint and Test Charts / install-chart (v1.16.15)` job failed because `x-kubernetes-map-type` is supported in v1.17+ as indicated in [this issue](https://github.com/kedacore/charts/issues/241)

#### Checklist
- [ ] Chart Version bumped


L3n41c (approved):
LGTM 👍🏻

celenechang (approved):

PR #1153, opened 2023-08-28
[Orchestrator] Enable orch_explorer by default on agent

JLineaweaver:
We're in the process of moving the orchestrator pod check from the process agent to corechecks. This means we need `orchestrator_explorer.enabled` to default to true in the agent in order to run the check

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -148,6 +148,8 @@
     - name: DD_CHECKS_TAG_CARDINALITY
       value: {{ .Values.datadog.checksCardinality | quote }}
     {{- end }}
+    - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+      value: {{ (include "should-enable-k8s-resource-monitoring" .) | quote }}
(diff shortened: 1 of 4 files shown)


charts/datadog/templates/_container-agent.yaml
```diff
@@ -148,6 +148,8 @@
     - name: DD_CHECKS_TAG_CARDINALITY
       value: {{ .Values.datadog.checksCardinality | quote }}
     {{- end }}
+    - name: DD_ORCHESTRATOR_EXPLORER_ENABLED
+      value: {{ (include "should-enable-k8s-resource-monitoring" .) | quote }}
```
clamoriniere, line 152: the cluster checks should not run in the "Daemonset" agent pods, but instead in the dedicated clusterchecks deployment, so instead, the envvar should be set here: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/agent-clusterchecks-deployment.yaml#L126

kangyili (approved):

celenechang (approved):

PR #1155, opened 2023-08-31
Add rule to allow agent role to curl `/metrics/slis`

jennchenn:
A rule was added to allow the agent to curl `/metrics/slis`, exposing the SLI metrics provided by Kubernetes as a part of the changes made in https://github.com/DataDog/integrations-core/pull/15731

Does the corresponding change need to be made in the operator as well?
 
#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.35.1
+
+* Add permissions to curl `/metrics/slis` to agent cluster role.
+
 ## 3.35.0
 
 * Default `Agent` and `Cluster-Agent` to `7.47.0` version.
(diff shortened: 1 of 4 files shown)


L3n41c (approved):
> Does the corresponding change need to be made in the operator as well?

Indeed, this change will have to be ported to the `datadog-operator`.
Here is where the `datadog-operator` is defining the agent cluster role:
* https://github.com/DataDog/datadog-operator/blob/480f69931540a7de1808ffd7bd9b5043c1249fe7/controllers/datadogagent/component/agent/rbac.go#L32-L37
* https://github.com/DataDog/datadog-operator/blob/480f69931540a7de1808ffd7bd9b5043c1249fe7/pkg/kubernetes/rbac/const.go#L87

But, in order for the operator to be able to grant some permissions, it needs to have them for itself. Those are defined in:
* https://github.com/DataDog/datadog-operator/blob/480f69931540a7de1808ffd7bd9b5043c1249fe7/config/rbac/role.yaml#L9-L12

PR #1159, opened 2023-09-05
Add option to allow the use of APM with k8s service without using hostports

fed-sv:
Add option to allow the use of APM with k8s service without using hostports

To be able to use APM without using hostports

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.36.0
+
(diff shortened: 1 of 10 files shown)


clamoriniere:
I think it is already possible to use the local kubernetes service without opening the host port. 

The chart should be able to detect if the kubernetes version allows local trafic policy. If it is the case it creates the service. And since the trace-agent is enable by default everything should work.

So I'm not sure that we need a new option for the creation of the service.

PR #1163, opened 2023-09-07
[NPM-2758] Update System Probe Chart to include liveness and readiness probes

tblake333:
As a follow up to https://github.com/DataDog/datadog-agent/pull/19297, this PR updates the external charts for `system-probe`.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/NOTES.txt
+++ b/charts/datadog/templates/NOTES.txt
@@ -209,6 +209,30 @@ More information about this change: https://github.com/DataDog/helm-charts/pull/
 {{- fail "You are using datadog.systemProbe.enabled which has been superseded by networkMonitoring.enabled, systemProbe.enableTCPQueueLength, systemProbe.enableOOMKill, and securityAgent.runtime.enabled. These options provide a more granular control of which features should be activated." }}
 {{- end }}
 
+{{- if eq (include "should-enable-system-probe" .) "true" }}
+  {{- $healthPort := .Values.agents.containers.systemProbe.healthPort }}
+  {{- with $liveness := .Values.agents.containers.systemProbe.livenessProbe.httpGet }}
+  {{- if and $liveness.port (ne $healthPort $liveness.port) }}
+
+  ##############################################################################
+  ####            ERROR: System Probe liveness probe misconfiguration       ####
+  ##############################################################################
+
+  System Probe liveness probe port ({{ $liveness.port }}) is different from the configured health port ({{ $healthPort }}).
+  {{- end }}
+  {{- end }}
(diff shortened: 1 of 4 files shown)


charts/datadog/templates/NOTES.txt
```diff
@@ -209,6 +209,30 @@ More information about this change: https://github.com/DataDog/helm-charts/pull/
 {{- fail "You are using datadog.systemProbe.enabled which has been superseded by networkMonitoring.enabled, systemProbe.enableTCPQueueLength, systemProbe.enableOOMKill, and securityAgent.runtime.enabled. These options provide a more granular control of which features should be activated." }}
 {{- end }}
 
+{{- if eq (include "should-enable-system-probe" .) "true" }}
+  {{- $healthPort := .Values.agents.containers.systemProbe.healthPort }}
+  {{- with $liveness := .Values.agents.containers.systemProbe.livenessProbe.httpGet }}
```
hmahmood, line 214: Where is `.Values.agents.containers.systemProbe.livenessProbe.httpGet` specified?

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -21,6 +21,9 @@
     {{- include "containers-common-env" . | nindent 4 }}
     - name: DD_LOG_LEVEL
       value: {{ .Values.agents.containers.systemProbe.logLevel | default .Values.datadog.logLevel | quote }}
+    - name: DD_SYSTEM_PROBE_HEALTH_PORT
```
hmahmood, line 24: Is this always going to be present? There should be a check for the health port value being > 0 to add this env var.

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -151,4 +154,10 @@
 {{- if .Values.agents.volumeMounts }}
 {{ toYaml .Values.agents.volumeMounts | indent 4 }}
 {{- end }}
+  livenessProbe:
```
hmahmood, line 157: Can we gate these behind a config value so that it can be disabled with one config?

charts/datadog/templates/NOTES.txt
tblake333: Interesting. I followed a similar pattern to how the core agent checks for misconfiguration here: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/NOTES.txt#L29-L49, but I can't seem to find where `.Values.agents.containers.agent.livenessProbe.httpGet` is specified either.

Maybe I am missing something, but if not, I'll add the appropriate fields to the `.Values.agents.containers.systemProbe.livenessProbe` object.

PR #1175, opened 2023-09-19
Mount `/etc/passwd` in process agent only if process collection is enabled

levan-m:
Request through support to not mount `/etc/passwd` when only container collection is enabled in the process agent. 

diff chart template for `datadog.processAgent.processCollection` set to `false` and `true`
```
1010a1011,1012
>           - name: DD_PROCESS_AGENT_ENABLED
>             value: "true"
1050a1053,1055
>           - name: passwd
>             mountPath: /etc/passwd
>             readOnly: true
1204c1209
```

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.36.3
+
+* Mount `/etc/passwd` in process agent only if `datadog.processAgent.processCollection` or `datadog.processAgent.processDiscovery` is enabled.
+
 ## 3.36.2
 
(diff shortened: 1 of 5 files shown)


L3n41c (commented):
You may want to update this comment: https://github.com/DataDog/helm-charts/blob/2fc2c758ca8d7947807174ee5db626148dfddf85/charts/datadog/values.yaml#L535
to specify that it’s not systematic anymore.

levan-m:
@L3n41c, thanks for the comment, I overlooked the comment and fixed no. However, not sure if it makes sense together with the rest of comment - `processAgent.enabled` doesn't mention live containers, so why would user enable it without enabling either `processCollection` and `processDiscovery`?

L3n41c (approved):

PR #1177, opened 2023-09-21
Revert enabling remote configuration by default

yshapiro-57:

This temporarily reverts https://github.com/DataDog/helm-charts/pull/1157 .

Why we need this is a long story. There is a crash loop in the cluster agent that a customer on GovCloud reported. I am fixing that crash loop in https://github.com/DataDog/datadog-agent/pull/19596 for the 7.49 build of the agent. I am backporting the fix in https://github.com/DataDog/datadog-agent/pull/19657 to the 7.48 build of the agent. However, right now we are still 1-2 weeks away from releasing 7.48, and we need a workaround to make the cluster agent not crash until then.

The crash only happens when the following conditions are true:

- `admission_controller.auto_instrumentation.patcher.enabled` is true - otherwise we will [never call StartControllers](https://github.com/DataDog/datadog-agent/blob/c0487d93f9b1cf0dd3ca4ef46bacfb513c167edb/cmd/cluster-agent/subcommands/start/command.go#L293) and the crash won't trigger
- `remote_configuration.enabled` [is false or this is a GovCloud customer](https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config.go#L2024) - otherwise [ctx.RcClient won't be nil in StartControllers](https://github.com/DataDog/datadog-agent/blob/c0487d93f9b1cf0dd3ca4ef46bacfb513c167edb/pkg/clusteragent/admission/patch/start.go#L32) and the crash won't happen

https://github.com/DataDog/helm-charts/pull/1157 enables the first flag by default and thus makes the crash inevitable for GovCloud customers. So I would like to revert it until the permanent fix takes effect in 7.48.

Jira: [Fix a crash loop in the cluster agent](https://datadoghq.atlassian.net/browse/AIT-8403)

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.37.1
+
+* Temporarily revert enabling Remote Config by default
+
 ## 3.37.0
 
 * Rename `datadog.securityAgent.compliance.xccdf.enabled` parameter to `datadog.securityAgent.compliance.host_benchmarks.enabled`.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.37.0
+version: 3.37.1
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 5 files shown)


BaptisteFoy (approved):
LGTM for Remote Config

vboulineau (approved):

Issue #1149, opened 2023-08-23
Allow for installation of extra integrations at startup


sblack4:
**Describe what happened:**
We are using a community [integration for redpanda](https://github.com/DataDog/integrations-extras/tree/master/redpanda). The [recommended way](https://docs.datadoghq.com/agent/guide/use-community-integrations/?tab=docker) to set up these integrations in docker is to build your own container. 
We are just installing this one integration to the datadog container and would LOVE to avoid building and supporting our own images. If we could have an init container or startup script install the integration that would be great. 
I don't see any other issues for this but [this one ](https://github.com/DataDog/helm-charts/issues/425)looks like the user also wanted the feature and[ this one](https://github.com/DataDog/helm-charts/issues/286#issuecomment-1632646132) openly asks for it. 

**Describe what you expected:**
To update the `values` file with the list of integrations we want installed and have them installed at startup. 

**Steps to reproduce the issue:**

**Additional environment details (Operating System, Cloud provider, etc):**
Amazon Linux / EKS

bradyburke:
This would be very helpful!

clamoriniere:
Hi @sblack4 and @bradyburke,

Thanks for bringing this up.

Apologies @sblack4—this issue came back to my attention following @bradyburke’s recent comment.

I completely understand the challenge you’re facing when working with an integration that’s not included in the official Datadog Agent image. While using an init container might seem like a convenient workaround, we believe it’s not the most appropriate solution in this case. Let me explain why:

**📉 Drawbacks of Using Init Containers**
- **Slower Pod Startup**
   - Each Agent pod needs to download the integration at launch and on every update.
   - This delays the Agent’s readiness, slows down cluster rollouts, and increases the window of pod unavailability on each node during updates.
- **Reduced Resiliency**
   - Downloading packages during pod startup introduces a network dependency at a critical path.
   - Any transient network or registry issue can prevent the Agent from starting altogether.
- **Redundant operation**
   - The same installation step is repeated on every Agent pod across the cluster, leading to unnecessary overhead.

As you mentioned, our [current recommendation](https://docs.datadoghq.com/agent/guide/use-community-integrations/?tab=containerized#installation) is to build a custom Agent image with the required integrations pre-installed:
```
FROM gcr.io/datadoghq/agent:latest

# Pre-install desired checks
RUN agent integration install -t datadog-<integration_name>
```

This approach offers several benefits:
- Faster, more predictable startup times – no runtime downloads
[...]

PR #294, opened 2021-06-30
Agent additional initcontainers

koletyst:
#### What this PR does / why we need it:
From #286: 
> I need to run an initContainer before datadog containers run.
>
> My use case is to run an authorization request before execution.
>
> Currently there is only a way of editing datadog initContainer resources via agents.containers.initContainers.resources. This map only affects those initContainers that are preset by default by the Chart's template files for different host operating systems.

#### Which issue this PR fixes
Fixes #286 

#### Checklist
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] Chart Version bumped
- [x] `CHANGELOG.md` has beed updated
- [x] Variables are documented in the `README.md`

Issue #286, opened 2021-06-21
datadog chart: Allow adding arbitrary initContainers specs to agent Pods


icelynjennings:
I need to run an initContainer before datadog containers run. 

My use case is to run an authorization request before execution.

Currently there is only a way of editing datadog initContainer resources via `agents.containers.initContainers.resources`. This map only affects those initContainers that are preset by default by the Chart's template files for different host operating systems.

icelynjennings:
> Could you elaborate a bit more what is the "authorization request".

Hits a @hashicorp Vault endpoint generating a new Vault `client_token` which should change across Datadog deployments.

> do you need the capability of adding init-container to only the agent running as daemonset,

Only for `datadog-agent`

icelynjennings:
Happy to create this if you're OK with upstreaming?

clamoriniere:
Thanks for the quick response, 

What is the purpose of this `client_token`? who is using it after?

I'm wondering if our ["secret-backend" feature](https://docs.datadoghq.com/agent/guide/secrets-management/?tab=linux) can be use here to call the hashicorp vault sever when the client_token will need to be used. 

We have customers that use the "secret backend" feature to retrieve sensitive information like credential from `vault`. Do you think it can fit your needs?

icelynjennings:
Thanks @clamoriniere. I followed your linked instruction and mounted a compiled secrets-backend binary into the datadog agent container (I modified the documentation's Go example to fit my needs). It may become a better solution to change datadog agent's container image to include Go, so it can compile the binary on startup, or use a Go image initContainer to do the same.

I found the process of distributing the binary into agent containers significantly more involved than authenticating from a discardable initContainer however, particularly because I'm unsure how well my team will be able to maintain this code, compared to a few lines of shell script in an initContainer.

icelynjennings:
> What is the purpose of this client_token? who is using it after?

This token is used by [Datadog's Vault integration](https://docs.datadoghq.com/integrations/vault/?tab=host#configuration)

icelynjennings:
Would @datadog still prefer to hide initContainers from the official chart? I have opened a fork to implement this if you'd like to push it through

winnieoursbrun:
With that, do you think we can install community integration more easily than create a new docker image ?

AnthonyHerman:
FYI in case this helps anyone, you can get the vault integration working with the authenticated metrics endpoint using the agents.podAnnotations with vault injector. This is what my setup looks like ->

[...]

PR #1183, opened 2023-09-27
add support for initContainer volume mounts

imdevin567:

This adds support for volume mounts inside the initContainers. For my use case, this allows custom shell scripts to be placed in `/etc/cont-init.d` to run when the container starts up.

While there isn't an issue _specifically_ for this, there are a number of issues that would be solved with this change:

- #286 requests support for additional initContainers. This is an option, but with a larger blast radius to consider.
- #294 was submitted over 2 years ago to fix the issue above, but was never discussed or merged.
- #1107 added direct support for community integrations, but it was closed due to operational risk. This change uses an existing mechanism built into the image rather than change the startup command.
- #1149 requests community integration installation support. The datadog-agent image already supports custom scripts via the `/etc/cont-init.d` directory; this change would allow for a custom installation script to be mounted in the initContainer.

#### Checklist
- [X] Chart Version bumped
- [X] For Datadog Operator chart or value changes update the test baselines (run: `make update-test-baselines`)

### DIFF
--- a/charts/datadog/templates/_containers-init-linux.yaml
+++ b/charts/datadog/templates/_containers-init-linux.yaml
@@ -53,6 +53,9 @@
       subPath: system-probe.yaml
       readOnly: true
     {{- end }}
+    {{- if .Values.agents.containers.initContainers.volumeMounts }}
+    {{ toYaml .Values.agents.containers.initContainers.volumeMounts | nindent 4 }}
+    {{- end }}
   env:
     {{- include "containers-common-env" . | nindent 4 }}
     {{- if and (eq (include "cluster-agent-enabled" .) "false") .Values.datadog.leaderElection }}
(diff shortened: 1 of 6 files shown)


charts/datadog/templates/_containers-init-linux.yaml
```diff
@@ -53,6 +53,9 @@
       subPath: system-probe.yaml
       readOnly: true
     {{- end }}
+{{- if .Values.agents.containers.initContainers.volumeMounts }}
+{{ toYaml .Values.agents.containers.initContainers.volumeMounts | indent 6 }}
```
levan-m, line 57:
```suggestion
{{ toYaml .Values.agents.containers.initContainers.volumeMounts | indent 4 }}
```
Could you please double-check indentation? I was unable to render the template with this indentation. For improved readability consider putting whole block at indent 4 (and using `nindent 4`).

levan-m (approved):

levan-m:
Thank you for the contribution!

PR #1199, opened 2023-10-07
ci: switch from kubeval to kubeconform

### LINKED ISSUE #360: Consider using alternative source for kubeval schemas
[instrumenta/kubernetes-json-schema](https://github.com/instrumenta/kubernetes-json-schema) is noticeably out of date at this point, with the last commit to master landing in April of 2020, this puts the latest provided Kubernetes version at 1.18.1.

As an alternative there is [yannh/kubernetes-json-schema](https://github.com/yannh/kubernetes-json-schema) which is self-updating and currently up-to-date.

It's a trivial change to update the schema source (see: [here](https://github.com/timberio/helm-charts/blob/master/.github/kubeval.sh#L6)).

This would allow for the Helm charts to be evaluated against more, and more recent versions of Kubernetes.

### LINKED ISSUE #1312: Duplicate `nodeSelector` values in datadog chart with CI defaults
**Describe what happened:**
While working on #1199, I noticed that if `kubeconform` is run with `-strict`, errors like this are observed
```
Deployment failed validation: release-name-datadog-cluster-agent error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 242: key "kubernetes.io/os" already set in map
```

Essentially, I think when `nodeSelector` is specified in a values file, it will be included in addition to the baked in defaults, rather than instead of it. I'm not sure if that's the intended behavior or not.
If it is, maybe the simple fix is to use a different example from the os default for `nodeSelector` in the ci values file, but if one is supposed to replace the other, the fix will be more complicated.

No error to be reported

1) Run:
```
% helm dep up charts/datadog && helm template --kube-version "1.22.7" --values charts/datadog/ci/kubeconform-values.yaml charts/datadog
```
2) Run kubeconform on the resulting output with the strict flag.

Note: I am guessing this is because of stuff like 

https://github.com/DataDog/helm-charts/blob/0e6401a01c732fa303f212f208f3a9289da26ffb/charts/datadog/templates/cluster-agent-deployment.yaml#L483-L487

wyardley:
- Switch from kubeval, which is unmaintained and outdated, to kubeconform
- Update kubernetes versions in test matrices
- Use upstream schema for CRDs, so we don't have to ignore missing schemas in validation
- Switch away from deprecated output format
  https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
- Switch CI nodeSelector examples to use disktype: ssd to avoid creating duplicates with baked in os defaults

Screenshot from local example:
image

  - fixes #360
  - fixes #1312

Note: I don't know if any chart changes are triggered. I tested this locally, but may be good to do some additional testing in CI if you'd like.

CI changes only; let me know if you want me to bump version / changelog, though I imagine not relevant here.

I'm happy to remove `v1.18.20` or add / remove other versions from the matrices - just let me know.

#### Checklist
- [x] Chart Version bumped


wyardley:
Note: I just updated it to resolve a deprecation warning with `set-output` - once the job is approved, I will double-check that it looks correct

And, as mentioned before, I can bump the chart version if / when you all want, which is one of the test failures.

I don't think the go test failure is related to my changes?

wyardley:
There is a separate (and similar) matrix of Kube versions for the go integration tests, which I'm happy to edit there / make standard with he ones in this PR if it helps (though I haven't yet)

wyardley:
@clamoriniere would you be able to get someone to review / comment on this, and / or is there anything you'd like me to do / update?

.github/kubeconform.sh
```diff
@@ -0,0 +1,35 @@
...
+  # Note: -ignore-missing-schemas could be added if needed, but not currently
+  # needed since we have the schema necessary to validate the CRDs themselves.
+  #
+  # Also, if at some point we needed to validate things _using_ these CRDs,
+  # they're available via
+  # https://github.com/datreeio/CRDs-catalog/tree/main/datadoghq.com
+  helm dep up "${CHART_DIR}" && helm template --kube-version "${KUBERNETES_VERSION#v}" \
+        --values "${CHART_DIR}"/ci/kubeconform-values.yaml "${CHART_DIR}" \
+    | ./kubeconform -schema-location default -schema-location "$CRD_SPEC_URL" \
+        -output pretty -verbose -kubernetes-version "${KUBERNETES_VERSION#v}" -
```
loutPhilipps, line 34:
```suggestion
        -output pretty -strict -verbose -kubernetes-version "${KUBERNETES_VERSION#v}" -
```
`-strict` was enabled with `kubeval`, any reason not to enable it with `kubeconform`?

.github/workflows/ci.yaml
```diff
@@ -80,12 +80,11 @@ jobs:
...
-          # - v1.14.10 (deprecated with kind 1.16)
-          - v1.16.15
           - v1.18.20
-          - v1.22.9
-          - v1.24.2
-          # v1.25.2 (when agent 7.40 is out)
+          - v1.22.17
+          - v1.24.17
+          - v1.25.15
+          - v1.26.10
```
loutPhilipps, line 87: For this I am not sure what the implications are for the agent - I'll let another reviewer comment for these

loutPhilipps (commented):
It looks like there's an issue with the output format:
```
Error: Unable to process file command 'output' successfully.
Error: Invalid format 'charts/datadog-operator'
```

.github/kubeconform.sh
wyardley: it has been a while since I put this together / tested it, but it looks like this is probably why:
```
✖ stdin: DaemonSet failed validation: release-name-datadog error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 879: key "kubernetes.io/os" already set in map
✖ stdin: Deployment failed validation: release-name-datadog-clusterchecks error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 181: key "kubernetes.io/os" already set in map
✖ stdin: Deployment failed validation: release-name-datadog-cluster-agent error unmarshalling resource: error converting YAML to JSON: yaml: unmarshal errors:
  line 242: key "kubernetes.io/os" already set in map
```

and in fact, this happens a bunch in the configs:
```yaml
      nodeSelector:
        kubernetes.io/os: linux
        kubernetes.io/os: linux
```

My suggestion is that we don't try to fix that in this PR, but I can rebase if someone wants to throw up a fix for it?

.github/workflows/ci.yaml
wyardley: This is also pretty old / crusty, so may need to be updated again.

.github/kubeconform.sh
loutPhilipps: I'm fine with that, up to the maintainers of these charts 👍

wyardley:
> It looks like there's an issue with the output format:

Ah! probably shouldn't have tried to fix this error at the same time.
https://stackoverflow.com/questions/74137120/how-to-fix-or-avoid-error-unable-to-process-file-command-output-successfully

I think it should have been `changed` anyway vs `{changed}`; not sure if just changing `{changed}=` to `changed=` will be sufficient, but I can just add in the solution using a heredoc and we can see if that works...

.github/workflows/ci.yaml
```diff
@@ -33,7 +33,9 @@ jobs:
           if [[ -n "$changed" ]]; then
             echo -n "Charts changed:"
             echo "$changed"
-            echo "::set-output name=changed::$changed"
+            echo "changed<<EOF" >> $GITHUB_OUTPUT
+            echo "$changed" >> $GITHUB_OUTPUT
+            echo "EOF" >> $GITHUB_OUTPUT
```
wyardley, line 38: I think this will result in
```
changed
[lines changed]
```
vs `changed=[first line]`. Not 100% sure if that will work or not

.github/kubeconform.sh
wyardley: Are you able to bug someone about it? I can try to take a quick look and see if there's an obvious fix, but it would nice to be able to enable strict here vs. waiting, since it's kind of buried here.

I checked, and it's only `charts/datadog` that has the issue - all the others succeed with `-strict`. If it's a really easy fix, I can try to open up a PR for that

wyardley:
```
      nodeSelector:
        {{ template "label.os" . }}: {{ .Values.targetSystem }}
      {{- if .Values.agents.nodeSelector }}
{{ toYaml .Values.agents.nodeSelector | indent 8 }}
      {{- end }}
```

Maybe because of this (and similar code in other sections)?

.github/workflows/ci.yaml
wyardley: I'm happy to plugin whatever values you all would like here.

I'm not currently using Datadog, but I'm happy to do some small changes if it will help get this shipped.

.github/kubeconform.sh
wyardley: I created #1312 

One possibility is that the chart _intends_ to allow multiple `nodeSelector` statements, and that this isn't intended to override the baked in defaults? It looks like a feature was released in 2.x to always includle the target system nodeSelector; in that case, there's no real bug, and the "fix" would probably be to just use some kind of arbitrary `nodeSelector` value that doesn't match in this PR. I'm happy to make that change if we can confirm this is the intended behavior.

wyardley:
@loutPhilipps if you want to trigger again, we can see if my output fixes now work.

.github/kubeconform.sh
loutPhilipps: Indeed it worked

loutPhilipps:
> Note: I just updated it to resolve a deprecation warning with `set-output` - once the job is approved, I will double-check that it looks correct
> 
> And, as mentioned before, I can bump the chart version if / when you all want, which is one of the test failures.
> 
> I don't think the go test failure is related to my changes?

I believe you do need to bump the chart versions

wyardley:
> I believe you do need to bump the chart versions

I can if you want, but isn't the bump of version just in cases where the chart itself changes? this should (IMO) be a non-triggering change? The actual kubeconform jobs do already run successfully here, so other than the failing CI check, is there a reason the version should be bumped, when the charts themselves are not changing?

The only issue I have with updating the chart versions is that it may make it harder to keep this PR up to date -- if it's ready to ship and they really need to be updated, I can go ahead and do it.

.github/workflows/ci.yaml
wyardley: https://github.com/yannh/kubernetes-json-schema
these are the current schema versions available.

I can add 1.27 / 1.28 if desired, and / or update to `v1.25.16` and `v1.26.13`.

Note that `.github/workflows/go-test.yaml` also has a matrix of Kube versions. If someone wants to provide a list of what versions are wanted for either / both, I can update them, but I'll probably default to not touching more things than necessary.

clamoriniere: If we can we would like to keep Kubernetes 1.16 in our tests. 
Unfortunately we still have some user with this version, but this version is the pre-version of several resources deprecation in the API server.

.github/workflows/ci.yaml
clamoriniere: should be ok

.github/workflows/ci.yaml
clamoriniere: If `kubeconform` is not available for 1.16, maybe we can keep kubeval for 1.16, and use `kubeconform` for the more ressent versions.  WDYT

charts/datadog/ci/kubeconform-values.yaml
```diff
@@ -36,7 +36,7 @@ clusterAgent:
   enabled: true
   createPodDisruptionBudget: true
   nodeSelector:
-    kubernetes.io/os: linux
+    disktype: ssd
```
clamoriniere, line 39: does this change is needed?

wyardley: See comment above and #1312 - short answer is yes if we want to have `-strict` and the intent is to allow multiple (different) values to be passed in to `nodeSelector`.

wyardley:
#360 is technically already fixed, but I'm adding it in here as it will be extra-fixed once this is included.

charts/datadog/ci/kubeconform-values.yaml
clamoriniere: thanks for the explanation.

clamoriniere:
Hi @wyardley 

thanks you for this contribution and your reactivity on comments. 
I made this PR on your branch https://github.com/wyardley/helm-charts/pull/1/ to add the missing change to make ou CI happy. If you can merge it. 

Then I think we will be able to merge the PR.

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.53.2
+version: 3.53.3
```
clamoriniere, line 3:
```suggestion
version: 3.53.4
```

charts/datadog/README.md
```diff
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.53.2](https://img.shields.io/badge/Version-3.53.2-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.53.3](https://img.shields.io/badge/Version-3.53.3-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
```
clamoriniere, line 3:
```suggestion
![Version: 3.53.4](https://img.shields.io/badge/Version-3.53.4-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
```

.github/workflows/ci.yaml
```diff
@@ -80,12 +82,12 @@ jobs:
         # https://github.com/instrumenta/kubernetes-json-schema is no
         # longer updated
         k8s:
-          # - v1.14.10 (deprecated with kind 1.16)
-          - v1.16.15
+          - v1.16.4
```
wyardley, line 85: note: https://github.com/instrumenta/kubernetes-json-schema doesn't seem to have 1.16.15; 1.16.4 is the latest. Presumably this is Good Enough, but if someone can find the right schema version (I'm surprised `kubeval` worked with this) online in the right format, I can update the secondary schema location 🤷

wyardley:
I rebased locally and slightly tweaked the wording, since there were conflicts from something merged after the previous version bump. Sorry for squashing your changes @clamoriniere

wyardley:
Seems like it wanted additional bumps?
I bumped `charts/datadog-crds` and `charts/datadog-operator` too, in a separate commit, but let me know if you want me to revert it to how it was.

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.53.3
+version: 3.53.4
```
clamoriniere, line 3:
```suggestion
version: 3.54.1
```

charts/datadog/README.md
```diff
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.53.3](https://img.shields.io/badge/Version-3.53.3-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.53.4](https://img.shields.io/badge/Version-3.53.4-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
```
clamoriniere, line 3:
```suggestion
![Version: 3.54.1](https://img.shields.io/badge/Version-3.54.1-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
```

charts/datadog/CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
 # Datadog changelog
 
+## 3.53.4
```
clamoriniere, line 3:
```suggestion
## 3.54.1
```

.github/workflows/ci.yaml
```diff
@@ -95,26 +97,26 @@ jobs:
...
+      - kubeconform-chart
     strategy:
       matrix:
         k8s:
-          - v1.16.15
           - v1.18.20
           - v1.22.9
           - v1.24.2
-          - v1.25.2
+          - v1.25.14
```
clamoriniere, line 118: if seems that the kind node image 1.25.14 is not available 

but v1.25.16 is

.github/workflows/ci.yaml
```diff
@@ -95,26 +97,26 @@ jobs:
...
     strategy:
       matrix:
         k8s:
-          - v1.16.15
           - v1.18.20
           - v1.22.9
           - v1.24.2
-          - v1.25.2
+          - v1.25.16
+          - v1.26.9
```
clamoriniere, line 119: this version is not available but v1.26.13 it

```suggestion
          - v1.26.13
```

charts/datadog/ci/kubeconform-values.yaml
clamoriniere, line 39: I think the [CI is failing](https://github.com/DataDog/helm-charts/actions/runs/7872214797/job/21479319052?pr=1199#step:6:43) because we use this file in the CI Kind cluster. but the node doesn't have this nodeSelector. 

my suggestion for now is to rename this file: `charts/datadog/ci/kubeconform.yaml` so it will not be selected for the CI installation.

wyardley: I did that, but failed because I hadn't renamed it for each directory. But I do think there's some value to making sure to test / lint the values files, so since it looks like you've got a different / better fix in #1316 anyway, I reverted the change.

If you don't want it to be setting the disk type, I think we could probably set something arbitrary or harmless instead?

clamoriniere: Thanks @wyardley 

I'm still on it with another PR https://github.com/DataDog/helm-charts/pull/1317. Hopefully it will be the last one :)

clamoriniere (approved):
Hi @wyardley 
thanks again for all your work on this PR. Our CI is finally green ! 🥳 
We will be able to merge it

wyardley:
Thanks all!

PR #1210, opened 2023-10-19
Added option to disable external metrics registration as an APIService

adel121:

External metrics server is started by the datadog agent when `clusterAgent.metricsProvider.enabled` is set to `true`. It is also registered by the helm chart in the kube api server as an APIService and an rbac is created to support it. 

This PR allows the user to enable creating an external metrics server while still disabling registering the external metrics server as an APIService. 

It does so by adding the option `clusterAgent.metricsProvider.registerAPIService` to allow user to disable registering external-metrics server as an `APIService`

This allows the user to use other external-metrics providers (such as KEDA) in the cluster, but also use the datadog cluster-agent as a metrics proxy to datadog. 

### Notes for reviewer:
- `clusterAgent.metricsProvider.registerAPIService` is defaulted to `true` in `values.yaml` for backward compatibility. By default we register the endpoint if someone enable external metrics server.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -925,6 +925,9 @@ clusterAgent:
   metricsProvider:
     # clusterAgent.metricsProvider.enabled -- Set this to true to enable Metrics Provider
     enabled: false
+    
+    # clusterAgent.metricsProvider.register -- Set this to false to disable external metrics registration as an APIService
+    register: true
 
     # clusterAgent.metricsProvider.wpaController -- Enable informer and controller of the watermark pod autoscaler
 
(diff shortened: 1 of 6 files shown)


vboulineau (approved):

clamoriniere (approved):
💯

charts/datadog/values.yaml
```diff
@@ -926,6 +926,9 @@ clusterAgent:
     # clusterAgent.metricsProvider.enabled -- Set this to true to enable Metrics Provider
     enabled: false
 
+    # clusterAgent.metricsProvider.register -- Set this to false to disable external metrics registration as an APIService
```
wdhif, line 929:
```suggestion
    # clusterAgent.metricsProvider.registerAPIService -- Set this to false to disable external metrics registration as an APIService
```

wdhif (approved):
LGTM 👍

PR #1228, opened 2023-11-08
Enable container image collection

L3n41c:

Enable “container image” collection by default.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -632,7 +632,7 @@ datadog:
 
   containerImageCollection:
     # datadog.containerImageCollection.enabled -- Enable collection of container image metadata
-    enabled: false
+    enabled: true
 
   orchestratorExplorer:
     # datadog.orchestratorExplorer.enabled -- Set this to false to disable the orchestrator explorer
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.45.0
+
+* Enable container image collection by default.
+
 ## 3.44.0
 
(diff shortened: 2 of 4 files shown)


davidor (approved):

charts/datadog/values.yaml
```diff
@@ -632,7 +632,7 @@ datadog:
 
   containerImageCollection:
     # datadog.containerImageCollection.enabled -- Enable collection of container image metadata
-    enabled: false
+    enabled: true
```
celenechang, line 635: I didn't think of this earlier; do you think it's worth adding a minimum agent version in the description?

L3n41c: This sounds like something we should have done for all the other options.
For this parameter to work, we need DataDog/datadog-agent#16847. Hence `7.46+`.
Container images collection is a feature which is much older than that. But the way to enable it was different.
So, 7.46+ is the minimum version so that it is enabled by this version of the chart.

Here is the change in the description: https://github.com/DataDog/helm-charts/pull/1228/commits/b0b335b6979cb438d01e9c34911990bba5ad53bd

charts/datadog/values.yaml
```diff
@@ -632,7 +632,9 @@ datadog:
 
   containerImageCollection:
     # datadog.containerImageCollection.enabled -- Enable collection of container image metadata
-    enabled: false
+
+    # This parameter will be honored only by Agent version 7.46+
```
celenechang, line 636:
```suggestion
    # This parameter requires Agent version 7.46+
```

charts/datadog/values.yaml
celenechang: thanks

celenechang (approved):

badovm (approved):

PR #1246, opened 2023-11-16
Update agent-clusterchecks-rbac.yaml

kzap:
- Cluster Check pods cannot access k8s secrets for Secret Backend decryption

  - fixes #

#### Checklist
- [ ] Chart Version bumped


charts/datadog/templates/agent-clusterchecks-rbac.yaml
```diff
@@ -13,6 +13,25 @@ subjects:
...
+subjects:
+  - kind: ServiceAccount
+    name: {{ template "datadog.fullname" $ }}-cluster-checks
+    apiGroup: ""
+    namespace: {{ $.Release.Namespace }}
+roleRef:
+  kind: Role
+  name: {{ template "datadog.fullname" $ }}-secret-reader-{{ $role.namespace }}
+  apiGroup: ""
+{{- end }} # end range $role := .Values.datadog.secretBackend.roles
```
kzap, line 34: without this the cluster checks wont have access to k8s secrets if `.Values.datadog.secretBackend.roles` is used. The cluster check would only have access if the ClusterRole had access to all k8s secrets which some cluster operators may not want to allow.

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -243,6 +243,10 @@ spec:
             value: "kube_endpoints kube_services"
           - name: DD_EXTRA_LISTENERS
             value: "kube_endpoints kube_services"
+          {{- if .Values.datadog.secretBackend.command -}}
+          - name: DD_SECRET_BACKEND_SKIP_CHECKS
+            value: "true"
+          {{- end }}
```
kzap, line 249: I cluster checks was not decrypting the secrets until I enabled this.

I am not sure why and could not find any documentation except this PR, I think there is missing Documentation about this setting
- https://github.com/DataDog/datadog-agent/pull/8492

charts/datadog/templates/agent-clusterchecks-rbac.yaml
CharlyF: I think this is likely because it's copied from the [other file](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/rbac.yaml#L191-L195), but could you remove the comment and add the apiGroup `rbac.authorization.k8s.io` ?

charts/datadog/templates/cluster-agent-deployment.yaml
CharlyF: that's a good catch, we don't have the helm option to use this.
I don't think it should be flipped on from the usage of a command though.
I would recommend adding a new parameter `Values.datadog.secretBackend.skipChecks`.
I can do this as a follow up PR, and for now maybe use:
```
clusterAgent
  env:
      - name: DD_SECRET_BACKEND_SKIP_CHECKS
        value: true
```

kzap:
Thanks, will update as requested 👍

charts/datadog/templates/agent-clusterchecks-rbac.yaml
kzap: ok applied that change too to the other file

CharlyF (approved):

kzap (commented):
re-signed all commits

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -259,6 +259,8 @@ spec:
             value: "kube_endpoints kube_services"
           - name: DD_EXTRA_LISTENERS
             value: "kube_endpoints kube_services"
+          - name: DD_SECRET_BACKEND_SKIP_CHECKS
+            value: "true"
```
kzap, line 263: I do wonder if you are not using the Cluster Checks agent, and want your secrets decrypted on the Cluster Agent, will this affect that?

CharlyF: I was about to leave a note. Can you remove this part entirely? 
It will break checks containing ENC that are supposed to be run by the Cluster Agent.
I will make a subsequent PR to add this with a flag and document it.
For now you can use:
```
clusterAgent
  env:
      - name: DD_SECRET_BACKEND_SKIP_CHECKS
        value: true
```

kzap: removed

CharlyF (approved):

CharlyF:
Thank you very much @kzap - This PR will merge on Monday.

CharlyF:
@kzap - circling back on this, it seems like the automerge is blocked because all commits need to be signed. I updated the branch to resolve the conflicts from the rebase but would you mind maybe squashing into one commit and signing it?

kzap:
> @kzap - circling back on this, it seems like the automerge is blocked because all commits need to be signed. I updated the branch to resolve the conflicts from the rebase but would you mind maybe squashing into one commit and signing it?

ok just saw this, will do

kzap (commented):
squashed into 1 single commit as requested @CharlyF

PR #1250, opened 2023-11-20
Add command and args options for Agent and ClusterAgent containers

Kummerkasten:
This PR adds the possibility to configure the command and args parameter for clusterAgent and Agent containers in order to overwrite entrypoints or to add custom arguments.
We do use the vault-agent-injector company wide to inject secrets into containers. For the observabillity-pipelines-worker we do this currently and this requires us to run a custom command and args for the observabillity-pipelines-worker container which is possible to set via the  observabillity-pipelines-worker values file.
Like in this example: https://developer.hashicorp.com/vault/docs/platform/k8s/injector/examples#environment-variable-example

We want to migrate our remaining deployments to use the datadog helm charts as well and also keep using the vault-agent-injector to keep configuration unified.
 
  - fixes #

#### Checklist
- [x] Chart Version bumped


clamoriniere:
Hi @Kummerkasten 

We would like to keep control of the `command` and `args` container parameter in the Agent and Cluster-Agent pods for several reasons:
* Allows us to manage configuration evolution without conflicting with users configuration: if for example we would like to add or remove a parameter.
* On some kubernetes environments like `GKE autopilot` the `command` and `args` must not be changed, else the Daemonset will be refused.

However, for managing and retrieving secrets, the Datadog Agent/Cluster-Agent provide a solution to retrieve secrets call "secret backend": https://docs.datadoghq.com/agent/configuration/secrets-management/?tab=linux 

It should be possible to ask the vault client to store the secret in a file, and configure the datadog secret backend to read from this file.

Please let us know if this solution can work for you?

Kummerkasten:
Hi @clamoriniere,
thanks for that quick reply.  Please correct me if I am wrong. After reading through the documentation link you provided regarding "secret backend" as far as I understand we would need to build custom code to communicate and retrieve secrets from vault and build a custom datadog-agent docker image which has the binary/executable/script built in. 

Under this assumption this would be out of scope for us now. Like I mentioned we use the vault-agent-injector in all our services which basically provides this functionality already and we could combine it with the usage of the helper-script from "secret backend" to read mounted secrets via file to avoid loading secrets into the container environment.

I do get the point that you have the need to have control over the `command` and `args` container parameters. I personally am well aware that there might be changes to the default settings you have in the templates and this might change in the future. So every update of future Charts on our side needs to be reviewed properly.
To float some ideas around which I had in mind to make it more clear to other users would be:

- Having comments in the values file which warn users that changing the command and args might break things and this should only be customized when absolutely necessary 
- Move the default `command` and `args` parameters to the values file as well so it is easier to see what the defaults are compared to inside the templates.The downside here is that templated paths to config files would not work like for example in the process-agent container

What do you think?

clamoriniere:
Hi @Kummerkasten 

The Datadog agent and cluster-agent image already contains a secret-backend implementation that is able to read from a local file secret (doc: https://docs.datadoghq.com/agent/configuration/secrets-management/?tab=linux#script-for-reading-from-multiple-secret-providers) 
So you should still be able to use the `vault-agent-injector` to create the local file (as it currently done to create the file that later it is "sourced" to inject the the secret as envvar (ref: https://developer.hashicorp.com/vault/docs/platform/k8s/injector/examples#environment-variable-example).

so what I suggested was to configure the `vault-agent-injector` to write only the secret in the generate file with a configuration like:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: web-deployment
  labels:
    app: web
spec:
  replicas: 1
  selector:
    matchLabels:
      app: web
  template:
    metadata:
      labels:
        app: web
      annotations:
        vault.hashicorp.com/agent-inject: 'true'
        vault.hashicorp.com/role: 'web'
        vault.hashicorp.com/agent-inject-secret-config: 'secret/data/datadog-api-key'
        # Environment variable export template
        vault.hashicorp.com/agent-inject-template-config: |
          {{- with secret "secret/data/web" -}}
          {{ .Data.data.payments_api_key }}
          {{- end }}
    spec:
      serviceAccountName: web
      containers: #...

```

if we translate it in `datadog` chart configuration, we could have something like this `values.yaml` file

```yaml
datadog:
  apiKey: "ENC[file@/secrets/datadog-api-key]". # provide the path were the secret will be available
  # [...]
  secretBackend:
    command: "/readsecret_multiple_providers.sh". # configure the secret-backend default implementation
# [...]
agents:
  daemonsetAnnotations:  # configure vault agent injector
    vault.hashicorp.com/agent-inject-secret-apikey-config: 'secret/data/datadog-api-key'
    vault.hashicorp.com/agent-inject-file-apikey-config: "/datadog-api-key"
    vault.hashicorp.com/secret-volume-path-apikey-config: "/secrets"
    # export secret in a file that only contains the secret
    vault.hashicorp.com/agent-inject-template-apikey-config: |
      {{- with secret "secret/data/datadog-api-key" -}}
      {{ .Data.data.api_key }}
      {{- end }}
```

I'm not sure about the `vault-agent-injector`options configuration, but is should illustrate a bit more my initial idea. Does my idea make sense?

Kummerkasten:
Hey @clamoriniere,

I just came back from a walk and I had that sudden enlightenment that I gave myself the answer in my previous comment which is exactly what you had in mind here as well :D Thank you for providing an example! I will test it and if it works as intended which I guess it will I am going to close this PR

PR #1263, opened 2023-12-05
[AIT-8952] Generate DD_INSTRUMENTATION_INSTALL_TIME and DD_INSTRUMENTATION_INSTALL_ID

liliyadd:
Generates DD_INSTRUMENTATION_INSTALL_TIME and DD_INSTRUMENTATION_INSTALL_ID on installation of Cluster Agent. These env variables are used to power [APM time-to-value KPI](https://docs.google.com/document/d/14vsrCbnAKnXmJAkacX9I6jKPGKmxsq0PKUb3dfiZpWE/edit#heading=h.8d3o7vtyu1y1)

https://datadoghq.atlassian.net/browse/AIT-8952

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 3.50.0
 
+Add env variables to support APM Telemetry KPIs to Tracer Agent
+
+## 3.50.0
+
 Add env variables to support APM Telemetry KPIs
 
 ## 3.49.7
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.51.0
 
-* Add env variables to support APM Telemetry KPIs
+* Add `DD_INSTRUMENTATION_INSTALL_TIME`, `DD_INSTRUMENTATION_INSTALL_ID`, `DD_INSTRUMENTATION_INSTALL_TYPE` env variables to the Trace and Cluster agents to support APM Telemetry KPIs.
 
 ## 3.50.5
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.51.0
 
-* Add env variables to support APM Telemetry KPIs to Tracer Agent
+* Add env variables to support APM Telemetry KPIs
 
 # 3.50.2
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.50.0
+
+Add env variables to support APM Telemetry KPIs
+
 ## 3.49.5
 
 Fix registry selection with GKE Autopilot until new registries are allowed.
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Datadog changelog
 
-## 3.50.0
+## 3.50.1
 
 Add env variables to support APM Telemetry KPIs to Tracer Agent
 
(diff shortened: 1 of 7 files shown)


charts/datadog/templates/kpi-telemetry-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kpi-telemetry-configmap
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "datadog.labels" . | indent 4 }}
+data:
+  install_id: {{ uuidv4 | quote }}
+  install_type: unknown
```
CharlyF, line 10: is this desired?

charts/datadog/templates/kpi-telemetry-configmap.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: kpi-telemetry-configmap
```
clamoriniere, line 4: I'm not sure it is good to have a fix name here. because we have some use case (like windows install that require to deploy/install twice this chart in the same namespace)

Also can we create this configmap only if APM SSI is enabled?

liliyadd: - Parametrized the Configmap name with `.Release.name`
- We need this env variables to get KPIs for all of manual instrumentation, local library injection and SSI. Thus we cannot condition configmap on SSI enablement.

charts/datadog/templates/_containers-common-env.yaml
```diff
@@ -151,3 +151,13 @@ Return a list of env-vars if the cluster-agent is enabled
         key: token
 {{- end }}
 {{- end -}}
+
+
+{{- define "kpi-envvar" -}}
```
levan-m, line 156: Isn't this redundant since env vars are inlined in both trace agent and DCA containers?

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.51.0
+
+* Add env variables to support APM Telemetry KPIs
```
levan-m, line 5:
```suggestion
* Add `DD_INSTRUMENTATION_INSTALL_TIME`, `DD_INSTRUMENTATION_INSTALL_ID`, DD_INSTRUMENTATION_INSTALL_TYPE` env variables to the Trace and Cluster agents to support APM Telemetry KPIs.
```

charts/datadog/templates/_containers-common-env.yaml
liliyadd: Thanks for catching it. This code re-appeared after merge from remote branch.

levan-m (approved):

charts/datadog/templates/kpi-telemetry-configmap.yaml
clamoriniere: Ok but maybe at least when the admission controller is enabled

levan-m: Sorry, I was too quick merging this. This is not in parity with the Operator, which adds these envs only if APM feature is enabled. If we add AC check here, we should update Operator code, otherwise we add APM check here.

liliyadd: > This is not in parity with the Operator, which adds these envs only if APM feature is enabled
This is the [Operator code](https://github.com/DataDog/datadog-operator/blob/5ef94b21b94a17254efadfac73ee189eba7b0c41/controllers/datadogagent/component/clusteragent/default.go#L148-L159) that sets these env variables. I believe we always set these env variables and not only when APM is enabled.

levan-m: I was referring to trace agent, which is not run when APM is disabled so behavior should match. Sorry for confusion.

PR #1272, opened 2023-12-15
Unify Dogstatsd configuration between all containers

hush-hush:

This PR unify the configuration for all agents across the different containers. This PR first focus on Dogstatsd settings.

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_containers-common-env.yaml
+++ b/charts/datadog/templates/_containers-common-env.yaml
@@ -1,5 +1,6 @@
 # The purpose of this template is to define a minimal set of environment
 # variables required to operate dedicated containers in the daemonset
+
 {{- define "containers-common-env" -}}
 - name: DD_API_KEY
   valueFrom:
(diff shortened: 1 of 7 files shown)


charts/datadog/templates/_containers-common-env.yaml
```diff
@@ -150,4 +151,32 @@ Return a list of env-vars if the cluster-agent is enabled
         name: {{ .Values.existingClusterAgent.tokenSecretName | quote }}
         key: token
 {{- end }}
+
+# Dogstatsd configuration
```
clamoriniere, line 155: some of these parameters are only use by the dogstatsd server, so why do we want to add them to all containers?

hush-hush: As part of our OKRs we want to align the configuration between all containers. The goal is for each process/container to see the same configuration. That will solve the discrepancy in the metadata we see in the backend (in remote-config and fleet automation).
Main use case, for now, is for the core-agent to send metadata about other products accurately. For example, with our helm chart, the core-agent doesn't have access to the actual system-probe config. We therefore send the wrong metadata to the backend.

clamoriniere: the main issue that I have with this approach is that we blowup the other processes with useless envvars. 
For the agent deployment in containerise environment we try to follow the cloud native app best practice: https://12factor.net/

hush-hush: This replicates what we do with other deployment methods (Ansible, Chef, DEB, ...) where all processes share the same configuration file and have the same state when it comes to configuration/settings.

Having the same behavior for containers allow us to use the same logic for both backend and agent side with containers deployments. I don't think it's a workaround since it aligns the agents behavior to be the same than in non-k8s world.
Also, I would not call a few envars  a blowup, were only talking about the configuration.

hush-hush:
We decided to go in another direction for the configuration. Closing this PR.

PR #1288, opened 2024-01-10
[synthetics] Add comments to clarify the usage of configSecret for public PL chart

loutPhilipps:
This PR clarifies the usage of `configSecret` in the `synthetics-private-location` chart.

It is only used if you want to manage the secret holding the configuration outside of the Helm chart, and it needs to have a given format (key name needs to be `synthetics-check-runner.json`).


gkampitakis:
From the CI i understand the the README is autogenerated? Maybe we need to update `values.yaml` and generate the readme. Also looks like we need to bump chart version for this change?

PR #1290, opened 2024-01-11
Add option to use containerd snapshotter to generate SBOMs

lebauce:

This PR set up container image scanning using the containerd snapshotter.
To do so, it bind mounts the containerd folder and adds the required
capability to mount container images. This is intended to be used when
facing the `discard_uncompress_layers` issue with containerd.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -3,7 +3,12 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   command: ["agent", "run"]
+{{- if .Values.datadog.sbom.containerImage.uncompressedLayersSupport -}}
+{{- $capabilities := dict "capabilities" (dict "add" (list "SYS_ADMIN")) }}
+{{ include "generate-security-context" (dict "securityContext" (merge $capabilities .Values.agents.containers.agent.securityContext) "targetSystem" "foobar" "seccomp" "" "kubeversion" .Capabilities.KubeVersion.Version) | indent 2 }}
+{{- else }}
 {{ include "generate-security-context" (dict "securityContext" .Values.agents.containers.agent.securityContext "targetSystem" .Values.targetSystem "seccomp" "" "kubeversion" .Capabilities.KubeVersion.Version) | indent 2 }}
+{{- end }}
   resources:
 {{ toYaml .Values.agents.containers.agent.resources | indent 4 }}
   ports:
@@ -171,6 +176,10 @@
     - name: DD_SBOM_CONTAINER_IMAGE_ENABLED
       value: "true"
     {{- end }}
+    {{- if .Values.datadog.sbom.containerImage.uncompressedLayersSupport }}
+    - name: DD_SBOM_CONTAINER_IMAGE_USE_MOUNT
+      value: "true"
+    {{- end }}
     {{- if .Values.datadog.sbom.host.enabled }}
     - name: DD_SBOM_HOST_ENABLED
       value: "true"
@@ -253,6 +262,11 @@
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
     {{- end }}
+    {{- end }}
+    {{- if .Values.datadog.sbom.containerImage.uncompressedLayersSupport }}
+    - name: host-containerd-dir
+      mountPath: /host/var/lib/containerd
+      readOnly: true
     {{- end }}
     {{- if .Values.datadog.sbom.host.enabled }}
     - name: host-apk-dir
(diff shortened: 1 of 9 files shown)


charts/datadog/values.yaml
```diff
@@ -706,6 +706,11 @@ datadog:
       # datadog.sbom.containerImage.enabled -- Enable SBOM collection for container images
       enabled: false
 
+      # datadog.sbom.containerImage.uncompressedLayersSupport -- Use container runtime snapshotter
+      # This should be set to true when using EKS, GKE or if containerd is configured to
+      # discard uncompressed layers.
```
clamoriniere, line 711: could you add in the comment that it will force the addition of the `SYS_ADMIN` capability in the agent container's security context even if the user has defined their own `securityContext`.

charts/datadog/templates/_container-agent.yaml
```diff
@@ -3,7 +3,12 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   command: ["agent", "run"]
+{{- if .Values.datadog.sbom.containerImage.uncompressedLayersSupport -}}
+{{- $capabilities := dict "capabilities" (dict "add" (list "SYS_ADMIN")) }}
+{{ include "generate-security-context" (dict "securityContext" (merge $capabilities .Values.agents.containers.agent.securityContext) "targetSystem" "foobar" "seccomp" "" "kubeversion" .Capabilities.KubeVersion.Version) | indent 2 }}
+{{- else }}
```
clamoriniere, line 9: we already have a function `generate-security-context` that tune the securityContext. Can we add this logic to the function too?

clamoriniere (commented):
Hi @lebauce 

I made some comments. Also could you add or CI example in https://github.com/DataDog/helm-charts/tree/main/charts/datadog/ci
to make sure that the securityContext capability merge function works as expected 🙇

charts/datadog/values.yaml
lebauce: Done

lebauce:
@clamoriniere I have little knowledge regarding apparmor. Could you please provide the annotations that should be added ?

charts/datadog/templates/_container-agent.yaml
lebauce: Done

charts/datadog/templates/daemonset.yaml
```diff
@@ -58,6 +58,9 @@ spec:
         container.seccomp.security.alpha.kubernetes.io/system-probe: {{ .Values.datadog.systemProbe.seccomp }}
         {{- end }}
         {{- end }}
+        {{- if and .Values.datadog.sbom.containerImage.uncompressedLayersSupport }}
```
clamoriniere, line 61: we can add it only if appArmor is enabled too

```suggestion
        {{- if and .Values.agents.podSecurity.apparmor.enabled .Values.datadog.sbom.containerImage.uncompressedLayersSupport }}
```

clamoriniere (commented):
added a small suggestion for the annotation condition.
Also please add test case in the `ci` folder 🙇

clamoriniere:
> @clamoriniere I have little knowledge regarding apparmor. Could you please provide the annotations that should be added ?

Sorry, I just saw your message. I think you found it.

charts/datadog/templates/daemonset.yaml
lebauce: It was therefore - hence the "and" - but I probably removed it by mistake

clamoriniere (approved):
Thanks for all the updates

charts/datadog/values.yaml
```diff
@@ -706,6 +706,12 @@ datadog:
...
+      # This feature will cause the SYS_ADMIN capability to be added to the Agent container.
+      uncompressedLayersSupport: false
```
vboulineau, line 713: I feel like the option name does not exactly represent what it does. Using `mount` has different facets (more performant, more permissions, etc.) and perhaps in the future we'll have another way to support uncompressed layers while still offering the `mount` option.

lebauce: This is intended: the idea of this flag is more to make the feature work in the case of discarded layers. This will allow us to more to a better solution in the future. Naming it `mountImages` would make the migration more complex as we wouldn't know if the feature was specifically request or if it's because it was used as a workaround.
Does it make sense ?

vboulineau: It makes sense to have a dedicated `uncompressedLayersSupport` that activates `useMount`. What does not make sense is that `useMount` does not exist.

I know we did not advertise this option before, but I believe it should have its own existence and we could put in the docstring that `uncompressedLayersSupport` implies `useMount`, with the proper explanation on `useMount`

PR #1292, opened 2024-01-11
Fix host OS detection for SBOMs

lebauce:

This PR adds additional bind mounts of system files for proper OS detection when generating SBOM for the host. 

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -264,6 +264,26 @@
     - name: host-rpm-dir
       mountPath: /host/var/lib/rpm
       readOnly: true
+    {{- if ne .Values.datadog.osReleasePath "/etc/redhat-release" }}
+    - name: etc-redhat-release
+      mountPath: /host/etc/redhat-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/fedora-release" }}
+    - name: etc-fedora-release
+      mountPath: /host/etc/fedora-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/lsb-release" }}
+    - name: etc-lsb-release
+      mountPath: /host/etc/lsb-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/system-release" }}
+    - name: etc-system-release
+      mountPath: /host/etc/system-release
+      readOnly: true
+    {{- end }}
     {{- end }}
     {{- end }}
     {{- if eq .Values.targetSystem "windows" }}
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.50.4
+
+* Mount host files for proper OS detection in SBOMs.
+
 ## 3.50.3
 
 * Set default `Agent` and `Cluster-Agent` version to `7.50.3`.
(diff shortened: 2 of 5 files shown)


charts/datadog/templates/_container-agent.yaml
```diff
@@ -264,6 +264,26 @@
     - name: host-rpm-dir
       mountPath: /host/var/lib/rpm
       readOnly: true
+    {{- if ne .Values.datadog.osReleasePath "/etc/redhat-release" }}
```
clamoriniere, line 267: I think we should us `eq` instead of `ne` here in order to mount the redhat-release volume if the user set it in the `.Values.datadog.osReleasePath`
WDYT?

```suggestion
    {{- if eq .Values.datadog.osReleasePath "/etc/redhat-release" }}
```

charts/datadog/templates/_container-agent.yaml
```diff
@@ -264,6 +264,26 @@
...
+    - name: etc-redhat-release
+      mountPath: /host/etc/redhat-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/fedora-release" }}
```
clamoriniere, line 272:
```suggestion
    {{- if eq .Values.datadog.osReleasePath "/etc/fedora-release" }}
```

charts/datadog/templates/_container-agent.yaml
```diff
@@ -264,6 +264,26 @@
...
+    - name: etc-fedora-release
+      mountPath: /host/etc/fedora-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/lsb-release" }}
```
clamoriniere, line 277:
```suggestion
    {{- if eq .Values.datadog.osReleasePath "/etc/lsb-release" }}
```

charts/datadog/templates/_container-agent.yaml
```diff
@@ -264,6 +264,26 @@
...
+    - name: etc-lsb-release
+      mountPath: /host/etc/lsb-release
+      readOnly: true
+    {{- end }}
+    {{- if ne .Values.datadog.osReleasePath "/etc/system-release" }}
```
clamoriniere, line 282:
```suggestion
    {{- if eq .Values.datadog.osReleasePath "/etc/system-release" }}
```

charts/datadog/templates/_container-agent.yaml
clamoriniere: also instead of adding 4 if condition can we something more generic

something like 

```yaml
    {{- if ne .Values.datadog.osReleasePath "" }}
    - name: etc-os-release
      mountPath: /host/{{.Values.datadog.osReleasePath}}
      readOnly: true
    {{- end }}

```

clamoriniere: After an offline discussion with @lebauce . `ne` is the expected behaviour since we don't want to mount twice the same path if never the user has set a path in `.Values.datadog.osReleasePath`

lebauce: This is intended (and shamelessly copy pasted from somewhere else in the charts).
The idea is to avoid having the volumes to be defined twice in the case where datadog.osReleasePath is set to /etc/redhat-release.

lebauce: Chips

clamoriniere (approved):
We can think late to add an option to void adding every volume mount by default. like it is done with `datadog.systemProbe.enableDefaultOsReleasePaths` option for the system-probe container

PR #1299, opened 2024-01-23
added `otlp.logs.enabled` option to datadog agent to set `DD_OTLP_CONFIG_LOGS_ENABLED`

### LINKED ISSUE #1298: Chart is missing support for OTLP logs
**Describe what happened:**
The helm chart doesn't have an option to enable `DD_OTLP_CONFIG_LOGS_ENABLED`.

I'd expect an option for `otlp.logs.enabled`.

This config is available here - https://github.com/DataDog/datadog-agent/blob/main/pkg/config/config_template.yaml

Is it something that just needs to be added here? https://github.com/DataDog/helm-charts/blob/716fc5c56344d64ddf9a1841e9d70d96a7a8fd94/charts/datadog/templates/_containers-common-env.yaml#L72-L90

marleypowell:
Chart is missing support for OTLP logs.
Added `otlp.logs.enabled` option to datadog agent to set `DD_OTLP_CONFIG_LOGS_ENABLED`.

  - fixes #1298

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_containers-common-env.yaml
+++ b/charts/datadog/templates/_containers-common-env.yaml
@@ -70,6 +70,7 @@
   value: {{ .Values.datadog.containerExcludeLogs | quote }}
 {{- end }}
 {{- if .Values.datadog.otlp }}
+
 {{- if .Values.datadog.otlp.receiver }}
 {{- if .Values.datadog.otlp.receiver.protocols }}
 {{- with .Values.datadog.otlp.receiver.protocols }}
@@ -87,6 +88,16 @@
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{- if .Values.datadog.otlp.logs }}
+{{- with .Values.datadog.otlp.logs }}
+
+- name: DD_OTLP_CONFIG_LOGS_ENABLED
+  value: {{ .enabled | quote }}
+
+{{- end }}
+{{- end }}
+
 {{- end }}
 {{- if eq (include "agent-has-env-ad" .) "true" }}
 {{- if .Values.datadog.dockerSocketPath }}
--- a/charts/datadog/templates/_containers-common-env.yaml
+++ b/charts/datadog/templates/_containers-common-env.yaml
@@ -89,13 +89,11 @@
 {{- end }}
 {{- end }}
 
-{{- if .Values.datadog.otlp.logs }}
 {{- with .Values.datadog.otlp.logs }}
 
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/_containers-common-env.yaml
```diff
@@ -87,6 +88,16 @@
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{- if .Values.datadog.otlp.logs }}
```
clamoriniere, line 92: I think the {{ if ...}} is not needed since {{ with ...}} will be executed only if the value is not empty or I suppose false. 
See: https://helm.sh/docs/chart_template_guide/control_structures/

Otherwise we can only use a "if" since the value will always be "true"

clamoriniere:
Small comment

clamoriniere (commented):
Small comment to simplify the modification

charts/datadog/templates/_containers-common-env.yaml
marleypowell: yeah that seems right:
> Note that we removed the if conditional from the previous exercise because it is now unnecessary - the block after with only executes if the value of PIPELINE is not empty.

I have removed the additional `if`

charts/datadog/templates/_containers-common-env.yaml
```diff
@@ -87,6 +88,14 @@
 {{- end }}
 {{- end }}
 {{- end }}
+
+{{- with .Values.datadog.otlp.logs }}
+
+- name: DD_OTLP_CONFIG_LOGS_ENABLED
+  value: {{ .enabled | quote }}
+
+{{- end }}
```
clamoriniere, line 97: small nit for code clarity: remove the 2 new line between the `{{- with ... }} ... {{- end }}` condition

```suggestion
{{- with .Values.datadog.otlp.logs }}
- name: DD_OTLP_CONFIG_LOGS_ENABLED
  value: {{ .enabled | quote }}
{{- end }}
```

clamoriniere (commented):
small comment,
Also. the command `.github/helm-docs.sh` needs to be run again to update the `README.md` content.

marleypowell:
> small comment, Also. the command `.github/helm-docs.sh` needs to be run again to update the `README.md` content.

Updated the README

clamoriniere (approved):
Hi @marleypowell 

thanks for this contribution, everything looks good now. we will merge this PR soon

Issue #472, opened 2021-12-14
Datadog doesn't collect logs from journald source


ziaenezhad:
I'm using a managed cluster that uses `Systemd-journald` service for collecting and storing logging data. I used annotations as described in the doc but no log collected so far.

### My helmfile:
```yaml
repositories:
- name: datadog
  url: https://helm.datadoghq.com

releases:
  - name: datadog
    chart: datadog/datadog
    namespace: monitoring
    version: 2.27.3
    values:
      - values/datadog.yaml.gotmpl
```

### `values/datadog.yaml.gotmpl`
```yaml
datadog:
  apiKey: {{ .Values.datadog.apiKey }}  
  site: {{ .Values.datadog.site }}  
  tags:
    - "env:{{ .Values.environment }}"

  clusterChecks:
    enabled: false

  logs:
    enabled: true
    containerCollectAll: true

  apm:
    enabled: false
  
  kubelet:
    tlsVerify:  false
  
  processAgent:
    enabled: false
  
  orchestratorExplorer:
    enabled: false
    container_scrubbing:
      enabled: false
  
  env:
    - name: DD_CONTAINER_INCLUDE
      value: "image:registry.gitlab.com/foo/bar/core-service"
    - name: DD_CONTAINER_EXCLUDE
      value: "image:.*"
    - name: DD_ENABLE_PAYLOADS_EVENTS
      value: "false"
    - name: DD_ENABLE_PAYLOADS_SERIES
      value: "false"
    - name: DD_ENABLE_PAYLOADS_SERVICE_CHECKS
      value: "false"
    - name: DD_ENABLE_PAYLOADS_SKETCHES
      value: "false"

clusterAgent:
  enabled: false
```

### My Service deployment
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: core-service
  labels:
    tags.datadoghq.com/env: development
    tags.datadoghq.com/version: "v1.0.0"
spec:
  replicas:  1
  selector:
    matchLabels:
      app: core-service
      type: backend
  template:
    metadata:
      name: core-service
      labels:
        tags.datadoghq.com/env: development
        tags.datadoghq.com/version: "v1.0.0"
        app: core-service
        type: backend
      annotations:
        ad.datadoghq.com/core-service.logs: "[{\"source\": \"journald\", \"service\": \"core-service\"}]"        
    spec:
      containers:
        - name: core-service
          image: registry.gitlab.com/foo/bar/core-service:v1.0.0
[...]

PR #1308, opened 2024-01-30
Add support for mounting host journald persistent dir

rdark:
Enables journald log source to used from DaemonSet

The `journald` log source is not currently usable in context of a DaemonSet agent running on a k8s node. This PR adds support for mounting the host journald persistent dir into the container so that journald logs from the host can be read and submitted by the agent.

* possibly related: #472 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.54.0
+
+* Add support for mounting host journald persistent and runtime directories to the agent - `datadog.journaldMount`
+
 ## 3.53.0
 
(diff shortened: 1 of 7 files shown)


clamoriniere:
Hi @rdark 

Thanks you for this contribution. 
We are trying to limit the number of options in our chart and add only options that are not already covered by another options.

I'm wondering if you have considered using the options `agents.volumeMounts` and `agent.Volumes` for adding the journald host path?

rdark:
> I'm wondering if you have considered using the options `agents.volumeMounts` and `agent.Volumes` for adding the journald host path?

I totally missed this option in the sea of available config; works for this use case just fine (and also supports the argument for limiting the number of options!)

PR #1309, opened 2024-01-31
Exclude agent pod from admission controller library injection

fanny-jiang:

Exclude the agent pod from library injection (receiving new environment variables) when the admission controller is enabled. The agent doesn't send traces and it can be configured to send dogstatsd metrics so it is unnecessary to perform library injection on the agent pod. 

This change will also prevent daemonset/agent pod creation problems in GKE Autopilot, which does not allow privileged pod to be mutated. 

  - fixes #

#### Checklist
- [X] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/daemonset.yaml
+++ b/charts/datadog/templates/daemonset.yaml
@@ -9,7 +9,7 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
     app.kubernetes.io/component: agent
-    admission.datadoghq.com/enabled: false
+    admission.datadoghq.com/enabled: "false"
     {{- if .Values.agents.additionalLabels }}
 {{ toYaml .Values.agents.additionalLabels | indent 4 }}
     {{- end }}
--- a/charts/datadog/templates/daemonset.yaml
+++ b/charts/datadog/templates/daemonset.yaml
@@ -9,9 +9,7 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
     app.kubernetes.io/component: agent
-    {{- if .Values.clusterAgent.admissionController.enabled }}
     admission.datadoghq.com/enabled: false
-    {{- end }}
     {{- if .Values.agents.additionalLabels }}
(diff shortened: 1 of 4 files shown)


charts/datadog/templates/daemonset.yaml
```diff
@@ -9,6 +9,9 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
     app.kubernetes.io/component: agent
+    {{- if .Values.clusterAgent.admissionController.enabled }}
```
clamoriniere, line 12: maybe we dont even needs the `{{- if ...}}` condition since it should have effect on the agent if the admission controller is not enabled

clamoriniere (commented):
that was quick 👏 
thanks @fanny-jiang

clamoriniere (approved):

clamoriniere:
@fanny-jiang the CI job seems to fail on the Daemonset labels

https://github.com/DataDog/helm-charts/actions/runs/7744232710/job/21119084815?pr=1309#step:5:48

charts/datadog/templates/daemonset.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
     app.kubernetes.io/component: agent
+    admission.datadoghq.com/enabled: false
```
clamoriniere, line 12:
```suggestion
    admission.datadoghq.com/enabled: "false"
```

the `false` need to be considered as a string.

clamoriniere (changes requested):
I forgot about the boolean value issue on yaml

clamoriniere (approved):

liliyadd:
Did we add the label in the right place? For the label to be applied to the pods managed by the deployments/daemonset, it needs to be under `spec.template.metadata.labels`. Setting the label under `metadata.labels` only applies it to the deployment/daemonset itself.

liliyadd:
I have created [this PR](https://github.com/DataDog/helm-charts/pull/1330) that should set the label on pods.

PR #1310, opened 2024-02-02
fix(datadog-agent): Properly set DD_APM_ENABLED in the core agent container

BaptisteFoy:
The `DD_APM_ENABLED` value has always been hardcoded to `false` in the core agent container, leading to situations where APM is indeed enabled and the value of `apm.enabled` is set to false in the core agent and true in the trace agent. This hasn't been an issue until we started reporting the agent configuration to the backend, and to display it in the UI. The agent configuration is reported from the core agent, meaning that customers will see `apm.enabled: false` in the UI even if the trace agent exists.

Here we fix this by setting DD_APM_ENABLED to true in the core agent if the trace agent exists.

This has been reported in Zendesk

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Datadog changelog
 
+## 3.53.3
+
+* Set `DD_APM_ENABLED` value in the core agent container to properly report its value.
+
 ## 3.53.2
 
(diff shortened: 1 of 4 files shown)


BaptisteFoy:
There will be a bigger effort to centralize configuration between agent processes, closing this

fanny-jiang (approved):

PR #1318, opened 2024-02-12
[APM Onboarding] Update configuration options for APM Instrumentation

liliyadd:
https://github.com/DataDog/datadog-agent/pull/21673/files#diff-5d1f23f8fcec88188fb13bade49122f4cff59445aa97c64da3b84639d8355040 changed the configuration options for APM Instrumentation. This change was released as part of DCA 7.51.0.

With proposed helm-chart change, the behavior will change in two cases:

1. if users have apm.instrumentation.enabled=false && apm.instrumentation.enabledNamespaces=[ns_a]. Old behavior - SSI will be enabled in ns_a. New behavior - SSI is disabled.
2. if users have apm.instrumentation.enabled=true && apm.instrumentation.enabledNamespaces=[ns_a]. Old behavior - SSI will be enabled in the whole cluster. New behavior - SSI is enabled in ns_a only.

Without proposed change to helm-chart, customers using agent 7.51.0 will already see changed behavior - DCA will NOT instrument any application unless apm.instrumentation.enabled=true. (edited) 

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.5.2
+
+* Add deprecation warning for `DatadogAgent` `v1alpha1` CRD version.
+
 ## 1.5.1
 
 * Add configuration for Operator flag `introspectionEnabled`: this parameter is used to enable the Introspection. It is disabled by default.
--- a/charts/datadog-operator/templates/NOTES.txt
+++ b/charts/datadog-operator/templates/NOTES.txt
@@ -40,6 +40,22 @@ The maximumGoroutines parameter isn't supported by the Operator 1.0.0-rc.12 and 
 Setting a value will not change the default defined in the Operator.
     {{- end }}
 {{- end }}
+
+{{- if (semverCompare ">=1.0.0" .Values.image.tag) }}
+    {{- if .Values.datadogCRDs.migration.datadogAgents.conversionWebhook.enabled }}
+##############################################################################
+####         WARNING: v1alpha1 and conversion webhook deprecation.        ####
+##############################################################################
+
+DatadogAgent v1alpha1 reconciliation in the Operator is deprecated since v1.2.0+ and will be removed in v1.7.0.
+Once removed, the Datadog Operator cannot be configured to reconcile the v1alpha1 DatadogAgent CRD.
+However, you will still be able to apply a v1alpha1 manifest with the conversion webhook enabled (using `datadogCRDs.migration.datadogAgents.conversionWebhook.enabled`).
+DatadogAgent v1alpha1 and the conversion webhook will be removed in v1.8.0.
+See the migration page for instructions on migrating to v2alpha1: https://docs.datadoghq.com/containers/guide/datadogoperator_migration/
+    {{- end }}
+{{- end }}
+
+
 {{- if not (and (semverCompare ">=1.0.0-0" .Values.image.tag) (eq .Values.datadogCRDs.migration.datadogAgents.version "v2alpha1")) }}
 {{- fail "The Datadog Operator `1.0.0` reconciles `DatadogAgent` versions `v2alpha1`. Using an old version of the Datadog Operator (< 1.0.0) with the new version of the DatadogAgent Customer Resource, or the Datadog Operator `1.X` with the `v1alpha1` as stored version of the DatadogAgent is not supported. If you are using a DatadogAgent `v1alpha1`, refer to the Migration Steps: https://github.com/DataDog/helm-charts/blob/main/charts/datadog-operator/README.md#migrating-to-the-version-10-of-the-datadog-operator."}}
 {{- end }}


charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -240,13 +240,13 @@ spec:
           - name: DD_APM_INSTRUMENTATION_ENABLED
             value: "true"
           {{- end }}
-          {{- if ne (include "apmInstrumentation.enabledNamespaces" .) "" }}
+          {{- if .Values.datadog.apm.instrumentation.enabledNamespaces }}
           - name: DD_APM_INSTRUMENTATION_ENABLED_NAMESPACES
-            value: {{ include "apmInstrumentation.enabledNamespaces" . }}
+            value: {{ .Values.datadog.apm.instrumentation.enabledNamespaces | toJson | quote }}
```
clamoriniere, line 245: I was thinking that namespace list separated by a space that the supported format. So why do we convert it to json now?

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -240,13 +240,13 @@ spec:
           - name: DD_APM_INSTRUMENTATION_ENABLED
             value: "true"
```
clamoriniere, line 241: I think this value should now return the value of `.Values.datadog.apm.instrumentation.enabled`

```suggestion
            value: {{ .Values.datadog.apm.instrumentation.enabled | quote }}
```

charts/datadog/templates/NOTES.txt
```diff
@@ -155,24 +155,14 @@ The Datadog Agent is listening on port {{ $apmPort }} for APM service.
...
-
-{{- if and .Values.datadog.apm.instrumentation.enabled .Values.datadog.apm.instrumentation.enabledNamespaces }}
-
-#################################################################
-####               WARNING: Configuration notice             ####
-#################################################################
-
-The options `datadog.apm.instrumentation.enabled` and `datadog.apm.instrumentation.enabledNamespaces` are set together.
-APM Single Step Instrumentation will be enabled in the whole cluster.
+The option `datadog.apm.instrumentation.enabledNamespaces` is set while `datadog.apm.instrumentation.enabled` is disabled.
```
clamoriniere, line 164: Correct me if i'm wrong but the new behaviour is that `datadog.apm.instrumentation.enabled` is true to enable the admission controller and so `datadog.apm.instrumentation.enabledNamespaces` or `datadog.apm.instrumentation.disabledNamespaces` is only took into account if `datadog.apm.instrumentation.enabled: true` 

```suggestion
The option `datadog.apm.instrumentation.enabledNamespaces` can be set while `datadog.apm.instrumentation.enabled` is enabled.
```

clamoriniere: same change need to be done for `disabledNamespaces` a bit bellow in this file. 

IMO the only WARNING linked to SSI in NOTES.txt can be about the fact that the user can't use at the same time `enabledNamespaces` and `disabledNamespaces` options

clamoriniere (commented):
to reduce possible impact on these changes if never a user is using a cluster-agent version < 7.51.0 we should add the `admission.datadoghq.com/enabled=false` on top of the Agent, Cluster-Agent, and Cluster-Check-Runner pods.

charts/datadog/templates/NOTES.txt
```diff
@@ -155,24 +155,14 @@ The Datadog Agent is listening on port {{ $apmPort }} for APM service.
 
 {{- end }}
 
-{{- if and .Values.datadog.apm.instrumentation.enabled_namespaces (eq (include "cluster-agent-enabled" .) "false")}}
+{{- if and .Values.datadog.apm.instrumentation.enableddNamespaces (not .Values.datadog.apm.instrumentation.enabled) }}
```
davidor, line 158: there's a typo: should be `enabledNamespaces`.

charts/datadog/templates/NOTES.txt
liliyadd: > IMO the only WARNING linked to SSI in NOTES.txt can be about the fact that the user can't use at the same time enabledNamespaces and disabledNamespaces options

I agree with this in general. But after we changed configuration options, I think it would be useful to notify customers that their existing configuration won't lead to instrumentation. For example, if users used SSI for the past few months to enable SSI in selected namespaces (when apm.instrumentation.enabled=false) and now upgrade to the new chart, showing the warning will help them to understand why they SSI suddenly stopped working.

charts/datadog/templates/cluster-agent-deployment.yaml
liliyadd: We are reading these env variables in datadog-agent using [config.Datadog.GetStringSlice(...)](https://github.com/DataDog/datadog-agent/blob/main/pkg/clusteragent/admission/mutate/auto_instrumentation.go#L182). GetStringSlice is implemented to [json unmarshal a string to the slice](https://github.com/DataDog/cast/blob/1ee8c8bd14a3d768a7ff681617ed56bc6c204940/caste.go#L1133).
 
Without converting to json, env variables set on DCA would be:
```
    - name: DD_APM_INSTRUMENTATION_ENABLED_NAMESPACES
      value: '[rc-tests apps]'
    - name: DD_APM_INSTRUMENTATION_LIB_VERSIONS
      value: map[java:v1.22.0 python:v1.20.6]
```
and won't be unmarshalled into the slice.

charts/datadog/templates/NOTES.txt
clamoriniere: So maybe instead we should add an error/warning when `datadog.apm.instrumentation.enabled == true && len(datadog.apm.instrumentation.enabledNamespaces) == 0 && len(datadog.apm.instrumentation.disabledNamespaces) == 0`

to say that when `datadog.apm.instrumentation.enabled: true` they must configure either `enabledNamespaces` or `disabledNamespaces`. 

WDYT?

liliyadd:
@adel121 @clamoriniere I've added the description of new config options to the changelog to address concern about customers using new agent version v7.51.0 with older chart version.

charts/datadog/templates/NOTES.txt
```diff
@@ -155,24 +155,14 @@ The Datadog Agent is listening on port {{ $apmPort }} for APM service.
 
 {{- end }}
 
-{{- if and .Values.datadog.apm.instrumentation.enabled_namespaces (eq (include "cluster-agent-enabled" .) "false")}}
+{{- if and .Values.datadog.apm.instrumentation.enabledNamespaces (not .Values.datadog.apm.instrumentation.enabled) }}
```
liliyadd, line 158: This was the typo in the initial version of the check. The configuration name was always `datadog.apm.instrumentation.enabledNamespaces`.

clamoriniere (approved):
LGTM

As we mention offline, This PR introduces a behaviour change for customers that update the chart to the new version but still use an Agent version < 7.51.0. because the agent deployment namespace will not be excluded.
However the PR #1330 should remove the possible impact on the Agent component deployment, by excluding the Datadog agent pods from the APM instrumentation admission controller.

PR #1376, opened 2024-04-17
feat(datadog): support for registry overrrides

### LINKED ISSUE #1078: Datadog chart: use the global container registry setting for admission controller instrumentation init container
**Describe what happened:**

By default, the admission controller injects an init container with the image of `gcr.io/datadoghq/dd-lib-java-init`, even though we override the registry value as:
```yaml
registry: public.ecr.aws/datadog
```

Pod events:
```
Container image "gcr.io/datadoghq/dd-lib-java-init:v1.14.0" already present on machine
```

We expected that the admission controller inherited the global registry setting for the injected init container.

**Workaround:**

We had to additionally inject the following environment variable through the value file.
The reference documentation is [here](https://docs.datadoghq.com/tracing/trace_collection/library_injection_local/?tab=kubernetes).

```yaml
clusterAgent:
  env:
    - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_CONTAINER_REGISTRY
      value: public.ecr.aws/datadog
```

Now it looks better:
```
Pulling image "public.ecr.aws/datadog/dd-lib-java-init:v1.14.0"
```

**Proposal:**

Integrate it into the chart [here](https://github.com/DataDog/helm-charts/blob/5d4eb04e37754bfce11e9fc6512fad2de0fd193d/charts/datadog/templates/cluster-agent-deployment.yaml#L188) by adding something like this:
```
          - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_CONTAINER_REGISTRY
            value: {{ .Values.registry | quote }}
```

stanistan:

(1) Adds support for setting the container registry at the `admissionController` level, which is the [default checked by the datadog-agent](https://github.com/DataDog/datadog-agent/blob/e7aacda40982105c0d9a411f618f268e8d0e9dfd/pkg/clusteragent/admission/mutate/common/common.go#L185-L194).

(2) Adds an override for autoInstrumentation (if present).

(3) Removes the setting of DD_ADMISSION_CONTROLLER_AGENT_SIDECAR_CONTAINER_REGISTRY to registry becauset that behavior is implied by (1).

Fixes #1078 

#### Checklist

- [x] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -1121,6 +1127,12 @@ clusterAgent:
         #       cpu: "2"
         #       memory: "1024Mi"
 
+    autoInstrumentation:
+
+      # clusterAgent.admissionController.autoInstrumentation.containerRegistry -- Override the default registry for autoInstrumentation.
+      containerRegistry:
```
davidor, line 1133: All the options about apm instrumentation are under `apm.instrumentation` so maybe it's better to include the registry there as well.

charts/datadog/templates/_ac-agent-sidecar-env.yaml
```diff
@@ -14,12 +14,22 @@
...
+{{- if .Values.clusterAgent.admissionController.containerRegistry }}
+- name: DD_ADMISSION_CONTROLLER_CONTAINER_REGISTRY
+  value: {{ .Values.clusterAgent.admissionController.containerRegistry }}
+{{- else if .Values.registry }}
+- name: DD_ADMISSION_CONTROLLER_CONTAINER_REGISTRY
+  value: {{ .Values.registry }}
+{{- end }}
+
+{{- if .Values.clusterAgent.admissionController.autoInstrumentation.containerRegistry }}
+- name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_CONTAINER_REGISTRY
```
liliyadd, line 26: If users set DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_CONTAINER_REGISTRY in `clusterAgent.env` or with --set, which value will win? If we are overwriting values that some customers had set manually with  a different value in `.Values.clusterAgent.admissionController.autoInstrumentation.containerRegistry`, we might break their instrumentations.

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Datadog changelog
 
+## 3.59.7
+
+* Add configuration option to specify clusterAgent.admissionController.containerRegistry, which defaults to registry
+* Add configuration option to specify clusterAgent.admissionController.autoInstrumentation.containerRegistry
```
liliyadd, line 6: Do we need both `clusterAgent.admissionController.autoInstrumentation.containerRegistry` and `clusterAgent.admissionController.containerRegistry`?

stanistan: Thinking the same thing, will anyone actually use both? I can drop it.

stanistan:
@liliyadd @davidor Updated the PR... removed the setting for `autoInstrumentation.containerRegistry` entirely until we actually need it.

davidor (approved):

PR #1391, opened 2024-05-08
Set kubelet core check option to be enabled by default

jennchenn:

Enables the kubelet core check by default. We have tested the kubelet core check internally and would like to push it out to a broader audience so we can make any fixes necessary before we can enable the kubelet core check everywhere and eventually deprecate the Python check.

  - fixes #

#### Checklist
- [x] Chart Version bumped


charts/datadog/README.md
```diff
@@ -733,7 +733,7 @@ helm install <RELEASE_NAME> \
 | datadog.kubeStateMetricsEnabled | bool | `false` | If true, deploys the kube-state-metrics deployment |
 | datadog.kubeStateMetricsNetworkPolicy.create | bool | `false` | If true, create a NetworkPolicy for kube state metrics |
 | datadog.kubelet.agentCAPath | string | /var/run/host-kubelet-ca.crt if hostCAPath else /var/run/secrets/kubernetes.io/serviceaccount/ca.crt | Path (inside Agent containers) where the Kubelet CA certificate is stored |
-| datadog.kubelet.coreCheckEnabled | bool | `false` |  |
+| datadog.kubelet.coreCheckEnabled | bool | `true` |  |
```
clamoriniere, line 736: The option description doesn't seems to be extracted properly from the `values.yaml` file.
could you investigate why and fix it? it might be an indentation issue.

jennchenn: Thanks for catching this, the name of the variable did not match in `values.yaml` so the description didn't get populated 🙇

PR #1338, opened 2024-03-04
add language detection apm config

adel121:
#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Verification

- Deploying with `language_detection.enabled: true`:

```
datadog:
  apiKeyExistingSecret: datadog-secret
  appKeyExistingSecret: datadog-secret
  kubelet:
    tlsVerify: false
  apm:
    instrumentation:
      enabled: true
```

From cluster agent config:
```
language_detection:
  enabled: true
  reporting:
    buffer_period: 10s
    enabled: true
    refresh_period: 20m


cluster_agent:
.........
  language_detection:
    cleanup:
      language_ttl: 30m
      period: 10m
    patcher:
      enabled: true

.....

auto_instrumentation:
    container_registry: gcr.io/datadoghq
    enabled: true
    endpoint: /injectlib
    inject_auto_detected_libraries: false
```

From cluster agent cluster role (patch permission is added):
```
- apiGroups:
  - apps
  resources:
  - deployments
  - replicasets
  - daemonsets
  - statefulsets
  verbs:
  - list
  - get
  - watch
  - patch
```

From agent config:
```
language_detection:
  enabled: true
  reporting:
    buffer_period: 10s
    enabled: true
    refresh_period: 20m
```

From process agent config:
```
process_collection:
  enabled: true
```

- Deploying with `language_detection.enabled: false` or with apm instrumentation disabled:

```
datadog:
  apiKeyExistingSecret: datadog-secret
  appKeyExistingSecret: datadog-secret
  kubelet:
    tlsVerify: false
  apm:
    instrumentation:
      enabled: true
      language_detection:
         enabled: false
```

From cluster agent config:
```
language_detection:
  enabled: false
  reporting:
    buffer_period: 10s
    enabled: false
    refresh_period: 20m


cluster_agent:
.........
  language_detection:
    cleanup:
      language_ttl: 30m
      period: 10m
    patcher:
      enabled: false

.....

auto_instrumentation:
    container_registry: gcr.io/datadoghq
    enabled: true
    endpoint: /injectlib
    inject_auto_detected_libraries: false
```

From cluster agent cluster role (patch permission is not added):
```
- apiGroups:
  - ""
  resources:
  - services
  - endpoints
  - pods
  - nodes
  - namespaces
  - componentstatuses
  verbs:
  - get
  - list
  - watch
```

[...]

PR #1410, opened 2024-05-31
do not enable live process collection by default when language detection is enabled

adel121:

This PR removes enabling live process collection by default when language detection is enabled.

Fixes the issue of enabling live process detection by default for APM SSI users.

- [A fix](https://github.com/DataDog/datadog-agent/pull/26116) was implemented in the agent to allow language detection to work in case live process collection is not enabled.
- The initial PR that activated live process collection by default when language detection is enabled can be found [here](https://github.com/DataDog/helm-charts/pull/1338).

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.65.2
+
+* Do not enable live process collection by default when language detection is enabled for `APM SSI`.
+
 ## 3.65.1
 
 * Make sure the security agent is aware of `datadog.securityAgent.runtime.useSecruntimeTrack`.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.65.1
+version: 3.65.2
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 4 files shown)


robertjli (approved):

clamoriniere (commented):
quick remarks
could you link the PR that have introduced the activation by default in the PR description.

Also, where do we activate the langage-detection remote endpoint in the process-agent? is it one by default?

clamoriniere (approved):

adel121:
> quick remarks could you link the PR that have introduced the activation by default in the PR description.
> 
> Also, where do we activate the langage-detection remote endpoint in the process-agent? is it one by default?
> 
> for the second part I found my answer:
> 
> https://github.com/DataDog/helm-charts/blob/b3a6c4d06e1441b813115b9d8a0e741613542202/charts/datadog/templates/_components-common-env.yaml#L4-L7

For your first point, I updated the PR description.

For your second point, I don't think the part you indicated is the relevant one.
By checking the code, it looks like the process collector runs in the process agent by default, and is not controlled by any config flag. (Check [here](https://github.com/DataDog/datadog-agent/blob/e1fc869376d3eb0c366883dd2284f695c584796c/pkg/process/metadata/workloadmeta/collector/process.go#L27-L28)). The part you indicated is the one that allows workload meta remote process collector to receive and process the data that was collected by the process collector in the process agent. But whether this is set or not, I think the process collector will still run.

PR #1439, opened 2024-07-02
[PROCS-3709] Add support to run process checks in core agent

daniel-taf:
This PR adds initial support for running Processes-related checks in the core agent instead of the process-agent. This is documented as an experimental feature. 

N/A

There is a possible misconfiguration: If `processAgent.enabled=true` and the process checks are moved to core agent without other running checks on the core agent, the process agent container enters crash loop. We advise in the documentation to set `processAgent.enabled=false` in that case.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -775,7 +775,7 @@ helm install <RELEASE_NAME> \
 | datadog.processAgent.enabled | bool | `true` | Set this to true to enable live process monitoring agent |
 | datadog.processAgent.processCollection | bool | `false` | Set this to true to enable process collection in process monitoring agent |
 | datadog.processAgent.processDiscovery | bool | `true` | Enables or disables autodiscovery of integrations |
-| datadog.processAgent.runInCoreAgent | bool | `false` | Set this to true to run the following features in the core agent: Live Processes, Live Containers, Process Discovery. # This is an experimental feature requiring Agent 7.53.0+ and Linux. Currently not compatible with APM Single Step Instrumentation. # If Orchestrator Explorer, Network Performance Monitoring is not being used, processAgent.enabled should be set to false to remove the process-agent container. |
+| datadog.processAgent.runInCoreAgent | bool | `false` | Set this to true to run the following features in the core agent: Live Processes, Live Containers, Process Discovery. # This is an experimental feature requiring Agent 7.53.0+ and Linux. Currently not compatible with APM Single Step Instrumentation. # If Network Performance Monitoring is not being used, processAgent.enabled should be set to false to remove the process-agent container. |
 | datadog.processAgent.stripProcessArguments | bool | `false` | Set this to scrub all arguments from collected processes |
 | datadog.prometheusScrape.additionalConfigs | list | `[]` | Allows adding advanced openmetrics check configurations with custom discovery rules. (Requires Agent version 7.27+) |
 | datadog.prometheusScrape.enabled | bool | `false` | Enable autodiscovering pods and services exposing prometheus metrics. |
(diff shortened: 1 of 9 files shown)


charts/datadog/README.md
```diff
@@ -775,6 +775,7 @@ helm install <RELEASE_NAME> \
 | datadog.processAgent.enabled | bool | `true` | Set this to true to enable live process monitoring agent |
 | datadog.processAgent.processCollection | bool | `false` | Set this to true to enable process collection in process monitoring agent |
 | datadog.processAgent.processDiscovery | bool | `true` | Enables or disables autodiscovery of integrations |
+| datadog.processAgent.runInCoreAgent | bool | `false` | Set this to true to run the following features in the core agent: Live Processes, Live Containers, Process Discovery. # This is an experimental feature requiring Agent 7.53.0+ and Linux. Currently not compatible with APM Single Step Instrumentation. # If Orchestrator Explorer, Network Performance Monitoring is not being used, processAgent.enabled should be set to false to remove the process-agent container. |
```
robertjli, line 778: orchestrator is now a core check, it doesn't run in the process agent

charts/datadog/templates/_processes-common-env.yaml
```diff
@@ -0,0 +1,18 @@
+
```
robertjli, line 1: Is this blank line intentional?

robertjli (approved):
A few updates, but good to go afterward

charts/datadog/values.yaml
```diff
@@ -624,6 +624,11 @@ datadog:
     # datadog.processAgent.processDiscovery -- Enables or disables autodiscovery of integrations
     processDiscovery: true
 
+    # datadog.processAgent.runInCoreAgent -- Set this to true to run the following features in the core agent: Live Processes, Live Containers, Process Discovery.
+    ## This is an experimental feature requiring Agent 7.53.0+ and Linux. Currently not compatible with APM Single Step Instrumentation.
+    ## If Orchestrator Explorer, Network Performance Monitoring is not being used, processAgent.enabled should be set to false to remove the process-agent container.
```
celenechang, line 629: to clarify, if a user is using NPM and `runInCoreAgent: true`, then `processAgent.enabled` still should be `true` ?

charts/datadog/templates/_container-agent.yaml
```diff
@@ -58,6 +58,7 @@
   env:
     {{- include "containers-common-env" . | nindent 4 }}
     {{- include "fips-envvar" . | nindent 4 }}
+    {{- include "processes-common-envs" . | nindent 4 }}
```
wiyu, line 61: Any reason not to guard this with a check of `runInCoreAgent`?

charts/datadog/README.md
daniel-taf: I was thinking about old versions where the orchestrator was still on the process agent, but that is < 7.51.0 so it shouldn't actually be an issue. I removed it from the description here.

charts/datadog/values.yaml
daniel-taf: Yes

charts/datadog/templates/_container-agent.yaml
daniel-taf: Added it

celenechang: sorry i think we actually want to keep this as it was, for the effort to have all container configs replicated in the core agent

daniel-taf: No worries, I removed the guard.

celenechang (approved):

PR #1442, opened 2024-07-03
Add support for enabling autoscaling

vboulineau:

Add support for enabling autoscaling

Variable is not yet documented on purpose. Once `datadog-crds` chart is merged, we need to upgrade again the deps in `datadog-operator` and `datadog-agent`

#### Checklist
- [x] Chart Version bumped


clamoriniere (commented):
Looks good.
could you add a scenario in https://github.com/DataDog/helm-charts/tree/main/charts/datadog/ci to validate that the option is working?

vboulineau:
> Looks good. could you add a scenario in https://github.com/DataDog/helm-charts/tree/main/charts/datadog/ci to validate that the option is working?

Added

charts/datadog/requirements.yaml
```diff
@@ -2,7 +2,7 @@ dependencies:
   - name: datadog-crds
     version: 1.0.1
     repository: https://helm.datadoghq.com
-    condition: clusterAgent.metricsProvider.useDatadogMetrics
+    condition: datadog.autoscaling.workload.enabled,clusterAgent.metricsProvider.useDatadogMetrics
```
clamoriniere, line 5: dummy question: `condition` is using "OR" operator between the fields?

charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -344,6 +344,10 @@ spec:
             value: {{ .Values.datadog.prometheusScrape.version | quote }}
           {{- end }}
           {{- end }}
+          {{- if (((.Values.datadog.autoscaling).workload).enabled) }}
```
clamoriniere, line 347: should we also check remote-config enabled option?

charts/datadog/requirements.yaml
vboulineau: Unfortunately no, it's fallback mechanism: it takes the first field that **exists** (not that says `true`), hence it relies on the fact that `datadog.autoscaling.workload.enabled` does not exist in our default `values.yaml`.

charts/datadog/templates/cluster-agent-deployment.yaml
vboulineau: I have added a check in NOTES

clamoriniere (approved):

PR #1443, opened 2024-07-04
Add support for `datadog.profiling`

szegedi:

Adds support for `datadog.profiling` variable to enable Continuous Profiler in pods.

Allows Continuous Profiler to be enabled (or disabled) by Helm Charts config on the controller pod. The cluster-agent, running on the controller pod, will read an environment variable and use this to mutate the configuration of other pods to set the environment variables that will activate profiling within the tracer/profiler client libraries.

* This PR depends on cluster-agent interpreting the environment variable: https://github.com/DataDog/datadog-agent/pull/27185 

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -266,6 +266,10 @@ spec:
           - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_IAST_ENABLED
             value: "true"
           {{- end }}
+          {{- if .Values.datadog.profiling.enabled }}
+          - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_PROFILING_ENABLED
+            value: {{ .Values.datadog.profiling.enabled | quote }}
+          {{- end }}
```
eliottness, line 272: After the reading the doc you created. I don't think this `if` does what you want it to do about `nil` values. cf [helm chart docs](https://helm.sh/docs/chart_template_guide/control_structures/#ifelse)

charts/datadog/values.yaml
```diff
@@ -520,6 +520,28 @@ datadog:
...
 
+  ## Continuous Profiler configuration
+  ##
+  ## Continuous Profiler is disabled by default and can be enabled by setting the `enabled` field to
+  ## either `auto` or `true` value under the `datadog.profiling` section.
+  ## Manually adding the `DD_PROFILING_ENABLED` variable to a pod will take precedence over the
+  ## value in the Helm chart.
+  ## These will only have an effect on containers that have Datadog client libraries installed,
+  ## either manually or via Single Step Instrumentation (under the `datadog.apm.instrumentation`
+  ## section).
```
eliottness, line 531: If you follow the implementation ASM did in the cluster-agent then it will only trigger on pod with SSI and not those with manually installed library. Injecting vars into manually installed pods should be done by the end of Q3 following the work done by APM on service discovery.

szegedi: I pretty much took over the comment from the ASM part as well – it, too, says "either manually or via Single Step Instrumentation" so I guess it should say the same in both places?

charts/datadog/templates/cluster-agent-deployment.yaml
szegedi: I _think_ it does what I want (omits the env var on null), but it doesn't do what I want for "false" (which would be emitting explicit "false" value, but it also omits the env var) so I need to change that. Right now it's:
```
% helm template ./datadog --set datadog.profiling.enabled=null | grep PROFILING -A 1
% helm template ./datadog --set datadog.profiling.enabled=false | grep PROFILING -A 1
% helm template ./datadog --set datadog.profiling.enabled=auto | grep PROFILING -A 1
          - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_PROFILING_ENABLED
            value: "auto"
% helm template ./datadog --set datadog.profiling.enabled=true | grep PROFILING -A 1
          - name: DD_ADMISSION_CONTROLLER_AUTO_INSTRUMENTATION_PROFILING_ENABLED
            value: "true"
```
but with a change to the predicate to say `{{- if not (eq .Values.datadog.profiling.enabled nil) }}` it's doing what I'd expect it to. If there's maybe some other aspect that I'm not seeing, please let me know.

PR #1449, opened 2024-07-10
support setting annotations on datadog-operator deployment

0robustus1:

In our deployment of datadog-operator we want to utilize https://github.com/stakater/Reloader to automatically restart (a.k.a. perform a fresh rollout) the deployment of the operator when secrets (e.g. api-key, app-key) change.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog-operator/templates/deployment.yaml
+++ b/charts/datadog-operator/templates/deployment.yaml
@@ -3,6 +3,10 @@ kind: Deployment
 metadata:
   name: {{ include "datadog-operator.fullname" . }}
   namespace: {{ .Release.Namespace }}
+{{- if .Values.deployment.annotations }}
+  annotations:
+{{ toYaml .Values.deployment.annotations | indent 4 }}
+{{- end }}
   labels:
 {{ include "datadog-operator.labels" . | indent 4 }}
 spec:
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.8.4
+
+* Add option to specify `deployment.annotations`.
+
 ## 1.8.3
 
 * Add `image.doNotCheckTag` option to permit skipping operator image tag compatibility.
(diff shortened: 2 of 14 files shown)


0robustus1:
fyi i was not able to run `make update-test-baselines` successfully. It also fails on the current `main` branch for me.

clamoriniere (approved):

clamoriniere:
> fyi i was not able to run `make update-test-baselines` successfully. It also fails on the current `main` branch for me.

thanks for the information. I have the same issue, we will fix it in on main soon.

charts/datadog-operator/templates/deployment.yaml
```diff
@@ -3,6 +3,8 @@ kind: Deployment
 metadata:
   name: {{ include "datadog-operator.fullname" . }}
   namespace: {{ .Release.Namespace }}
+  annotations:
+{{ include "datadog-operator.deployment.annotations" . | indent 4 }}
```
clamoriniere, line 7: for `labels` we have a helper function that compute the labels from what is provided by the user with `.Values.deployment.labels` and default labels that we always want to add.

but for annotations we don't have this kind of feature so the this line should be replace by 

```suggestion
{{ .Values.deployment.annotations" | indent 4 }}
```

clamoriniere (changes requested):
the linter reviled an issue in the Deployment template.
please check my comment https://github.com/DataDog/helm-charts/pull/1449/files#r1672598611

charts/datadog-operator/templates/deployment.yaml
clamoriniere, line 7: for `labels` we have a helper function that compute the labels from what is provided by the user with `.Values.deployment.labels` and default labels that we always want to add.

but for annotations we don't have this kind of feature so the this line should be replace by something like:

```suggestion
{{ if .Values.deployment.annotations }}
  annotations:
{{ .Values.deployment.annotations | indent 4 }}
{{- end }}
```

clamoriniere: With this change the `make update-test-baselines` doesn't panic anymore

0robustus1: Oh, good point. Fixed.

0robustus1:
Comments are addressed, please take another look.

clamoriniere (approved):

PR #1451, opened 2024-07-10
[PROCS-4201] Add automatic detection for enabling the process agent container

daniel-taf:
This PR adds a helper for determining whether or not the process agent container should be included in the deployment. The main benefit is the removal of the container when its not needed without manual configuration. This PR also cleans up the use of deprecated environment variables and adds configuration for container collection.

Unit tests were added, but the diff is not rendered by default as the change is large.

#### Checklist
- [x] Chart Version bumped


charts/datadog/templates/_helpers.tpl
```diff
@@ -914,4 +914,33 @@ Create RBACs for custom resources
...
+{{- define "should-enable-process-agent" -}}
+  {{- if or .Values.datadog.networkMonitoring.enabled .Values.datadog.serviceMonitoring.enabled -}}
+    true
+  {{- else if and (eq .Values.targetSystem "windows") (eq (include "process-checks-enabled" .) "true") -}}
+    true
+  {{- else if not .Values.agents.image.doNotCheckTag -}}
+    {{- $version := .Values.agents.image.tag | toString | trimSuffix "-jmx" -}}
+    {{- $length := len (split "." $version) -}}
+    {{- if and (eq $length 1) (eq $version "latest") -}}
+      {{- $version = "7.54.0" -}}
```
robertjli, line 934: Is the plan to update this value every time we release a new agent version? Is there some pre-existing variable we can use for this?

charts/datadog/values.yaml
```diff
@@ -605,19 +605,16 @@ datadog:
...
   processAgent:
     # datadog.processAgent.enabled -- Set this to true to enable live process monitoring agent
-
-    ## Note: /etc/passwd is automatically mounted when `processCollection` or `processDiscovery` is enabled.
+    # DEPRECATED. Set `processCollection` or `containerCollection` instead.
+    ## Note: /etc/passwd is automatically mounted when `processCollection`, `processDiscovery`, or `containerCollection` is enabled.
     ## ref: https://docs.datadoghq.com/graphing/infrastructure/process/#kubernetes-daemonset
     enabled: true
 
     # datadog.processAgent.processCollection -- Set this to true to enable process collection in process monitoring agent
```
robertjli, line 613: Should we remove `in process monitoring agent` from the comment here?

robertjli (approved):

charts/datadog/templates/_container-process-agent.yaml
```diff
@@ -79,7 +79,7 @@
       mountPath: /host/sys/fs/cgroup
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
-    {{- if or .Values.datadog.processAgent.processCollection .Values.datadog.processAgent.processDiscovery}}
+    {{- if or .Values.datadog.processAgent.processCollection .Values.datadog.processAgent.processDiscovery .Values.datadog.processAgent.containerCollection}}
```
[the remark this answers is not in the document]
daniel-taf: Maybe its not a big deal, but its not equivalent to any of the helpers I added so I did not.

charts/datadog/templates/_helpers.tpl
daniel-taf: I don't believe there is. The version is updated by changing the default value for `.Values.agents.image.tag` but users can override this. We only need to update the value to be a version recent enough for to be ahead of all version checks so it won't be often. I imagine one more time once we have the GA release of the run in core agent feature.

charts/datadog/templates/_container-process-agent.yaml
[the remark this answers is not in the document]
robertjli: Ah, right, because of the language detection part of `process-checks-enabled`. We can leave as is

charts/datadog/values.yaml
daniel-taf: Done

robertjli (approved):

charts/datadog/templates/_helpers.tpl
celenechang: maybe we could define a helper function that uses nearly the same logic as here: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_helpers.tpl#L3-L20 (which uses a very outdated agent version..)

to return the `$version` value. what do you think?

charts/datadog/values.yaml
```diff
@@ -605,19 +605,16 @@ datadog:
   ## Enable process agent and provide custom configs
   processAgent:
     # datadog.processAgent.enabled -- Set this to true to enable live process monitoring agent
-
-    ## Note: /etc/passwd is automatically mounted when `processCollection` or `processDiscovery` is enabled.
+    # DEPRECATED. Set `processCollection` or `containerCollection` instead.
```
celenechang, line 608:
```suggestion
    # DEPRECATED. Set `datadog.processAgent.processCollection` or `datadog.processAgent.containerCollection` instead.
```

celenechang: do you anticipate this deprecation could break any existing setup for users?

charts/datadog/values.yaml
```diff
@@ -605,19 +605,16 @@ datadog:
...
-    # datadog.processAgent.processCollection -- Set this to true to enable process collection in process monitoring agent
-
-    ## Requires processAgent.enabled to be set to true to have any effect
+    # datadog.processAgent.processCollection -- Set this to true to enable process collection
     processCollection: false
 
     # datadog.processAgent.stripProcessArguments -- Set this to scrub all arguments from collected processes
-
-    ## Requires processAgent.enabled and processAgent.processCollection to be set to true to have any effect
+    ## Requires processAgent.processCollection to be set to true to have any effect
```
celenechang, line 617:
```suggestion
    ## Requires datadog.processAgent.processCollection to be set to true to have any effect
```

charts/datadog/values.yaml
daniel-taf: No I don't believe so, it was only used in the two instances I replaced. The `containerCollection` config essentially supersedes its functionality in terms of enabling the process agent container by default but also allows customers to turn it off in a more obvious way.

charts/datadog/templates/_helpers.tpl
daniel-taf: Would we still need to manually update `$version` it in that case?

celenechang: yes! but i figure it could be a function that can be reused and updated in just one spot

daniel-taf: Sounds good, I added it

celenechang (approved):

PR #1459, opened 2024-07-17
[synthetics]  Add support for PDB for private location deployment

tr3mor:

It is crucial to have at least one instance of a private location running at any moment of time (not to lose visibility of your endpoints). Currently, there is no way to specify [PodDisruptionBudget ](https://kubernetes.io/docs/tasks/run-application/configure-pdb/)to prevent downtime during maintenance of k8s cluster. This pr introduces it (in the similar way it is done for other Dadadog product, for example https://github.com/DataDog/helm-charts/blob/main/charts/observability-pipelines-worker/templates/pdb.yaml

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-private-location/CHANGELOG.md
+++ b/charts/synthetics-private-location/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 0.17.0
+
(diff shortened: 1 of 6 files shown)


loutPhilipps (approved):
LGTM, thanks for the contribution!

PR #1460, opened 2024-07-19
[CONTP-342] Fix container lifecycle

zhuminyi:
CONTP-342
Currently when [datadog.containerLifecycle.enabled](https://github.com/DataDog/helm-charts/blob/b46720a3384eaabe658a8747a4134a99479cddd4/charts/datadog/README.md?plain=1#L700) is false, the setting is ignored. The fix sets a value based on the Helm chart's configuration (true or false), defaults to true if not specified

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -157,10 +157,8 @@
     - name: DD_CHECKS_TAG_CARDINALITY
       value: {{ .Values.datadog.checksCardinality | quote }}
     {{- end }}
-    {{- if .Values.datadog.containerLifecycle.enabled }}
     - name: DD_CONTAINER_LIFECYCLE_ENABLED
-      value: {{ .Values.datadog.containerLifecycle.enabled | quote }}
-    {{- end }}
+      value: {{ .Values.datadog.containerLifecycle.enabled | quote | default "true" }}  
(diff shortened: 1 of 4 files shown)


sblumenthal (approved):

charts/datadog/templates/_container-agent.yaml
```diff
@@ -157,10 +157,8 @@
     - name: DD_CHECKS_TAG_CARDINALITY
       value: {{ .Values.datadog.checksCardinality | quote }}
     {{- end }}
-    {{- if .Values.datadog.containerLifecycle.enabled }}
     - name: DD_CONTAINER_LIFECYCLE_ENABLED
-      value: {{ .Values.datadog.containerLifecycle.enabled | quote }}
-    {{- end }}
+      value: {{ .Values.datadog.containerLifecycle.enabled | default true | quote  }}
```
clamoriniere, line 161: from what I tested, it doesn't work as expected. 

if you set `datadog.containerLifecycle.enabled: false` it still returning "true".
because the `default` function with a boolean value use the default value (in that case `true`) if `false`. 

What I suggest is to juste use `{{ .Values.datadog.containerLifecycle.enabled | quote }}`

or to support an empty ` .Values.datadog.containerLifecycle.enabled` value

```suggestion
      value: {{ .Values.datadog.containerLifecycle.enabled | quote | default "\"true\"" }}
```

PR #1463, opened 2024-07-23
Add terminationGracePeriodSeconds option

aquiladayc:
### What this PR does / why we need it:
Requested via support. This is made by internal TEE and reaching out to Engineering separately.

### Which issue this PR fixes
Add option to set terminationGracePeriodSeconds for the agent, cluster agent and cluster checks runner.

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Datadog changelog
 
+## 3.68.3
+* Add `agents.terminationGracePeriodSeconds`, `clusterAgent.terminationGracePeriodSeconds` and `clusterChecksRunner.terminationGracePeriodSeconds`
+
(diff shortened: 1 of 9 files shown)


nijave:
Any updates on this? It'd be great if we could configure terminationGracePeriodSeconds and the preStop hook which would allow delaying the shutdown of the Agent long enough for other pods on the same draining node to send data.

Right now, the Agent pods terminate while our web pods drain so we lose a small window of data when nodes automatically scale down with the cluster auto scaler.

This would allow us to delay termination of the Agent pods until after our terminationGracePeriodSeconds of our app pods

brycekahle:
@aquiladayc What is the plan for this PR? If it isn't ready, could you please move it to Draft or close it? It shows up in our review requests daily. If you still want it, I suggest rebasing and reaching out to the reviewers.

PR #1477, opened 2024-08-05
OTEL-2003: Bind ports 4317/4318 by default

mackjmr:
This PR binds host ports 4317/4318 by default for the OTel Agent. This change enables Daemonset deployment to work OOTB, without needing to bind the ports manually.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Datadog changelog
 
-## 3.69.3
+## 3.69.4
 
 * Add hostPort `4317` and `4318` for otel agent.
 
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 3.69.3
 
+* Add hostPort `4317` and `4318` for otel agent.
+
+## 3.69.3
+
 * Update `datadog-crds` dependency to `1.7.2`.
 
 ## 3.69.2
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.69.3
+version: 3.69.4
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.69.3](https://img.shields.io/badge/Version-3.69.3-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.69.4](https://img.shields.io/badge/Version-3.69.4-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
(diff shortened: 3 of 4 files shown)


clamoriniere (commented):
Hi @mackjmr 

until now for security reason we are avoiding opening hostPort or hostNetwork by default. The main reason is that if the user, doesn't have proper firewall rules on the host, any attacker can send metrics, traces to the agent from the internet. It can result to a DDOS attack, and it can cost lot of money to the user.

mackjmr:
@clamoriniere don't we auto open hostport for APM port `8126` OOTB ? Or did we remove this ?

clamoriniere:
> @clamoriniere don't we auto open hostport for APM port `8126` OOTB ? Or did we remove this ?

by looking a the code in the trace-agent container definition, the chart doesn't open the `hostport` by default.

The condition is ` {{- if or .Values.datadog.apm.portEnabled .Values.datadog.apm.enabled }}`
and the default values are
```yaml
    # datadog.apm.portEnabled -- Enable APM over TCP communication (hostPort 8126 by default)

    ## ref: https://docs.datadoghq.com/agent/kubernetes/apm/
    portEnabled: false

    # datadog.apm.enabled -- Enable this to enable APM and tracing, on port 8126
    # DEPRECATED. Use datadog.apm.portEnabled instead

    ## ref: https://github.com/DataDog/docker-dd-agent#tracing-from-the-host
    enabled: false

```

see:
*  https://github.com/DataDog/helm-charts/blob/815ab85248e99fdcb0161d65280c3207786f64eb/charts/datadog/templates/_container-trace-agent.yaml#L14-L18
* https://github.com/DataDog/helm-charts/blob/815ab85248e99fdcb0161d65280c3207786f64eb/charts/datadog/values.yaml#L452-L461

mackjmr:
@clamoriniere thanks for the input. I think for APM it makes sense to have this, as by default the UDS socket is enabled instead. However, for the otel-agent we do not have this option, which means that by using our helm the OOTB experience will be broken (unless the app with client is in the same pod as the otel agent, which is the recommended setup). Additionally, looking at the collector helm chart, it looks as though the hostport is open by default: https://github.com/open-telemetry/opentelemetry-helm-charts/blob/main/charts/opentelemetry-collector/values.yaml#L248-L262.

That said, I understand the security concern, and we could solve this with documentation stating that it is required to set hostPort for apps in a different pod than otel agent. I'll check in with our team and circle back here.

clamoriniere:
Hi @mackjmr 

>  I think for APM it makes sense to have this, as by default the UDS socket is enabled instead. However, for the otel-agent we do not have this option

By default the APM (trace-agent) is accessible with UDS, but also with the Agent Kubernetes service: https://github.com/DataDog/helm-charts/blob/b949cce9a3fac04c1af9ca3c1a9dd7bcdfae8a22/charts/datadog/templates/agent-services.yaml#L67-L112
This service is configure to ensure that the Agent behind the K8s service is the agent on the same K8s node (host) without the need of a NodePort. 

for the end user, they need to target the k8s service name which is staticly name base on the agent install name (or can be override if needed with `agents.localService.overrideName` option) instead of the IP of the node.
See this doc: https://docs.datadoghq.com/tracing/guide/setting_up_apm_with_kubernetes_service/?tab=datadogoperator#manual-configuration 

And I think this method can be apply to OTEL ports too.

mackjmr:
Thanks for the suggestion @clamoriniere. We will start by adding documentation, and will consider using the Agent Kubernetes service as well. Closing this PR.

PR #1483, opened 2024-08-08
add nodeAnnotationsAsTags configuration option

adel121:
  - fixes #

#### Checklist
- [ ] Chart Version bumped


charts/datadog/values.yaml
```diff
@@ -233,6 +233,12 @@ datadog:
   #   kubernetes.io/role: kube_role
   #   <KUBERNETES_NODE_LABEL>: <DATADOG_TAG_KEY>
 
+  # datadog.nodeAnnotationsAsTags -- Provide a mapping of Kubernetes Node Annotations to Datadog Tags
+  nodeAnnotationsAsTags:
+    cluster.k8s.io/machine: kube_machine
```
davidor, line 238: This should be commented

adel121: I intend to have this as a default value

adel121: see [here](https://github.com/DataDog/datadog-agent/blob/6c570633d6cc089c433d55bf4b2fba7e83d9943a/pkg/config/setup/config.go#L355)

charts/datadog/values.yaml
```diff
@@ -233,6 +233,12 @@ datadog:
   #   kubernetes.io/role: kube_role
   #   <KUBERNETES_NODE_LABEL>: <DATADOG_TAG_KEY>
 
+  # datadog.nodeAnnotationsAsTags -- Provide a mapping of Kubernetes Node Annotations to Datadog Tags
+  nodeAnnotationsAsTags:
```
davidor, line 237: This option is not used in any template

charts/datadog/values.yaml
davidor: It might be better to not set any default here and let the one in the agent apply if there's no value specified here.

charts/datadog/values.yaml
adel121: My bad, missed adding the template file before committing. 

updated.

davidor (approved):

PR #1489, opened 2024-08-14
Simplify OTel Agent default pipelines

krlv:

Simplifies OTel Agent OOTB pipelines.

- OTel Agent is in private beta; no actual customers will be affected by the change.
- The initial OTel Agent config had 2 traces pipelines: `traces` to compute APM stats and `traces/otlp` _potentially_ for probabilistic sampling. After internal discussions, we decided that `traces/otlp` pipeline is not required in OOTB config.

**Pipelines before:**
OTel-config-2-otel-config drawio

**Pipelines after:**
OTel-config-2-otel-config-simplified drawio

#### Checklist
- [x] Chart Version bumped
- N/A Variables are documented in the `README.md`
- N/A For Datadog Operator chart or value changes update the test baselines (run: `make update-test-baselines`)

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Datadog changelog
 
+## 3.74.0
+
+* Simplify OTel Agent OOTB pipelines:
+  * Remove `traces/otlp` pipeline from the default OTel Agent config
+  * Add `infaattributes` processor and `datadog` exporter to the `traces` pipeline.
+
 ## 3.73.3
 
 * Fix a few typos on OTel Agent configs.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.73.3
+version: 3.74.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/templates/_otel_agent_config.yaml
+++ b/charts/datadog/templates/_otel_agent_config.yaml
@@ -36,12 +36,8 @@ otel-config.yaml: {{- if .Values.datadog.otelCollector.config }} {{ toYaml .Valu
       pipelines:
         traces:
           receivers: [otlp]
-          processors: [batch]
-          exporters: [datadog/connector]
-        traces/otlp:
-          receivers: [otlp]
           processors: [infraattributes, batch]
-          exporters: [datadog]
+          exporters: [datadog, datadog/connector]
         metrics:
           receivers: [otlp, datadog/connector, prometheus]
           processors: [infraattributes, batch]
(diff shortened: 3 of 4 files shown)


dineshg13 (approved):

clamoriniere:
I suppose this PR https://github.com/DataDog/helm-charts/pull/1497 needs to be merge before?

charts/datadog/templates/_otel_agent_config.yaml
```diff
@@ -35,13 +35,9 @@ otel-config.yaml: {{- if .Values.datadog.otelCollector.config }} {{ toYaml .Valu
...
       pipelines:
         traces:
-          receivers: [otlp]
-          processors: [batch]
-          exporters: [datadog/connector]
-        traces/otlp:
           receivers: [otlp]
           processors: [infraattributes, batch]
-          exporters: [datadog]
+          exporters: [datadog, datadog/connector]
```
clamoriniere, line 40: do this change require a minimum agent version?

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.69.3
+version: 3.69.4
```
clamoriniere, line 3: if it is a "breaking change" for customer event if it is on a beta feature, it could be better to bump the minor version, so user might have chance to read the change log.

charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.69.4
+
+* Simplify OTel Agent default pipelines.
```
clamoriniere, line 5: The PR description mention that it might break previous users (still ok because the feature is still in beta). But could you reflect this information in the changelog entry, because by reading this change log it looks very safe to apply this change.

clamoriniere (changes requested):
I added few comments to make communication with users clearer about the breaking change introduced by this PR.

charts/datadog/CHANGELOG.md
krlv: I'd argue that these changes might break previous users 🙂 :

1. All the customers currently participating in the private beta program have their own OTel configs, none of them are relaying on the default config.
2. The simplified pipeline is 100% backward compatible with previous version: the data coming to DD ingest with the new simplified config is identical to the data from previous config.

Nonetheless, I've bumped the minor version as you suggested.

charts/datadog/Chart.yaml
krlv: Fixed, thanks!

charts/datadog/templates/_otel_agent_config.yaml
krlv: This change doesn't require a minimum version.

clamoriniere (approved):

PR #1504, opened 2024-08-26
[PROCS-4293] Improve support for `processAgent.runInCoreAgent` feature

daniel-taf:
 This PR updates the chart logic for using the `processAgent.runInCoreAgent` features. The changes aim to:
 * Make the logic for enabling the process agent container easier to follow.
 * Guard the `DD_PROCESS_CONFIG_RUN_IN_CORE_AGENT_ENABLED` from being mistakenly set to `true` in unsupported environments.
 * Add a method to ignore any compatibility checks by specifying an env var override. This was mostly done to support testing with custom agent images.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_helpers.tpl
+++ b/charts/datadog/templates/_helpers.tpl
@@ -943,8 +943,34 @@ Create RBACs for custom resources
   Return true if any process-related check is enabled
 */}}
 {{- define "process-checks-enabled" -}}
-  {{- if or .Values.datadog.processAgent.containerCollection .Values.datadog.processAgent.processCollection .Values.datadog.processAgent.processDiscovery .Values.datadog.apm.instrumentation.language_detection.enabled -}}
+  {{- if or .Values.datadog.processAgent.containerCollection .Values.datadog.processAgent.processCollection .Values.datadog.processAgent.processDiscovery (eq (include "language-detection-enabled" .) "true") -}}
     true
+  {{- else -}}
+    false
+  {{- end -}}
+{{- end -}}
+
+{{/*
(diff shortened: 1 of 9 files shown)


test/datadog/values/process-run-in-core-envvars.yaml
```diff
@@ -0,0 +1,10 @@
+agents:
```
daniel-taf, line 1: Added this because I could not figure out how to add an envvar override [here](https://github.com/DataDog/helm-charts/blob/c9d89c4a7897f7d969afac382e19e46172e012db/test/datadog/process_agent_test.go#L240). Attempts to represent as a string resulted in render failures :/

charts/datadog/templates/_helpers.tpl
```diff
@@ -943,30 +943,49 @@ Create RBACs for custom resources
...
+      {{- .value -}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+  Returns true if process-related checks should run on the core agent.
+*/}}
+{{- define "should-run-process-checks-on-core-agent" -}}
+  {{- if and (ne .Values.targetSystem "linux") (eq (include "process-checks-enabled" .) "true") -}}
```
robertjli, line 968: Why do we also check if process checks are enabled? Should we always return false if it's not linux?

daniel-taf: Remove the check for process checks being enabled. The check used to be for windows but I thought it would be more sensical to just always return false if were not on Linux. The checks can't run on the core agent otherwise.

fanny-jiang (approved):
LGTM

robertjli (approved):

PR #1511, opened 2024-08-29
[APPS-1924] Add ability to configure connection credentials in config.yaml

liubenjamin:
- Add ability to configure connection credentials (Basic Auth, Token Auth, Jenkins, Postgres) in `config.yaml`
  - [docs](https://docs.datadoghq.com/service_management/workflows/private_actions/private_action_credentials)
- Users shouldn't pull and untar to add their credentials to `secrets.yaml`

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/private-action-runner/README.md.gotmpl
+++ b/charts/private-action-runner/README.md.gotmpl
@@ -22,69 +22,6 @@ helm repo update
 1. Go to the [Private Action Runner tab](https://app.datadoghq.com/workflow/private-action-runners).
 2. Create a new Private Action Runner and follow the instructions for Kubernetes.
 
-## Use this chart with connection credentials
-1. Go to the [Private Action Runner tab](https://app.datadoghq.com/workflow/private-action-runners).
-2. Set up a new Private Action runner by following the Kubernetes instructions. When you reach step 4, instead of running `helm install`, make the following changes to the Helm chart.
-3. Download the chart locally.
-```bash
-helm pull datadog/private-action-runner --untar
-```
-4. Add connection credential json file to `templates/secrets.yaml` in the format corresponding to the credential and action types you want to use.
-
-HTTP Basic Auth:
-```
-{
-   auth_type: 'Basic Auth',
-   credentials: [
-      {
-         username: 'USERNAME',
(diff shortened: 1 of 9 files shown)


charts/private-action-runner/examples/config.yaml
```diff
@@ -35,3 +35,29 @@ runners:
         - com.datadoghq.kubernetes.core.listPod
         - com.datadoghq.http.request
         - com.datadoghq.jenkins.buildJenkinsJob
+
+connectionCredentials:
+  basicAuth:
+    credentials:
+      - key: "basic.json"
```
irubnich, line 42: suggestion - replace `key` with `fileName`. it's more descriptive and is also consistent with the connections interface where you need a file name

charts/private-action-runner/README.md.gotmpl
```diff
@@ -22,69 +22,6 @@ helm repo update
 1. Go to the [Private Action Runner tab](https://app.datadoghq.com/workflow/private-action-runners).
 2. Create a new Private Action Runner and follow the instructions for Kubernetes.
 
-## Use this chart with connection credentials
```
irubnich, line 25: i think it's worth documenting this. credentials are a lot easier to set up now in the chart but i think at least having some copy about how to link the chart credentials to a connection would be useful.

charts/private-action-runner/values.yaml
```diff
@@ -110,3 +110,17 @@ runners:
         # - com.datadoghq.temporal.getWorkflowResult
         # - com.datadoghq.temporal.listWorkflows
         # - com.datadoghq.temporal.runWorkflow
+
+connectionCredentials:
```
irubnich, line 114: a comment here like "see examples/config.yaml for an example" could be helpful

irubnich (commented):
great work! left some suggestions

charts/private-action-runner/README.md.gotmpl
ksun154: I agree I think especially clarifying what they should set for `Path to Secret Location` in the connection UI would be helpful

ksun154 (approved):

PR #1526, opened 2024-09-13
add livenessProbe to values.yaml

### LINKED ISSUE #1223: datadog-operator: livenessProbe isn't possible to edit in values.yaml
**Describe what happened:**
I had an issue this week with the livenessProbe using Helm charts and AWS EKS 1.27 because kubelet was having a tough time connecting to the container. The probe was failing, but the operator was reconciling resources without issue while it was still running until kubelet killed the container and we hosted our own modified version of the chart to solve the issue until we could identify a root cause

One way to handle it would be to edit the livenessProbe directly in the values without having to modify the templates to "putting out fires" and be able to edit the periodSeconds: time.

Create a flag within the deployment.yaml in the livenessProbe with `enabled: true` or `enabled: false`, where `enabled: true` would reference the values.yaml, and if it were `enabled: false`, it would retain the default value within the templates/deployment.yaml, something like that:

```#deployment.yaml
            {{- if .Values.livenessProbe.enabled  }}
          livenessProbe:
            httpGet:
              path: /healthz/
              port: 8081
            periodSeconds: 10
            {{- if .Values.livenessProbe.timeoutSeconds }}
            timeoutSeconds: .Values.livenessProbe.timeoutSeconds
            {{- end }}
            {{- end }}
```

```#values.yaml
          livenessProbe:
            enabled: true
            timeoutInSeconds: 20
```

N/A, this is a feature request

AWS cloud provider and Amazon EKS 1.27

swang392:
Previously, `livenessProbe` was not in `values.yaml` and therefore not configurable. This PR moves `livenessProbe` to `values.yaml` so its values can be edited.

#1223 

#### Checklist
- [x] Chart Version bumped


charts/datadog-operator/values.yaml
```diff
@@ -176,3 +176,13 @@ volumeMounts: []
...
+
+# livenessProbe -- Add default livenessProbe settings
+livenessProbe:
+  periodSeconds: 10
+  httpGet:
+    path: /healthz/
+    port: 8081
+
+# healthPort -- Port number to use in the Operator for the healthz endpoint
+healthPort: 8081
```
levan-m, line 188: We can't parametrize these two unfortunately as they are hard-coded in Operator.

https://github.com/DataDog/datadog-operator/blob/3c38ef29ed6f992459e5a2df4af07a883a73258b/cmd/main.go#L238

https://github.com/DataDog/datadog-operator/blob/3c38ef29ed6f992459e5a2df4af07a883a73258b/pkg/kubernetes/rbac/const.go#L91

We can leave these hard-coded in the template but add [settings](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes) which can change behavior of the liveness probe. For example, [this](https://github.com/DataDog/helm-charts/blob/451b67348d51c0375613353f359042dea14bc66c/charts/datadog/values.yaml#L1276-L1281) is what Agent chart does, we can support same config in Operator.

charts/datadog-operator/templates/_helpers.tpl
```diff
@@ -87,4 +87,17 @@ Check operator image tag version.
...
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{/*
+Returns probe definition based on user settings and default HTTP port.
+Accepts a map with `port` (default port) and `settings` (probe settings).
+*/}}
+{{- define "probe.http" -}}
+{{ if or .settings.httpGet .settings.tcpSocket .settings.exec }}
```
levan-m, line 97: I haven't tested but I doubt Operator can respond to tcpSocket and exec probe requests, so not sure if we need these.

test/datadog-operator/operator_deployment_test.go
```diff
@@ -85,3 +96,11 @@ func verifyDeployment(t *testing.T, manifest string) {
 func verifyAll(t *testing.T, manifest string) {
 	assert.True(t, manifest != "")
 }
+
+func verifyLivenessProbe(t *testing.T, manifest string) {
```
levan-m, line 100: Thanks for adding a test!

charts/datadog-operator/values.yaml
swang392: Would it be okay to keep healthPort (line 187-188)? Like how it is configured in the Agent chart

charts/datadog-operator/values.yaml
```diff
@@ -176,3 +176,7 @@ volumeMounts: []
 #   - name: <VOLUME_NAME>
 #     mountPath: <CONTAINER_PATH>
 #     readOnly: true
+
+# livenessProbe -- Add default livenessProbe settings. HTTP GET is not configurable as it is hardcoded in the Operator.
+livenessProbe:
+  periodSeconds: 10
```
levan-m, line 182: I think it would be more user friendly if we listed properties here either with defaults from agent chart or or k8s [docs](https://pkg.go.dev/k8s.io/api@v0.28.4/core/v1#Probe). This will save them searching names they want to override.

test/datadog-operator/operator_deployment_test.go
```diff
@@ -56,6 +56,18 @@ func Test_operator_chart(t *testing.T) {
 			assertions: verifyAll,
 			skipTest:   SkipTest,
 		},
+		{
+			name: "livenessProbe is correctly configured",
```
levan-m, line 60: This one tests just default case. Consider adding one to verify overrides since this is the feature being added, e.g. 
```go
				Overrides: map[string]string{
					"livenessProbe.timeoutSeconds":    "99",
					"livenessProbe.periodSeconds":    "99",
					"livenessProbe.failureThreshold": "9",
					// etc
				},
```

PR #1548, opened 2024-10-07
embed components common env vars in clc runner

adel121:

Common components env vars defined [here](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/_components-common-env.yaml) are not embedded in CLC runners, which results in missing some configurations in the cluster checks runner. 

The current usecase is for CLC runner to have access to pod labels and annotations as tags.

Language detection env vars should be set only in agent and cluster agent, this is why they were extracted to a separate file.

#### Checklist
- [ ] Chart Version bumped


charts/datadog/templates/agent-clusterchecks-deployment.yaml
```diff
@@ -125,6 +125,7 @@ spec:
 {{- end }}
 {{- end }}
         env:
+          {{- include "components-common-env" . | nindent 10 }}
```
davidor, line 128: Some of the ENVs defined below are part of `components-comment-env`, so now they're redundant. We should check that the values are the same before deleting them.

adel121: Right. Updated the PR

davidor (approved):

PR #1553, opened 2024-10-09
[SBOM] Use container runtime snapshotter by default on EKS and GKE

lebauce:

Detect whether the agent is deployed on EKS or GKE and use the container runtime snapshotter in these cases.

  - fixes #

#### Checklist
- [ ] Chart Version bumped


clamoriniere:
Hi @lebauce 

General question, have we investigated the possibility to handle it directly in the agent code base?
Indeed the missing layer issue is impacting a lot of GKE and EKS users, however it is not the only environment that might be impacted.

Can we easily detect if we should run the snapshotter by default (maybe based on the kubelet or containerd configuration) or maybe try with the default method and fallback automatically with the snapshotter if we get the "missing layer" error when the agent tries to scan a container image.

lebauce:
@clamoriniere 
The problem with the approach is that you first need to detect if we run on EKS/GKE/containerd with discard_uncompressed_layers
[3:06](https://dd.slack.com/archives/C037CDX0WJV/p1728911205257749?thread_ts=1728909391.718149&cid=C037CDX0WJV)
If it's the case, we need to give the CAP_SYS_ADMIN right and bind mount /var/lib/containers
[3:07](https://dd.slack.com/archives/C037CDX0WJV/p1728911269143359?thread_ts=1728909391.718149&cid=C037CDX0WJV)
In don't see how we could give these capabilities after the agent is started and that we detected we needed them

lebauce:
Longer term solution could be to use the agent "overlayfs direct scan" mode which "only" requires mount /var/lib/containers and use by default

clamoriniere:
> Longer term solution could be to use the agent "overlayfs direct scan" mode which "only" requires mount /var/lib/containers and use by default

But in that case, do we loose the image layer information to identify in which layer the package is part of?

lebauce:
No, we don't lose this information.

PR #1558, opened 2024-10-15
Fix autopilot error message

aquiladayc:
Currently, error message is
```
Error: execution error at (datadog/templates/NOTES.txt:341:4): On GKE Autopilot environments, System Probe is not supported. The option 'datadog.securityAgent.runtime.enabled' must be set 'false'
```
but system probe can be enabled by other options that are not mentioned in error message. This PR adds those options.

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.74.3
+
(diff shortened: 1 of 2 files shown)


aquiladayc:
Thanks for review @clamoriniere 
Recreated a PR, can you have a look please? I think I followed instructions but let me know if i missed anything. Thanks!
https://github.com/DataDog/helm-charts/pull/1562

PR #1560, opened 2024-10-15
Added helm chart option to enable event source mapping

ewoodthomas:

https://github.com/DataDog/datadog-agent/pull/26975 sets the Kubernetes event source based on a mapping of controller to integration. This PR adds the helm chart option `datadog.kubernetesEvents.sourceDetectionEnabled` to enable this feature.

#### Checklist

- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/cluster-agent-deployment.yaml
+++ b/charts/datadog/templates/cluster-agent-deployment.yaml
@@ -302,10 +302,8 @@ spec:
           - name: DD_COLLECT_KUBERNETES_EVENTS
             value: {{ .Values.datadog.collectEvents | quote }}
           {{- end }}
-          {{- if .Values.datadog.kubernetesEvents.sourceDetectionEnabled }}
           - name: DD_KUBERNETES_EVENTS_SOURCE_DETECTION_ENABLED
             value: {{ .Values.datadog.kubernetesEvents.sourceDetectionEnabled | quote }}
-          {{- end }}
           - name: DD_CLUSTER_AGENT_KUBERNETES_SERVICE_NAME
             value: {{ template "datadog.fullname" . }}-cluster-agent
           - name: DD_CLUSTER_AGENT_AUTH_TOKEN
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/cluster-agent-deployment.yaml
```diff
@@ -302,6 +302,10 @@ spec:
           - name: DD_COLLECT_KUBERNETES_EVENTS
             value: {{ .Values.datadog.collectEvents | quote }}
           {{- end }}
+          {{- if .Values.datadog.kubernetesEvents.sourceDetectionEnabled }}
+          - name: DD_KUBERNETES_EVENTS_SOURCE_DETECTION_ENABLED
+            value: {{ .Values.datadog.kubernetesEvents.sourceDetectionEnabled | quote }}
+          {{- end }}
```
L3n41c, line 308: I think we should remove the `if` around the definition of the environment variable for the following reason:

if `datadog.kubernetesEvents.sourceDetectionEnabled` is set to `true` in the Helm chart’s `values.yaml`, we explicitly set the env var `DD_KUBERNETES_EVENTS_SOURCE_DETECTION_ENABLED` to `"true"`, which is fine.

But, if `datadog.kubernetesEvents.sourceDetectionEnabled` is set to `false` in the Helm chart’s `values.yaml`, then, we don’t explicitly set the env var `DD_KUBERNETES_EVENTS_SOURCE_DETECTION_ENABLED` at all.

This behavior is fine today because `kubernetes_events_source_detection.enabled` defaults to `false` on agent’s side: https://github.com/DataDog/datadog-agent/blob/ea7d8a5e3cff71282cb53a382533494ee9501b50/pkg/config/setup/config.go#L1627
But when, in the future, `kubernetes_events_source_detection.enabled` will default to `true` in the agent, there will be no way to force it to `false` from the Helm chart anymore.

With the `if` block, the behavior of the `datadog.kubernetesEvents.sourceDetectionEnabled: false` setting is misleading.
It doesn’t mean : “Do not enable events source detection”,
It means “Use the default set on agents’ side, whatever it is.”.

```suggestion
          - name: DD_KUBERNETES_EVENTS_SOURCE_DETECTION_ENABLED
            value: {{ .Values.datadog.kubernetesEvents.sourceDetectionEnabled | quote }}
```

L3n41c (approved):

PR #1573, opened 2024-10-22
Enable support for uncompressed layers by default

lebauce:

`discard_uncompressed_layers` is enabled by default on GKE and EKS, causing all new customers of infrastructure vulnerabilities on these platforms to have a very high number of errors when generating SBOMs for container images.

  - fixes #

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.75.1
+version: 3.76.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.75.0
+version: 3.75.1
 appVersion: "7"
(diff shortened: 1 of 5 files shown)


charts/datadog/values.yaml
```diff
@@ -821,7 +821,7 @@ datadog:
       # This should be set to true when using EKS, GKE or if containerd is configured to
       # discard uncompressed layers.
       # This feature will cause the SYS_ADMIN capability to be added to the Agent container.
```
clamoriniere, line 823: Could you update the description to explains the limitation if they move the option back to `false`: high SBOM scan error rate.

charts/datadog/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: datadog
-version: 3.75.0
+version: 3.75.1
```
clamoriniere, line 3: for this change we should bump the `minor` version instead of the `patch` version

charts/datadog/values.yaml
lebauce: Done

clamoriniere (approved):

PR #1625, opened 2024-12-09
feat(error-tracking): add settings for backend standalone

GianlucaBortoli:

Exposes new settings available for the agent:
* `datadog.apm.errorTrackingStandalone.enabled` controls https://github.com/DataDog/datadog-agent/pull/30065

The equivalent for the operator will be done in a separate PR.

#### Checklist

- [x] Chart Version bumped


Guillaume-Barrier (approved):

charts/datadog/values.yaml
```diff
@@ -1519,6 +1524,11 @@ agents:
   ## get guaranteed delivery of the metrics in Datadog-per-namespace setup!
   enabled: true
 
+  ## Settings for the core agent
+  coreAgent:
```
clamoriniere, line 1528: it doesn't seems to be a good idea to expose a container name (`coreAgent`) to activate a disable a feature. We did it in the past, but we are now trying to chose the settings name base on the end user feature that an internal or technical details.
Also IMO this settings name seems too generic.

I would prefer to see in the setting name a clear and also self explanatory name like `agents.payloadForwarder.enabled` or by looking at the [datadog-agent code base ](https://github.com/DataDog/datadog-agent/blob/8d89378971e1074a64be7313c3bba7034a91fcc4/pkg/config/utils/miscellaneous.go#L36)behind this option `agents.metricsForwarder.enabled`

clamoriniere (commented):
I don’t feel enough knowledgeable enough yet about this new APM feature to recommend the best approach for enabling the "error tracking standalone" functionality.

However, if the APM team intends to reuse the `datadog` Helm chart while enabling only this APM feature, simply disabling the metrics forwarder doesn’t seem optimal. This is because several other functionalities currently run by default in the core-agent or process-agent container and would continue to execute within the core-agent process, consuming CPU and memory resources unnecessarily. We also run by default the cluster-agent component.

levan-m (approved):

PR #1637, opened 2024-12-16
Bump python version to 3.12 in CI

ewoodthomas:

The CI is consistently failing since Python version 3.7 is not found. Seems like GitHub already removed Python 3.7 for Ubuntu 24.04 and will shortly do the same for all other virtual environments.

PR removing 3.7 --> https://github.com/actions/runner-images/issues/10893
Python manifest --> https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json

Also adds `.DS_Store` to `.gitignore` file. File is created by Finder when you open the directory.

Created a dummy commit to trigger the CI and all tests ran successfully with bumped python version.


.DS_Store
jennchenn, line 1: should this file be removed?

ewoodthomas: Yea... not sure why it got added. It's in the .gitignore for the agent, but not in the helm chart. Can add it to the .gitignore file since I think apple always creates this if you look at the directory in finder.

jennchenn (approved):

PR #1653, opened 2024-12-27
[datadog] otel-agent: the otel-agent should be launched with --core-config

truthbk:

The `otel-agent` should be launched with the `--core-config` switch so as to honor other potential configuration directives not present in the OTel configuration (which are Datadog Agent specific). This also includes and affects environment variables, which, as we know, are an important configuration tool in containerized environments. 

  - fixes #

Currently the `otel-agent` does not honor environment variables, or other directives that may be specified in the `datadog.yaml` and that can define important configuration elements for the `otel-agent`.

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/templates/_container-otel-agent.yaml
+++ b/charts/datadog/templates/_container-otel-agent.yaml
@@ -3,10 +3,10 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml"]
+  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml"]
   {{- end -}}
   {{- if eq .Values.targetSystem "windows" }}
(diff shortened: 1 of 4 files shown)


charts/datadog/templates/_container-otel-agent.yaml
```diff
@@ -3,10 +3,10 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml"]
+  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml"]
```
clamoriniere, line 6: do you know if the option is compatible with all agent version? 
if not, are we sure the unknown argument will be ignore at start time and not trigger a process crash?

ankitpatel96 (approved):

charts/datadog/templates/_container-otel-agent.yaml
truthbk: Honestly, I don’t think we should have to make any provisions for older versions (given that we’re at a beta stage for the `otel-agent`). I think other than our first beta images from September, this will work well. We should probably document this in the OTel beta page however.

fanny-jiang (approved):

clamoriniere (approved):

clamoriniere (approved):

PR #1661, opened 2025-01-07
Add DD_AGENT_IPC_* env vars

mackjmr:
This PR adds DD_AGENT_IPC_PORT and DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL env vars in otel agent and core agent. It does not override them if they are added by the user. This is necessary for the otel-agent to pull the api key from core config in the case where backend secrets are used.

This PR also adds flag `--sync-delay=30s` to otel agent.

OTEL-2323

#### Checklist
- [ ] Chart Version bumped


levan-m (commented):
Could you please consider adding test like [this](https://github.com/DataDog/helm-charts/blob/main/test/datadog/process_agent_test.go) to make sure chart doesn't break and it works as intended.

charts/datadog/templates/_container-agent.yaml
```diff
@@ -59,6 +59,38 @@
     {{- include "containers-common-env" . | nindent 4 }}
     {{- include "fips-envvar" . | nindent 4 }}
     {{- include "processes-common-envs" . | nindent 4 }}
+    {{- if eq (include "should-enable-otel-agent" .) "true" }}
+    {{- $found := false }}
+    {{- range .Values.agents.containers.agent.env }}
```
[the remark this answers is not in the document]
mackjmr: Moved logic to helpers in: https://github.com/DataDog/helm-charts/pull/1661/commits/267473ac9b7a1e7e33ff364ef15e28fb71b41c6a.

charts/datadog/templates/_container-agent.yaml
```diff
@@ -59,6 +59,38 @@
...
+    {{- $found = true }}
+    {{- end }}
+    {{- end }}
+    {{- if not $found }}
+    - name: DD_AGENT_IPC_PORT
+      value: "5009"
+    {{- end }}
+    {{- $found := false }}
+    {{- range .Values.agents.containers.agent.env }}
+    {{- if eq .name "DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL" }}
```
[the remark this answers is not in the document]
mackjmr: Youre right, we dont set this in helm charts. This is to make sure we don't override it if the user sets it in `datadog.env`, `Values.agents.containers.agent.env` or `Values.agents.containers.otelAgent.env`.

mackjmr:
Thanks for the review @levan-m, I've addressed the feedback and added a test in: https://github.com/DataDog/helm-charts/pull/1661/commits/a20b97c473482fd4a2dd7b5bcb8eeb66380298b0. Unfortunately, it seems like I am unable to test the option to not override user provided env vars, as setting something like:
```
					"datadog.env": `
- name: DD_AGENT_IPC_PORT
  value: 1234
- name: DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL
  value: 13`,
```
in the test cases does not reflect these env vars in the test results, so I am guessing this gets done at another stage ? 

I was getting the same issue in not being able to test that we don't override in the operator PR: https://github.com/DataDog/datadog-operator/pull/1604#discussion_r1905453677.

charts/datadog/templates/_container-otel-agent.yaml
```diff
@@ -3,10 +3,10 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
   imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
   {{- if eq .Values.targetSystem "linux" }}
-  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml"]
+  command: ["otel-agent", "--config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml", "--sync-delay=30s"]
   {{- end -}}
   {{- if eq .Values.targetSystem "windows" }}
-  command: ["otel-agent", "-foreground", "-config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml"]
+  command: ["otel-agent", "-foreground", "-config={{ template "datadog.otelconfPath" . }}/otel-config.yaml", "--core-config={{ template "datadog.confPath" .  }}/datadog.yaml", "--sync-delay=30s"]
```
clamoriniere, line 9: does this new parameter has been always available in the `otel-agent` ? 

If not, what happen if the `otel-agent` version doesn't support this parameter? is it only a warning or an error?

charts/datadog/templates/_container-agent.yaml
```diff
@@ -59,6 +59,16 @@
     {{- include "containers-common-env" . | nindent 4 }}
     {{- include "fips-envvar" . | nindent 4 }}
     {{- include "processes-common-envs" . | nindent 4 }}
+    {{- if eq (include "should-enable-otel-agent" .) "true" }}
+    {{- if eq (include "agentIpcPort-envvar-agent-present" .) "false" }}
```
clamoriniere, line 63: usually we don't need to check if the user set an envvar because we add additional user envvars at the end of the list and so they override other envvars with the same key name.

charts/datadog/templates/_container-otel-agent.yaml
mackjmr: Appears like this was added x.58: https://github.com/DataDog/datadog-agent/pull/28319. Although please note that the otel-agent is still in beta and the only public docs we have: https://docs.datadoghq.com/opentelemetry/agent/install_agent_with_collector/#configure-the-datadog-agent mention x.61.

charts/datadog/templates/_container-agent.yaml
mackjmr: Good to know, this simplifies the PR a lot. Addressed in https://github.com/DataDog/helm-charts/pull/1661/commits/72eb8feeb23014f29adf9f7d34e13885af5fb104

levan-m (approved):
Looks good to me. Thanks for adding test!
Could you please update from main, bump chart version and regenerate docs.

levan-m (approved):

PR #1674, opened 2025-01-21
do not run process check in core agent on autopilot

### LINKED ISSUE #1671: Issue with 3.88 on autopilot cluster
When updating from 3.83 to 3.88, we end up with a working installation still, but a failed sync error in ArgoCD (see error below).

I _think_ that #1666 may fix this, at least partially. However, I also noticed that 3.88.1 still doesn't remove the volume mounts on `/etc/passwd`:
```
          - name: passwd
            mountPath: /etc/passwd
            readOnly: true
```

Also, I don't see any of the stuff that the sbom changes seemed to be removing in our rendered templates anyway, so want to verify that other fixes are not needed before testing 3.88.1.

```
Failed sync attempt to : one or more objects failed to apply, reason: error when patching "/dev/shm/592578800": admission webhook "warden-validating.common-webhooks.networking.gke.io" denied the request: GKE Warden rejected the request because it violates one or more constraints. Violations details: {"[denied by autogke-no-host-port]":["container agent specifies host ports [8125], which are disallowed in Autopilot."],"[denied by autogke-no-write-mode-hostpath]":["hostPath volume runtimesocketdir used in container agent uses path /var/run/containerd which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/].","hostPath volume procdir used in container agent uses path /proc which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/].","hostPath volume cgroups used in container agent uses path /sys/fs/cgroup which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/].","hostPath volume passwd used in container agent uses path /etc/passwd which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/].","hostPath volume procdir used in container init-config uses path /proc which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/].","hostPath volume runtimesocketdir used in container init-config uses path /var/run/containerd which is not allowed in Autopilot. Allowed path prefixes for hostPath volumes are: [/var/log/]."]}
```

tbavelier:
* Do not run process check in core Agent on autopilot by default as `passwd` mount is not allowed on `agent`, only for `process-agent`

* Fixes #1671 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.88.2
+
+* Disable running process check in core Agent by default feature for GKE Autopilot, as it is not supported.
+
 ## 3.88.1
 
 * Disable SBOM monitoring features for GKE Autopilot, as they are not supported
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: datadog
-version: 3.88.1
+version: 3.88.2
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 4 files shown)


celenechang (approved):

wyardley:
@tbavelier appreciate it. Confirmed that this seems to resolve sync issues for us.

PR #1696, opened 2025-02-10
Add PodResources mount

gjulianm:
#### What this PR does / why we need it:

Adds a mount to expose the PodResources kubelet API in the agent container. This will allow us to retrieve allocated resources for running pods, which is necessary for GPU monitoring (https://github.com/DataDog/helm-charts/pull/1681).

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] `CHANGELOG.md` has been updated
- [x] Variables are documented in the `README.md`
- [x] For Datadog Operator chart or value changes update the test baselines (run: `make update-test-baselines`)

PR #1714, opened 2025-02-24
Reapply "Add PodResources mount (#1696)"

gjulianm:

This PR adds (again) the pod resources socket mount to the agent, which is needed for proper assignment of GPU devices to containers in k8s deployments.

This is a re-application of #1696 which had to be reverted due to an issue with Windows and GKE clusters that prevented deployment.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/test/datadog/baseline/manifests/default_all.yaml
+++ b/test/datadog/baseline/manifests/default_all.yaml
@@ -816,6 +816,8 @@ spec:
               value: "true"
             - name: DD_KUBELET_CORE_CHECK_ENABLED
               value: "true"
+            - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
+              value: /var/lib/kubelet/pod-resources/kubelet.sock
           image: gcr.io/datadoghq/agent:7.63.2
           imagePullPolicy: IfNotPresent
           livenessProbe:
@@ -878,6 +880,9 @@ spec:
               mountPropagation: None
               name: runtimesocketdir
               readOnly: true
+            - mountPath: /var/lib/kubelet/pod-resources
+              name: pod-resources-socket
+              readOnly: false
             - mountPath: /var/run/datadog
               name: dsdsocket
               readOnly: false
@@ -1076,6 +1081,9 @@ spec:
           name: tmpdir
         - emptyDir: {}
           name: s6-run
+        - hostPath:
+            path: /var/lib/kubelet/pod-resources
+          name: pod-resources-socket
         - hostPath:
             path: /proc
           name: procdir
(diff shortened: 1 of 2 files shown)


charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -17,6 +17,9 @@
 - hostPath:
     path: /sys/fs/cgroup
   name: cgroups
+- name: pod-resources-socket
+  hostPath:
+    path: {{ .Values.datadog.kubelet.podResourcesSocketDir }}
```
fanny-jiang, line 22: This hostPath will also break GKE Autopilot installations. Can we add a new condition to prevent mounting in both Autopilot and GDC installations? 

```suggestion
{{- if and (not .Values.providers.gke.gdc) (not .Values.providers.gke.autopilot) }}
- name: pod-resources-socket
  hostPath:
    path: {{ .Values.datadog.kubelet.podResourcesSocketDir }}
{{- end }}
```
This can go after line 12

fanny-jiang (changes requested):

charts/datadog/templates/_daemonset-volumes-linux.yaml
gjulianm: Added both conditions

fanny-jiang (approved):

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -10,6 +10,11 @@
   configMap:
     name: {{ include "agents.confd-configmap-name" . }}
 {{- end }}
+{{- if and (not .Values.providers.gke.gdc) (not .Values.providers.gke.autopilot) }}
+- name: pod-resources-socket
+  hostPath:
```
clamoriniere, line 15: IMO you also need to modify the `_daemonset-volumes-windows.yaml` and us a different path depending of the OS.

you can see how it was done for the log volume https://github.com/DataDog/helm-charts/blob/419250a9913fb33e4cb1986aa2a56a165370181c/charts/datadog/templates/_daemonset-volumes-windows.yaml#L18

clamoriniere: I'm not a windows expert, however for the containerd named pipe we do have a hostpath mount see https://github.com/DataDog/helm-charts/blob/64e474ec3ae886ce2ef2b366ad672b966393bc36/charts/datadog/templates/_daemonset-volumes-windows.yaml#L33 

any way, my initial worries was about trying to mount a volume that wasn't present in the daemonset, but it was because I originally haven't saw the line `{{- if eq .Values.targetSystem "linux" }}` before the "pod-resources-socket" volumemount. 

So windows support should not be broken and can be added later.

gjulianm: Ah, I see. I'm not sure about how named pipes work either tbh, but I'll keep it in mind if we implement support in Windows for this. Thanks!

PR #1719, opened 2025-02-27
[CONTP-679] Fix bug in dca rbac generation for annotations and labels as tags: use deepcopy before merging

adel121:

Fixes bug that causes `DD_KUBERNETES_ANNOTATIONS_AS_TAGS` env var to be incorrectly set to the merged value of `.Values.datadog.kubernetesResourcesLabelsAsTags` and `.Values.datadog.kubernetesResourcesAnnotationsAsTags`.

  - fixes #

The helm chart creates rbacs for the DCA based on user configuration of kubernetes resources labels and annotations as tags. This is needed so that workloadmeta can list and watch related resources to feed the tagger collector with labels and annotations so that it can generate tags.

The rbac generation is done [here](https://github.com/DataDog/helm-charts/blob/19ce2bcdcef0f909c6d101b4df3402422ac5819c/charts/datadog/templates/cluster-agent-rbac.yaml#L510-L564).

We are using mergeOverwrite directly over the values set by the user. (see [here](https://github.com/DataDog/helm-charts/blob/19ce2bcdcef0f909c6d101b4df3402422ac5819c/charts/datadog/templates/cluster-agent-rbac.yaml#L521)).

As a result, `.Values.datadog.kubernetesResourcesAnnotationsAsTags` is being modified in place (it includes a merged version with labels as tags).

Consequently, if the user does this:

```
kubernetesResourcesLabelsAsTags:
  pods:
    foo-1: foo-1
    
kubernetesResourcesAnnotationsAsTags:
  pods:
    foo-2: foo-2
```

The user gets the following env vars:

```
DD_KUBERNETES_RESOURCES_LABELS_AS_TAGS: '{"pods": {"foo-1": "foo-1"}}'
DD_KUBERNETES_RESOURCES_ANNOTATIONS_AS_TAGS: '{"pods":{"foo-1":"foo-1", "foo-2":"foo-2"}}'
```

As you can see, the second env var got a merged version of the two.

[The documentation](https://helm.sh/docs/chart_template_guide/function_list/#mergeoverwrite-mustmergeoverwrite) of helm says the following:

**_This is a deep merge operation but not a deep copy operation. Nested objects that are merged are the same instance on both dicts. If you want a deep copy along with the merge then use the deepCopy function along with merging._**

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/test/datadog/baseline/agent-clusterchecks-deployment_default.yaml
+++ b/test/datadog/baseline/agent-clusterchecks-deployment_default.yaml
@@ -6,13 +6,13 @@ metadata:
   name: datadog-clusterchecks
   namespace: datadog-agent
   labels:
-    helm.sh/chart: 'datadog-3.98.0'
+    helm.sh/chart: "datadog-3.98.1"
     app.kubernetes.io/name: "datadog"
     app.kubernetes.io/instance: "datadog"
     app.kubernetes.io/managed-by: Helm
     app.kubernetes.io/version: "7"
     app.kubernetes.io/component: clusterchecks-agent
-    
+
 spec:
   replicas: 2
   revisionHistoryLimit: 10
@@ -33,7 +33,7 @@ spec:
         app.kubernetes.io/component: clusterchecks-agent
         admission.datadoghq.com/enabled: "false"
         app: datadog-clusterchecks
-        
+
       name: datadog-clusterchecks
       annotations:
         checksum/clusteragent_token: ce75393cbdc42f29bc23068e7ebd685d85a9d00f6eab86c9030153d065d7c2bc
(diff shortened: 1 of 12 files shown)


test/datadog/baseline/agent-clusterchecks-deployment_default.yaml
```diff
@@ -33,144 +33,138 @@ spec:
...
-        args:
-          - for script in $(find /etc/cont-init.d/ -type f -name '*.sh' | sort) ; do bash $script ; done
-        volumeMounts:
-          - name: config
-            mountPath: /etc/datadog-agent
-            readOnly: false # Need RW for writing datadog.yaml config file
-        resources:
-          {}
+        - name: init-volume
+          image: "gcr.io/datadoghq/agent:7.63.0"
```
clamoriniere, line 47: for all this linter change. It would be good to sync with agent-onboarding team to know if we can commit in the repo the linter config so we can all use the same config.

clamoriniere (approved):
The change looks good. However it can be interesting to sync with agent-onboarding team to avoid having change in the `baseline` files if not necessary.

PR #1726, opened 2025-02-28
Add NVIDIA_VISIBLE_DEVICES env var in GPU monitoring

gjulianm:

Include the `NVIDIA_VISIBLE_DEVICES` environment variable in the agent and system-probe containers when GPU monitoring is enabled.

This is needed in addition to the `/var/run/nvidia-container-devices` mount, as the latter will only be recognized by the k8s NVIDIA device plugin when `accept-nvidia-visible-devices-as-volume-mount` is enabled in its configuration.

  - fixes #

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.98.2
+
+* Add the `NVIDIA_VISIBLE_DEVICES` environment variable to the containers when GPU monitoring is enabled, as it might be needed if the NVIDIA device plugin does not have `accept-nvidia-visible-devices-as-volume-mount` enabled.
+
 ## 3.98.1
 
 * Fixes bug that causes `DD_KUBERNETES_ANNOTATIONS_AS_TAGS` env var to be incorrectly set to the merged value of `.Values.datadog.kubernetesResourcesLabelsAsTags` and `.Values.datadog.kubernetesResourcesAnnotationsAsTags`.
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 3.98.2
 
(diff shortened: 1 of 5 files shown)


charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.98.2
+
+* Add the `NVIDIA_VISIBLE_DEVICES` environment variable to the containers when GPU monitoring is enabled, as it might be needed if the NVIDIA device plugin does not have `accept-nvidia-visible-devices-as-volume-mount` enabled.
```
val06, line 5: nit: i would write here the explicit meaning: device plugin doesn't support gpu device mounting via volume mounts and mention in the brackets the name of the parameter (`accept-nvidia-visible-devices-as-volume-mount`)

charts/datadog/templates/_container-agent.yaml
```diff
@@ -205,6 +205,11 @@
     - name: DD_OTELCOLLECTOR_ENABLED
       value: "true"
     {{- end }}
+    {{- if .Values.datadog.gpuMonitoring.enabled }}
+    # depending on the NVIDIA container toolkit configuration, we might need to request visible devices via this env var or via the /var/run/nvidia-container-devices/all volume mount
+    - name: NVIDIA_VISIBLE_DEVICES
+      value: all
+    {{- end }}
```
val06, line 212: does this **always** set this env var? or just in cases where the volume mounts are not available?

val06 (commented):
left comments

charts/datadog/templates/_container-agent.yaml
gjulianm: It always sets it. There's no downside if it's not supported (same as the mounts), it just gets ignored, so I feel it's better to have no configuration and make sure this works out of the box no matter the configuration customers have.

val06 (approved):

tbavelier (approved):
LGTM once CI is green

PR #1733, opened 2025-03-04
feat(ssi): add target based workload selection

betterengineering:
This change implements [Kubernetes SSI | Workload Selection 🎯](https://docs.google.com/document/d/1Lol1ExLF1nGBe7njO6DBVkjuYAYxC1-sL2WP0rdqFRk/edit?usp=sharing) which is being released as part of the Datadog Agent in 7.64.0.

Targets can be found [here](https://github.com/DataDog/datadog-agent/blob/9c678234b26c3f68f00bcedda4dc33bd1b1f4c06/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go#L206-L226) in the cluster agent. The goal is to allow a single operator to be able to configure instrumentation without needing to update any apps or deployments already running there.

* [INPLAT-423](https://datadoghq.atlassian.net/browse/INPLAT-423)

I've added a `values.schema.json` to validate the values for targets and associated unit tests. This is the [blessed way](https://helm.sh/docs/topics/charts/#schema-files) to validate Helm values. While we do [validate these configuration options](https://github.com/DataDog/datadog-agent/blob/9c678234b26c3f68f00bcedda4dc33bd1b1f4c06/pkg/clusteragent/admission/mutate/autoinstrumentation/config.go#L170-L203) in the cluster agent, they simply cause the auto instrumentation webhook to fail to register and it still allows the cluster agent to start. This leaves a pretty glaring issue for a customer. Imagine the following:
* Customer updates config values in Helm with incorrect fields
* Cluster agent starts but fails to register the autoinstrumentation webhook
* All services stop sending telemetry after restarting

By adding validation to the helm chart, we can avoid this situation. For values under `datadog.apm.instrumentation`, there is strict validation so that extraneous properties are not allowed. For all other values, it's business as usual with no validation. I've added additional unit tests to ensure the schema works as expected.

#### Checklist
- [x] Chart Version bumped

[INPLAT-423]: https://datadoghq.atlassian.net/browse/INPLAT-423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/datadog/values.schema.json
+++ b/charts/datadog/values.schema.json
@@ -191,7 +191,7 @@
                           "type": "string"
                         }
                       },
-                      "ddTraceConfig": {
+                      "ddTraceConfigs": {
                         "type": "array",
                         "items": {
                           "type": "object",
--- a/charts/datadog/values.schema.json
+++ b/charts/datadog/values.schema.json
@@ -1,0 +1,300 @@
+{
+  "$schema": "https://json-schema.org/draft-07/schema#",
+  "title": "Values",
+  "type": "object",
+  "properties": {
+    "datadog": {
+      "type": "object",
+      "properties": {
+        "apm": {
+          "type": "object",
+          "properties": {
+            "instrumentation": {
+              "type": "object",
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                },
+                "enabledNamespaces": {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  }
+                },
+                "disabledNamespaces": {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  }
+                },
+                "libVersions": {
+                  "type": "object",
+                  "additionalProperties": {
+                    "type": "string"
+                  }
+                },
+                "targets": {
+                  "type": "array",
+                  "items": {
+                    "type": "object",
+                    "properties": {
+                      "name": {
+                        "type": "string"
+                      },
+                      "podSelector": {
+                        "type": "object",
+                        "properties": {
+                          "matchLabels": {
+                            "type": "object",
+                            "additionalProperties": {
+                              "type": "string"
+                            }
+                          },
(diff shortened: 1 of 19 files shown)


charts/datadog/values.schema.json
```diff
@@ -0,0 +1,300 @@
...
+        "apm": {
+          "type": "object",
+          "properties": {
+            "instrumentation": {
+              "type": "object",
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                },
+                "enabledNamespaces": {
```
stanistan, line 18: Note: For things that are `[]string` we can have `$defs` in place to make this a bit more compact.

```json
{ 
 "...": "...",
 "properties": {
   "enabledNamespaces": { "$ref": "#/$defs/stringArray" },
   "disabledNamespaces": { "$ref": "#/$defs/stringArray" },
 },
 "$defs": {
   "stringArray": { "type": "array", "items": { "type": "string" } }, 
 }
```

This might be more useful for `matchExpressions` and `matchLabels` for the selectors :) 

https://json-schema.org/understanding-json-schema/structuring#defs

test/datadog/apm_instrumentation_test.go
```diff
@@ -0,0 +1,172 @@
...
+					"../../charts/datadog/values.yaml",
+					"values/instrumentation/extra_namespaceselector_key.yaml",
+				},
+			},
+			isValid: false,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
stanistan, line 163: These tests could all be parameterized on a valuesFile and the test creates the chart with default values.

stanistan (approved):
This is great! A couple of stylistic notes.

betterengineering:
I'm going to split this into two PRs

PR #1761, opened 2025-03-19
[errortracking] fix error tracking standalone config

eloytoro:

It's not possible to use `datadog.apm.errorTrackingStandalone.enabled` without running into this error

`json: cannot unmarshal bool into Go struct field EnvVar.spec.template.spec.containers.env.value of type string`

This is because the value of the var is `true` where it should've been `"true"`

It also introduces the `errorTrackingStandalone` object into the schema for validation

#### Checklist
- [ ] Chart Version bumped

### DIFF
--- a/charts/datadog/values.schema.json
+++ b/charts/datadog/values.schema.json
@@ -9,6 +9,14 @@
         "apm": {
           "type": "object",
           "properties": {
+            "errorTrackingStandalone": {
+              "type": "object",
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                }
+              }
+            },
             "instrumentation": {
               "type": "object",
               "properties": {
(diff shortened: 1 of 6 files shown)


charts/datadog/values.schema.json
```diff
@@ -9,6 +9,14 @@
         "apm": {
           "type": "object",
           "properties": {
+            "errorTrackingStandalone": {
+              "type": "object",
+              "properties": {
+                "enabled": {
+                  "type": "boolean"
+                }
+              ]
```
levan-m, line 18: this is redundant I suppose?

eloytoro: just fixed

celenechang (approved):

PR #1766, opened 2025-03-21
[CECO-2073] Align core Agent config with processes-specific config

tbavelier:

* Ensures any specific processes (trace agent, process agent) config (e.g. APM enabled) is also present on the core Agent
    * `system-probe` configuration is done via a ConfigMap already mounted on the core Agent through `/etc/datadog-agent/system-probe.yaml`
    * security-agent configuration will be done in a subsequent PR / is lower priority since today, the core Agent queries the security Agent config (`security-agent.yaml`) so the information is already accurate

#### Motivation

* Ensures the core Agent metadata payload properly reflects configuration

#### Checklist
- [ ] Chart Version bumped


charts/datadog/templates/_container-agent.yaml
```diff
@@ -222,6 +247,14 @@
     - name: NVIDIA_VISIBLE_DEVICES
       value: all
     {{- end }}
+    {{- if eq  (include "should-enable-process-agent" .) "true" }}
```
hmahmood, line 250: There is also a `should-enable-system-probe` macro; any reason not to use that here?

tbavelier: yes, the intent is to make sure the core Agent gets the same config as the other processes. These 2 env vars are solely defined in the process Agent whether or not system-probe is enabled: https://github.com/DataDog/helm-charts/blob/d1a485fb11eb8df1faa3115489ee3ac8831683e7/charts/datadog/templates/_container-process-agent.yaml#L38-L42, so gating them behind should-enable-system-probe could lead to a discrepancy between what's on the process Agent vs what's on the core Agent

hmahmood (approved):

khewonc (approved):

PR #1774, opened 2025-03-27
[ACTP-462] allow to store PAR credential files outside of the helm chart

dd-gplassard:

Allow to retrieve credentials files from existing kubernetes secrets in order to be able to manage them outside of the chart

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/private-action-runner/README.md
+++ b/charts/private-action-runner/README.md
@@ -1,6 +1,6 @@
 # Datadog Private Action Runner
 
-![Version: 0.19.0](https://img.shields.io/badge/Version-0.19.0-informational?style=flat-square) ![AppVersion: v1.1.1](https://img.shields.io/badge/AppVersion-v1.1.1-informational?style=flat-square)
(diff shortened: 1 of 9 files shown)


charts/private-action-runner/README.md
```diff
@@ -60,12 +60,39 @@ runners:
       # privateKey: "STORED_IN_A_SECRET"
 ```
 
+### Use kubernetes secrets to store credentials
+
+If you want to store the credentials outside of the Helm chart, you can create a kubernetes secret and use it in the `values.yaml` file.
+```bash
+# Create a secret with the credentials files
```
merchristK, line 67: would it makes sense to use `gitlab_token.json` and `jenkins_token.json` just to match the public doc and the default values in the UI ?

merchristK (approved):

charts/private-action-runner/README.md
dd-gplassard: Yeah I hesitated a bit but lets do it, I don't see any reason to do it differently.
The complete path might be different since we need to introduce an additional level of directory nesting if we want to mount multiple secrets because they can't be mounted on the same path (`/etc/dd-action-runner/credentials/gitlab/gitlab_token.json` instead of `/etc/dd-action-runner/credentials/gitlab_token.json`) but the filename can be consistent.

PR #1778, opened 2025-03-28
[ACTP-460] PAR helm simplification / best practices

dd-gplassard:

Updates the Chart to simplify it and more closely follow best practices / conventions

Summary of changes
* migrate from an array of `runners` to a single `runner` object. Having multiple runners complicates the usage (passing parameters to the CLI, duplicate names, inconsistencies with shared and specific properties, ...) and we haven't seen any usage of an array of runners. The expected usage would instead be to deploy different revisions
* `image` is a no longer nested under `common`
* replace all names functions with the default ones created from a `helm create` project. (Note : best practice would be to namespace the functions properly but we can do that in a followup PR)
* `credentialFiles` and `credentialSecrets` are under `runner`
* `drop` the default value for `roleType`, it will actually take the one from our `values.yaml` if not specified
* add a `schema.values.json` schema which means helm will fail with an error when running it with the old format. This also gives autocompletion and validation when editing the values file in an IDE.
* more tests
* bump the version to 1.0.0

Still need to do before merging :
* [x] Check all docs for needed updates
* [x] Write the upgrade instructions
* [x] Write instructions on how to use old versions of the chart if needed

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/private-action-runner/examples/values.yaml
+++ b/charts/private-action-runner/examples/values.yaml
@@ -58,7 +58,7 @@ runner:
   # credential files provided here will be mounted in /etc/dd-action-runner/
   # it is safe to remove unneeded files from this section
   credentialFiles:
-    - fileName: "http_basic_creds.json"
+    - fileName: "http_basic.json"
       data: |
         {
           "auth_type": "Basic Auth",
@@ -69,7 +69,7 @@ runner:
             }
           ]
         }
-    - fileName: "http_token_creds.json"
+    - fileName: "http_token.json"
       data: |
         {
           "auth_type": "Token Auth",
@@ -80,7 +80,7 @@ runner:
             }
           ]
         }
-    - fileName: "jenkins_creds.json"
+    - fileName: "jenkins_token.json"
       data: |
         {
           "auth_type": "Token Auth",
@@ -92,7 +92,7 @@ runner:
             }
           ]
         }
-    - fileName: "creds.pgpass"
+    - fileName: "postgresql_token.json"
       data: |
         {
           "auth_type": "Token Auth",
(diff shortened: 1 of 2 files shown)


charts/private-action-runner/examples/values.yaml
```diff
@@ -1,203 +1,202 @@
...
-            "tokenName": "baseURL",
-            "tokenValue": "GITLAB_BASE_URL"
-          },
-          {
-            "tokenName": "gitlabApiToken",
-            "tokenValue": "GITLAB_API_TOKEN"
-          }
-        ]
-      }
+runner:
```
embeaken, line 1: since this isn't an array anymore and all values are under the `runner` key, is there any reason to not just drop the top-level `runner` key entirely?

embeaken (approved):
nice

charts/private-action-runner/examples/values.yaml
dd-gplassard: No super strong reason, I have the feeling we could have "runner" properties (applicative) under the runner object and "infra" properties (kubernetes / ....) as top level ? Maybe it will also be easier to manage properties deprecations ?

charts/private-action-runner/examples/values.yaml
```diff
@@ -1,203 +1,202 @@
...
+          "auth_type": "Token Auth",
+          "credentials": [
+            {
+              "username": "localhost:7233",
+              "token": "TOKEN",
+              "domain": "DOMAIN"
+            }
+          ]
+        }
+    - fileName: "creds.pgpass"
```
merchristK, line 95: are we not changing this file name ? Maybe we should follow the naming in the public doc

merchristK (approved):

PR #1783, opened 2025-03-31
Add CloudPrem Helm chart

guilload:
Add Helm chart for new CloudPrem product (private beta)


urseberry (approved):

fulmicoton-dd (approved):

gabedos (approved):
General chart set up looks good and mirrors other Datadog charts.

fmassot (approved):
LGTM

charts/cloudprem/README.md.gotmpl
```diff
@@ -0,0 +1,197 @@
...
+helm uninstall <release name>
+```
+
+This command removes all the Kubernetes resources associated with the chart and deletes the release.
+
+## Helm Chart values (non-exhaustive)
+
+| Key | Type | Default | Description
+| :--------------- |:---------------:| -----:|--- |
+|aws.accountId | string | null | AWS account ID used for the EKS role ARN service account annotation|
```
clamoriniere, line 179: nits:
In other charts, this table is generated from comments in the values.yaml file.
No need to do the same here, since it would require a significant number of comment updates in values.yaml.
That said, it could be a nice improvement in the future to ensure all options are documented and kept in sync with values.yaml.

.github/ct.yaml
```diff
@@ -2,6 +2,8 @@ remote: origin
 target-branch: main
 chart-dirs:
   - charts
+excluded-charts:
+  - cloudprem
```
clamoriniere: `ct` (chart-testing) is use to detect which chart is modified by a pull request.
If the `cloudprem` chart is excluded the CI will return an error saying that the PR doesnt modify any chart. In that case you will need to add the label `no-chart`

guilload: yep, found that out the hard way

PR #1791, opened 2025-04-02
Allow setting automountServiceAccountToken in the synthetics-private-location chart

### LINKED ISSUE #1790: The synthetics-private-location chart should allow setting automountServiceAccountToken
We have a gatekeeper policy that requires us to set `automountServiceAccountToken: false` on pods. The synthetics-private-location chart currently doesn't allow us to set this property.

moredatapls:

Add `automountServiceAccountToken` to the synthetics-private-location chart to control whether to automatically mount the service account token in the pod. 

This is needed in some cases where automounting the token is prevented by a policy.

Fixes #1790

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/synthetics-private-location/templates/deployment.yaml
+++ b/charts/synthetics-private-location/templates/deployment.yaml
@@ -38,6 +38,9 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ include "synthetics-private-location.serviceAccountName" . }}
+      {{- if .Values.automountServiceAccountToken }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
+      {{- end }}
       hostAliases:
         {{- .Values.hostAliases | toYaml | nindent 8 }}
       containers:
--- a/charts/synthetics-private-location/templates/deployment.yaml
+++ b/charts/synthetics-private-location/templates/deployment.yaml
@@ -38,7 +38,7 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ include "synthetics-private-location.serviceAccountName" . }}
-      {{- if .Values.automountServiceAccountToken }}
+      {{- if not (eq .Values.automountServiceAccountToken nil) }}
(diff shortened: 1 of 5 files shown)


charts/synthetics-private-location/values.yaml
```diff
@@ -37,6 +37,9 @@ serviceAccount:
   # serviceAccount.annotations -- Annotations for the service account
   annotations: {}
 
+# automountServiceAccountToken -- If true, automatically mount the ServiceAccount's API credentials
+automountServiceAccountToken: true
```
loutPhilipps, line 41: Given that we don't include `automountServiceAccountToken` to the deployment if it's not set, should we set no default value to match the previous behiavor?

moredatapls: @loutPhilipps you are right, fixed it

moredatapls:
@loutPhilipps sorry for not replying in so long, resolved your comment. You can review this again

loutPhilipps:
> @loutPhilipps sorry for not replying in so long, resolved your comment. You can review this again

Hey! No problem at all

charts/synthetics-private-location/templates/deployment.yaml
```diff
@@ -39,6 +39,9 @@ spec:
       securityContext:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       serviceAccountName: {{ include "synthetics-private-location.serviceAccountName" . }}
+      {{- if not (eq .Values.automountServiceAccountToken nil) }}
```
loutPhilipps, line 42: I don't think this can work ([proof](https://helm-playground.com/#t=N7C0AIEsDNwOwPYBdwAoCmBHcA6AagIYA2ArugM44ElIIC2CJcSAyugE4BukAxugII8ejZgBUEAa3Rx4kIgEpwAXyUAoarQZNWHbn0HDt4qXABc4ELkKkKVGvRE6uvAUMfHpytSAjSAJl5AA&v=LQhQEMFcBcHsFtaQHbQMoFMBOA3AlgMYYCCBBSqAKrANYbIBcABNFpBkA)). 

Do you want only to be able to set this to false? (as `true` is the default behavior of the kubelet)
```suggestion
      {{- if not .Values.automountServiceAccountToken }}
```

PR #1796, opened 2025-04-03
fail installation if apm and dogstatsd have different hostpath but same mount path

adel121:

Fixes issue where `datadog.apm.hostSocketPath` is not respected by the agent in case apm and dogstatsd sockets are configured with the different host paths but same mount path.

This PR adds an ERROR that shows upon installation of the chart if the configuration results in conflicting hostpath for apm and dogstatsd.

Trying to install the chart with conflicting hostpaths will show the following error:

```
#################################################################
####            ERROR: Conflicting socket host path          ####
#################################################################

Dogstatsd and APM sockets are configured with different paths on the host (datadog.dogstatsd.hostSocketPath and datadog.apm.hostSocketPath).
However, they have the same parent directory in the mount (datadog.dogstatsd.socketPath and datadog.apm.socketPath).

It is not possible to mount two different host paths at the same mount path. 

To resolve this:
- use the same value for datadog.dogstatsd.hostSocketPath and datadog.apm.hostSocketPath
- or use different values for datadog.dogstatsd.socketPath and datadog.apm.socketPath
```

This issue doesn't cause problems in the agent, but rather silently disrespects the user configuration when it is not possible to achieve. 

#### Checklist
- [x] Chart Version bumped

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.114.3
+
+* Show ERROR log if the chart is installed with different values for `datadog.dogstatsd.hostSocketPath` and `datadog.apm.hostSocketPath` while having same parent directories for `datadog.dogstatsd.socketPath` and `datadog.apm.socketPath`.
+
 ## 3.114.2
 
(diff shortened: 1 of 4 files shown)


gabedos (approved):
lgtm 🚀

PR #1805, opened 2025-04-09
Fix autopilot default containers

fanny-jiang:

Fixes bug in helper templates which were causing the system-probe container to enabled by default in GKE Autopilot clusters. This bug was obscured in the CI tests because the Autopilot helper templates use helm built-in methods which do not return the correct values when `helm template` is used. 

Improvements: 
* Updates the GKE Autopilot helpers to use the helm built-in `.Capabilities.APIVersions.Has` to validate cluster instead of performing node lookups. This option should be more reliable in determining WorkloadAllowlist support
* Adds autopilot NPM baseline test
* Fixes and updates tests to render Autopilot-enabled resources properly for the CI using `datadog.envDict.HELM_FORCE_RENDER` as a workaround for the helm template limitations

  - fixes #

#### Checklist


charts/datadog/templates/gke_autopilot_allowlist_synchronizer.yaml
```diff
@@ -1,10 +1,11 @@
-{{- if and .Values.providers.gke.autopilot (eq (include "gke-autopilot-workloadallowlists-enabled" .) "true")}}
+{{- if and .Values.providers.gke.autopilot (eq (include "gke-autopilot-workloadallowlists-enabled" .) "true") (not .Values.datadog.envDict.HELM_FORCE_RENDER) }}
```
fanny-jiang, line 1: Don't render this for the CI because the vanilla k8s clusters will not have the `AllowlistSynchronizer` CRD installed.

tbavelier (approved):

PR #1841, opened 2025-04-25
OTAGENT-387 Add support for configuring DDOT with a ConfigMap

krlv:

Adds support for configuring the DDOT Collector with an existing (external) ConfigMap instead of embedding the YAML in values.yaml.

#### New values

- `datadog.otelCollector.configMap.name` — name of the ConfigMap to mount
- `datadog.otelCollector.configMap.key` — key in that ConfigMap containing the collector config

#### Chart logic

If `datadog.otelCollector.configMap.name` is set, the specified ConfigMap is mounted for `otel-agent`, and collector config points to the referenced key (`datadog.otelCollector.configMap.key`).

If `datadog.otelCollector.configMap.name` is not set, the chart falls back to the current inline-config behaviour (fully backward-compatible).

- [OTAGENT-387](https://datadoghq.atlassian.net/browse/OTAGENT-387)

#### Checklist

[OTAGENT-387]: https://datadoghq.atlassian.net/browse/OTAGENT-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -624,6 +624,12 @@ datadog:
         name: otel-http
     # datadog.otelCollector.config -- OTel collector configuration
     config: null
+    # datadog.otelCollector.configMap -- Use an existing ConfigMap for DDOT Collector configuration
+    configMap:
+      # datadog.otelCollector.configMap.name -- Name of the existing ConfigMap that contains the DDOT Collector configuration
+      name: null
+      # datadog.otelCollector.configMap.key -- Key within the ConfigMap that contains the DDOT Collector configuration
+      key: otel-config.yaml
     # datadog.otelCollector.featureGates -- Feature gates to pass to OTel collector, as a comma separated list
     featureGates: null
 
--- a/test/datadog-operator/baseline/Operator_Deployment_default.yaml
+++ b/test/datadog-operator/baseline/Operator_Deployment_default.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: datadog-agent
   labels:
     app.kubernetes.io/name: datadog-operator
-    helm.sh/chart: datadog-operator-2.9.1
+    helm.sh/chart: datadog-operator-2.9.2
     app.kubernetes.io/instance: datadog-operator
     app.kubernetes.io/version: "1.14.0"
     app.kubernetes.io/managed-by: Helm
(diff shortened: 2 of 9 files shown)


test/datadog-operator/baseline/Operator_Deployment_default.yaml
```diff
@@ -7,7 +7,7 @@ metadata:
   namespace: datadog-agent
   labels:
     app.kubernetes.io/name: datadog-operator
-    helm.sh/chart: datadog-operator-2.9.0
+    helm.sh/chart: datadog-operator-2.9.1
```
krlv, line 10: This is the result of running `make update-test-baselines`. Since [2.9.1](https://github.com/DataDog/helm-charts/blob/main/charts/datadog-operator/Chart.yaml#L3) is the latest version of the Operator chart, I decided to commit this change.

charts/datadog/values.yaml
```diff
@@ -624,6 +624,12 @@ datadog:
         name: otel-http
     # datadog.otelCollector.config -- OTel collector configuration
     config: null
+    # datadog.otelCollector.configMap -- Use an existing ConfigMap for DDOT Collector configuration
+    configMap:
+      # datadog.otelCollector.configMap.name -- Name of the existing ConfigMap that contains the DDOT Collector configuration
+      name: null
+      # datadog.otelCollector.configMap.key -- Key within the ConfigMap that contains the DDOT Collector configuration
+      key: otel-config.yaml
```
mackjmr, line 632: This ensures that when key is unset, it will default to `otel-config.yaml ` right ? In the operator, we don't have support for key, and force users to set the key to `otel-config.yaml`, see https://github.com/DataDog/datadog-operator/blob/main/internal/controller/datadogagent/feature/otelcollector/feature.go#L193

levan-m (approved):

charts/datadog/values.yaml
krlv: Correct, if the user will not specify ConfigMap key they want to use, it will default to `otel-config.yaml`.
Technically speaking, there's no reason to force user to use specific key as long as it maps internally to the correct path in the volume mount (`otel-config.yaml`). I think we should consider **updating the DD Operator implementation** to loosen this requirement and allow users to use whatever key they already have (or prefer to use).

PR #1429, opened 2024-06-17
Add helm chart installation for OTel Agent

dineshg13:
#### What this PR does / why we need it:
Adds OTel Agent container  . OTel Agent is a Datadog's OpenTelemetry's collector distro that runs along with Agent. 

The PR add example collector configuration and corresponding agent configuration. 

To test the changes use the below helm command 

```
helm install my ./charts/datadog --set-file datadog.otelCollector.config=examples/datadog/otel_collector_config.yaml -f examples/datadog/agent_otel_collector.yaml  --set datadog.apiKey=${DD_API_KEY}              
```

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version bumped
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] `CHANGELOG.md` has been updated
- [x] Variables are documented in the `README.md`
- [ ] For Datadog Operator chart or value changes update the test baselines (run: `make update-test-baselines`)

PR #1845, opened 2025-04-29
OTAGENT-386 Add protocol support for DDOT ports

krlv:

1. Adds protocol support to `datadog.otelCollector.ports` to enable non-TCP receivers like StatsD (requires 8125 UDP port). The protocol field defaults to TCP for backward compatibility.

Example usage for StatsD UDP receiver:

```yaml
datadog:
  otelCollector:
    enabled: true
    ports:
      - containerPort: 8125
        hostPort: 8125
        name: otel-statsd
```

2. Wire up `agents.containers.otelAgent.ports` parameter to allow auxiliary port configurations (for debugging, custom extensions, etc.)

Example usage for zPages extension:

```yaml
agents:
  containers:
    otelAgent:
      ports:
        - containerPort: 55679
          name: otel-zpages
```

  - fixes #

#### Checklist


charts/datadog/templates/_container-otel-agent.yaml
```diff
@@ -32,9 +32,12 @@
         {{- if .hostPort }}
         hostPort: {{ .hostPort }}
         {{- end }}
-        protocol: TCP
+        protocol: {{ .protocol | default "TCP" }}
         name: {{ .name }}
     {{- end }}
+{{- if .Values.agents.containers.otelAgent.ports }}
```
mackjmr, line 38: Whats the benefit of adding a new `agents.containers.otelAgent.ports` vs using the existing `datadog.otelCollector.ports` ?

krlv: @mackjmr Just to clarify: I didn't introduce the `agents.containers.otelAgent.ports` parameter -- it was part of @dineshg13 's [original DDOT PR](https://github.com/DataDog/helm-charts/pull/1429/files#diff-2fcec98016e714405bc936fd949012c2e72f23c20a1a94acf695a9cd3e29d75fR1693-R1694) (but actual implementation of `hostPorts` was missing). I had two options here: either remove it entirely or implement it. I chose the latter because having this flexibility might be helpful when dealing with custom extensions, debugging, etc.

The key difference between the `agents.containers.otelAgent.ports` and `datadog.otelCollector.ports` params is their intended use and exposure.

1. `datadog.otelCollector.ports` are automatically exposed as Kubernetes Service ports ([reference](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/agent-services.yaml#L103-L112)). Thus, they're best suited for core OTel Collector functionalities like OTLP or StatsD receivers.
2. `agents.containers.otelAgent.ports` are intended for scenarios where you explicitly want to expose ports only at the host level (`hostPort`) without exposing them as service ports

mackjmr (approved):

charts/datadog/templates/agent-services.yaml
```diff
@@ -102,7 +102,7 @@ spec:
  {{- end }}
 {{- if eq (include "should-enable-otel-agent" .) "true" }}
 {{- range .Values.datadog.otelCollector.ports }}
-    - protocol: TCP
+    - protocol: {{ .protocol | default "TCP" }}
```
levan-m, line 105: nit - default seems redundant since it's set to the `TCP` in values.yaml anyway.

levan-m (approved):

PR #1850, opened 2025-04-30
[OTAGENT-391] Update _container-otel-agent.yaml

songy23:
Set `DD_OTELCOLLECTOR_ENABLED` from `datadog.otelCollector.enabled` in values.yaml

  - fixes #

#### Checklist

### DIFF
--- a/charts/datadog/templates/_container-otel-agent.yaml
+++ b/charts/datadog/templates/_container-otel-agent.yaml
@@ -51,8 +51,10 @@
       value: "5009"
     - name: DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL
       value: "60"
+    {{- if .Values.datadog.otelCollector.enabled }}
     - name: DD_OTELCOLLECTOR_ENABLED
-      value: {{ .Values.datadog.otelCollector.enabled | default "false" }}
+      value: "true"
+    {{- end }}
     {{- include "fips-envvar" . | nindent 4 }}
     - name: DD_LOG_LEVEL
       value: {{ .Values.agents.containers.otelAgent.logLevel | default .Values.datadog.logLevel | quote }}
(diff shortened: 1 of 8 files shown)


krlv (approved):

songy23:
`Go Test Datadog` does not look related, I tested they fail on mainline as well

charts/datadog/templates/_container-otel-agent.yaml
```diff
@@ -51,6 +51,8 @@
       value: "5009"
     - name: DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL
       value: "60"
+    - name: DD_OTELCOLLECTOR_ENABLED
```
levan-m, line 54: this will set env var to boolean `true` instead of boolean string, need to use `quote` or only add this envvar when it's enabled (assuming it's disabled by default in otel agent).

test/datadog/baseline/manifests/otel-agent_logs_collection.yaml
```diff
@@ -1109,6 +1109,8 @@ spec:
               value: "5009"
             - name: DD_AGENT_IPC_CONFIG_REFRESH_INTERVAL
               value: "60"
+            - name: DD_OTELCOLLECTOR_ENABLED
+              value: true
```
levan-m, line 1113: this should be "true"

songy23: fixed

charts/datadog/templates/_container-otel-agent.yaml
songy23: Good catch, yes by default this is disabled, so I change to only add the env var when this config is enabled

levan-m (approved):

PR #1878, opened 2025-05-19
(chore) Clean up CI values files for datadog chart

tbavelier:
* Removes `DD_HOSTNAME` env overrides by instead relying on Agent to get node name using kubelet, which is possible on `kind` using `datadog.kubelet.tlsVerify=false`
* Removes moving images, so CI is consistently tested for a given Agent version.
    * For OTEL tests, we use the `full` suffix to have the `otel-agent` binary, that is not in OOTB images
* Removes `kubeStateMetricsEnabled: false` as it has been the default for a long time, so need to re-specify it
* Specifies for otel:
    * Guarantee string quality of api key using `f0000` instead of `0000`:
        * ~~`'api.key' expected type 'configopaque.String', got unconvertible type 'int', value: '0'`~~ this was fixed in later release but is not in `7.65.x-full`
        * fixed in .67: https://github.com/DataDog/datadog-agent/pull/36449
    * Provides `protocols.grpc` when `receivers.otlp` is provided:
        * `Error: receivers::otlp: must specify at least one protocol when using the OTLP receiver`

* Avoid red-herring errors in CI from hostname not being detected as we're using `kind` for CI
* Removes drift with CI that could get broken if a moving image (e.g. `datadog/agent-dev:nightly-ot-beta-main`) gets updated with new requirements

* CI should be green


.ctignore
```diff
@@ -0,0 +1 @@
+charts/datadog/ci/
```
fanny-jiang, line 1: Let's keep the `/ci` as it is and remove this file

fanny-jiang (approved):
small nit, but otherwise looks good!

.ctignore
tbavelier: yes, forgot to push its removal
image

PR #1924, opened 2025-06-17
Add support for the `extraEnv` in the datadog-operator helm chart.

### LINKED ISSUE #1923: Add Support for Full `env` Passthrough Using `extraEnv` in Datadog Operator Chart
## Description

Yes. The current `values.yaml` only allows limited configuration of environment variables using static `name` + `value` key-pairs. It does not support `valueFrom`, nor does it provide a flexible passthrough of full environment variable definitions.

This limits dynamic use cases, such as injecting environment variables via the Downward API:

```yaml
- name: DD_ENV
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['tags.datadoghq.com/env']
```

## Solution

Introduce a new values field such as extraEnv, which allows full passthrough of structured environment variable definitions.

In values.yaml:
```yaml
extraEnv:
  - name: DD_ENV
    valueFrom:
      fieldRef:
        fieldPath: metadata.labels['tags.datadoghq.com/env']
  - name: DD_VERSION
    valueFrom:
      fieldRef:
        fieldPath: metadata.labels['tags.datadoghq.com/version']
```
In the Helm chart deployment template:
```
        env:
          {{- with .Values.extraEnv }}
          {{- toYaml . | nindent 10 }}
          {{- end }}

```

This approach provides complete flexibility without the need to hardcode Helm logic for value or valueFrom.

barucijah:
  - fixes #1923 

#### Checklist

### DIFF
--- a/test/datadog-operator/baseline/Operator_Deployment_default.yaml
+++ b/test/datadog-operator/baseline/Operator_Deployment_default.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: datadog-agent
   labels:
     app.kubernetes.io/name: datadog-operator
-    helm.sh/chart: datadog-operator-2.11.0-dev.2
+    helm.sh/chart: datadog-operator-2.11.0-dev.3
     app.kubernetes.io/instance: datadog-operator
     app.kubernetes.io/version: "1.16.0-rc.1"
     app.kubernetes.io/managed-by: Helm


swang392:
Hi @barucijah, we already have an `env` field in the `datadog-operator` helm chart with the same configuration and functionality as `extraEnv`. Could you explain the potential use case for `extraEnv` that's not supported by `env`?

barucijah:
Hi @swang392,

In my use case, I need to inject values into environment variables using Pod metadata labels via the Kubernetes Downward API. However, the current `env` configuration in the chart only supports static `name`/`value` pairs, which makes this type of dynamic configuration impossible.

For example, I’d like to use the following:

```yaml
- name: DD_ENV
  valueFrom:
    fieldRef:
      fieldPath: metadata.labels['tags.datadoghq.com/env']
```

This pattern is not currently supported due to the hardcoded template logic that expects only name and value. It would be great to support use cases like this if the chart allowed raw passthrough of full env object, perhaps via an extraEnv field rendered with toYaml.

Thanks in advance for considering this enhancement!

swang392:
Hi, after discussing with my team, we think it makes more sense to add a field under `env` instead of adding another similar `extraEnv` value which could confuse users. Instead, could you follow an [implementation](https://github.com/DataDog/helm-charts/blob/cf49e8ec9e8e8b3623add1bdacd0d8ce0562b4e8/charts/datadog/templates/_helpers.tpl#L840-L854) like the `datadog` helm chart which handles `value` and `valueFrom` with the same `env` configuration?

PR #1949, opened 2025-07-03
Update RBAC for CRDs metric collection

gabedos:

RBACs for the KSM custom resources metric collection need to be updated to properly use the entity "resource" type instead of "kind + "s".

We will expose the option for the user to provide the **_resource_** type of the CRD so that RBACs can be generated using the specified resource name.

  - fixes #

CONS-7448

#### Checklist

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.123.0
+
+* Update RBAC for CRDs metric collection ([#1949](https://github.com/DataDog/helm-charts/pull/1949)).
+
 ## 3.122.0
 
 * Support a lifecycle handler for the agent via `agents.lifecycle`.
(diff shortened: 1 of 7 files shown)


swang392 (approved):

test/integ/operator_integ_test.go
```diff
@@ -19,8 +19,8 @@ import (
 )
 
 const (
-	apiKeyEnv     = "API_KEY"
-	appKeyEnv     = "APP_KEY"
+	apiKeyEnv     = "DD_API_KEY"
+	appKeyEnv     = "DD_APP_KEY"
```
clamoriniere: just checking if this change wasn't only for local testing

gabedos: Ah yes, I often make this change to have local testing work. Will revert it back but it seems like it doesn't break the tests in the CI.

Could be worth intentionally changing these envvar keys in the future to have the "DD" prefix since that aligns more with the standard naming convention we use for our Datadog API and APP keys.

PR #1967, opened 2025-07-22
fix(NOTES.txt): avoid "incompatible types for comparison" error when …

y-kacher:

Helm fails with a type mismatch when comparing a nil value to the string "csi" in the admission controller warning block : 

> Error: INSTALLATION FAILED: template: datadog/templates/NOTES.txt:715:48: executing "datadog/templates/NOTES.txt" at <eq .Values.clusterAgent.admissionController.configMode "csi">: error calling eq: incompatible types for comparison

This change guards the comparison by checking if `clusterAgent.admissionController.configMode` is defined before using `eq`, avoiding the error when it is not set.

#### Checklist

### DIFF
--- a/charts/datadog/templates/NOTES.txt
+++ b/charts/datadog/templates/NOTES.txt
@@ -712,7 +712,7 @@ To learn more about it please refer to the following documentation:
 https://docs.datadoghq.com/agent/guide/fips-agent/
 {{- end }}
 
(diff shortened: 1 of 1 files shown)


adel121:
Hello @y-kacher 

Thanks for fixing this. 

Are you sure it is causing an issue? 

I checked out the main branch, and installed the agent with `helm install datadog-agent ./charts/datadog/` without setting the configMode, and keeping the default value of csi.enabled as `false`, but I didn't get any error.

PR #1969, opened 2025-07-22
Update helm chart to enable advanced dispatching by default

zhuminyi:
  - fixes #

#### Checklist

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -1189,6 +1189,16 @@ clusterAgent:
   # clusterAgent.shareProcessNamespace -- Set the process namespace sharing on the Datadog Cluster Agent
   shareProcessNamespace: false
 
+  # clusterAgent.clusterChecks -- define cluster checks config options
+  clusterChecks:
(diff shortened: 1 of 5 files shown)


charts/datadog/values.yaml
```diff
@@ -1189,6 +1189,16 @@ clusterAgent:
   # clusterAgent.shareProcessNamespace -- Set the process namespace sharing on the Datadog Cluster Agent
   shareProcessNamespace: false
 
+  # clusterAgent.clusterChecks -- define cluster checks config options
+  clusterChecks:
```
davidor, line 1193: I’m not sure if we should include these options in the Helm chart. They’re very specific.
Wouldn’t it be better to update the defaults in the agent so they apply to both Helm and the operator?

tbavelier: Fully agreed with @davidor , making it the default, the only value users would somehow want to tweak could be the excluded checks (and that's in rare cases), and even then, don't think it's worth its own parameter for it.

PR #2000, opened 2025-08-12
feat: Made `PodDisruptionBudget` configurable

houman-moallemi:
### Core Issue
At the moment it's not possible to configure the `PodDisruptionBudget` for the cluster agent. The existing configuration is hardcoded and set to `minAvailable: 1` which isn't highly available.

By allowing users to set `minAvailable` or `maxUnavailable` we can achieve proper HA in any respective environment.
With this example configuration we can have a highly available cluster agent.
- CA `replicas: 3`
- PDB `maxUnavailable: 1`

### Interface Consistency
In order to have a consistent interface in the chart, the same changes for `clusterAgent` are applied to `clusterChecksRunner` as well. Previous behavior are kept and changes are backwards compatible.

Both `createPodDisruptionBudget` are copied into the new `pdb` block so we can deprecate the additional flag in a future minor release.

#### Checklist

### DIFF
--- a/charts/datadog/templates/agent-clusterchecks-pdb.yaml
+++ b/charts/datadog/templates/agent-clusterchecks-pdb.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.clusterChecksRunner.createPodDisruptionBudget -}}
+{{- if .Values.clusterChecksRunner.createPodDisruptionBudget .Values.clusterChecksRunner.pdb.enable -}}
 apiVersion: {{ template "policy.poddisruptionbudget.apiVersion" . }}
 kind: PodDisruptionBudget
 metadata:
@@ -7,7 +7,11 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
 spec:
-  maxUnavailable: 1
+  {{- if .Values.clusterChecksRunner.pdb.minAvailable -}}
+  minAvailable: {{ .Values.clusterChecksRunner.pdb.minAvailable }}
+  {{- else -}}
+  maxUnavailable: {{ .Values.clusterChecksRunner.pdb.maxUnavailable | default 1 }}
+  {{- end -}}
   selector:
     matchLabels:
       app: {{ template "datadog.fullname" . }}-clusterchecks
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/agent-clusterchecks-pdb.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.clusterChecksRunner.createPodDisruptionBudget -}}
+{{- if .Values.clusterChecksRunner.createPodDisruptionBudget .Values.clusterChecksRunner.pdb.enable -}}
```
LukasdeBoer: You're missing an `or` here.

houman-moallemi: Thanks Lukas! Fixed ❤️

tbavelier:
Thanks for the contribution @houman-moallemi , I'll be closing it in favour of https://github.com/DataDog/helm-charts/pull/2053 as I can't push to your branch

charts/datadog/templates/cluster-agent-pdb.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.clusterAgent.createPodDisruptionBudget -}}
+{{- if or .Values.clusterAgent.createPodDisruptionBudget .Values.clusterAgent.pdb.enable -}}
```
tbavelier, line 1: The values option is create, while the if block is enable

charts/datadog/templates/cluster-agent-pdb.yaml
```diff
@@ -7,7 +7,11 @@ metadata:
   labels:
 {{ include "datadog.labels" . | indent 4 }}
 spec:
-  minAvailable: 1
+  {{- if .Values.clusterChecksRunner.pdb.minAvailable -}}
+  minAvailable: {{ .Values.clusterChecksRunner.pdb.minAvailable }}
```
tbavelier, line 11: Typo: This should be values.clusterAgent not runners

houman-moallemi:
Thanks @tbavelier. Works for us.

PR #2028, opened 2025-09-04
[ACTP][PAR] allow to use a separate secret for credentials to match default expected values

dd-gplassard:

Introduces an option to create a separate kubernetes secret that will be mounted to `/etc/dd-action-runner/config/credentials` in order for connection files to be at their expected default location

The default value is `false` in order to not introduce a breaking change but the example uses `true` so new runners should default to the new behavior

Tested by
```bash
helm upgrade --install par  ./charts/private-action-runner -f ./charts/private-action-runner/examples/values.yaml
```
image

And 
```bash
 helm upgrade --install par  ./charts/private-action-runner -f ./charts/private-action-runner/examples/values.yaml --set-json 'runner.useSeparateSecretForCredentials=false'
```
image

#### Checklist

### DIFF
--- a/test/private-action-runner/__snapshot__/config-overrides.yaml
+++ b/test/private-action-runner/__snapshot__/config-overrides.yaml
@@ -78,7 +78,7 @@ metadata:
   name: custom-full-name
   namespace: datadog-agent
   labels:
-    helm.sh/chart: private-action-runner-1.7.0
+    helm.sh/chart: private-action-runner-1.8.0
     app.kubernetes.io/name: private-action-runner
     app.kubernetes.io/instance: override-test
     app.kubernetes.io/version: "v1.7.0"
@@ -93,7 +93,7 @@ spec:
   template:
     metadata:
       labels:
-        helm.sh/chart: private-action-runner-1.7.0
+        helm.sh/chart: private-action-runner-1.8.0
         app.kubernetes.io/name: private-action-runner
         app.kubernetes.io/instance: override-test
         app.kubernetes.io/version: "v1.7.0"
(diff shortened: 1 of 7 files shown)


test/private-action-runner/schema_validation_test.go
dd-gplassard, line 1: this is already broken on main, probably because of a dependency upgrade

the new error message is painful to check because all the attributes are on a single line in a random order

```

Expected :[]string{"- (root): Additional property credentialFiles is not allowed", "- (root): Additional property credentialSecrets is not allowed", "- (root): Additional property runners is not allowed", "Error: values don't meet the specifications of the sch ...

Actual   :[]string{"- at '': additional properties 'credentialSecrets', 'runners', 'credentialFiles' not allowed", "Error: values don't meet the specifications of the schema(s) in the following chart(s):", "private-action-runner:"}
```

So simplifying and just verifying we get a schema validation error

merchristK: maybe we should instead sort by the map keys first before building the list of unique lines.

merchristK (approved):

test/private-action-runner/schema_validation_test.go
dd-gplassard: I believe we would need to 
* find the right line from the stderr
* extract the list of additional properties 
* sort them and compare them to the expected list

Not that much but I wasn't sure it was super valuable. Happy to do it if you believe its worth it

merchristK: Nan it was just a suggestion. Current solution also sgtm

PR #2042, opened 2025-09-11
[CONTP-977] Bump Datadog CSI Driver chart dependency version.

adel121:

Bumps the version of csi chart dependency in datadog chart.  

The new version supports installation in gke autopilot.

#### Checklist

### DIFF
--- a/test/datadog-operator/baseline/DatadogAgent_CRD_default.yaml
+++ b/test/datadog-operator/baseline/DatadogAgent_CRD_default.yaml
@@ -7,7 +7,7 @@ metadata:
     controller-gen.kubebuilder.io/version: v0.16.3
   name: datadogagents.datadoghq.com
   labels:
-    helm.sh/chart: 'datadogCRDs-2.12.0-dev.2'
+    helm.sh/chart: 'datadogCRDs-2.12.0'
     app.kubernetes.io/managed-by: 'Helm'
     app.kubernetes.io/name: 'datadogCRDs'
     app.kubernetes.io/instance: 'datadog-operator'
--- a/test/datadog-operator/baseline/Operator_Deployment_default.yaml
+++ b/test/datadog-operator/baseline/Operator_Deployment_default.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: datadog-agent
   labels:
     app.kubernetes.io/name: datadog-operator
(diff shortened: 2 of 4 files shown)


charts/datadog/templates/_container-process-agent.yaml
```diff
@@ -90,11 +90,9 @@
       readOnly: true
     {{- end }}
     {{- if eq .Values.targetSystem "linux" }}
-    {{- if (not .Values.providers.gke.autopilot) }}
     - name: dsdsocket
       mountPath: {{ (dir .Values.datadog.dogstatsd.socketPath) }}
-      readOnly: false # Need RW for UDS DSD socket
```
adel121, line 96: RW access is needed to /var/run/datadog only if the container needs to create the socket. 

Mounting the volume in readonly mode does not hinder the container from writing to the socket file as long as it has write access to the file itself. 

I already validated it still works even if readonly is set to false (by works I mean: profiling of process agent works, traces are submit).

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -70,7 +70,7 @@
     type: File
   name: kubelet-ca
 {{- end }}
-{{- if eq (include "trace-agent-use-uds" .) "true" }}
+{{- if and (not .Values.providers.gke.autopilot) (eq (include "trace-agent-use-uds" .) "true") }}
```
adel121, line 73: On gke autopilot `dsdsocket` volume will be mounted, so this volume is not necessary as it is a duplicate and will cause violation of the gke allowlist synchronizer.

charts/datadog/values.yaml
```diff
@@ -543,7 +543,7 @@ datadog:
     socketPath: /var/run/datadog/apm.socket
 
     # datadog.apm.hostSocketPath -- Host path to the trace-agent socket
-    hostSocketPath: /var/run/datadog/
```
adel121, line 546: The extra forward slash causes a mismatch with gke autopilot allowlist synchronizer. They expect the path to have an exact string match.

charts/datadog/templates/_container-process-agent.yaml
kkhor-datadog: Would you mind adding some of this context as a code comment next to the `readOnly: true` config?

PR #2120, opened 2025-10-28
[OTAGENT-634] Support autoscaling with HPA in DDOT Gateway

songy23:
Support autoscaling using Horizontal Pod Autoscaler (HPA) in OTel Agent Gateway, configured by `otelAgentGateway.autoscaling`.

https://datadoghq.atlassian.net/browse/OTAGENT-634

#### Checklist

### DIFF
--- a/charts/datadog/templates/NOTES.txt
+++ b/charts/datadog/templates/NOTES.txt
@@ -187,16 +187,6 @@ The Datadog Agent is listening on port {{ $apmPort }} for APM service.
 
 {{- end }}
 
-{{- if and (semverCompare "<1.23.0" .Capabilities.KubeVersion.Version) (.Values.otelAgentGateway.autoscaling.enabled) }}
-
-###################################################################################
-####               ERROR: OTel Agent Gateway misconfiguration               ####
-###################################################################################
-
-{{- fail "OTel Agent Gateway autoscaling can only be enabled in Kubernetes version 1.23.0 and above" }}
-
-{{- end }}
-
 {{- if and .Values.datadog.apm.instrumentation.enabled (eq (include "cluster-agent-enabled" .) "false")}}
 
 #################################################################
@@ -763,5 +753,14 @@ You have set `datadog.processAgent.runInCoreAgent` to `false`.
 Support for this configuration will be deprecated in a future version.
 This configuration controlled whether the Process Agent or Core Agent runs the following features: Live Processes, Live Containers, Process Discovery.
 This behavior will be adjusted automatically and the Core Agent will be used by default on versions 7.60+.
+
+{{- end }}
+
+{{- if and (semverCompare "<1.23.0" .Capabilities.KubeVersion.Version) (.Values.otelAgentGateway.autoscaling.enabled) }}
+
+###################################################################################
+####               WARNING: OTel Agent Gateway misconfiguration               ####
+###################################################################################
+OTel Agent Gateway autoscaling has no effect in Kubernetes version 1.22.x and below
 
 {{- end }}


charts/datadog/templates/otel-agent-gateway-hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.otelAgentGateway.autoscaling.enabled }}
-apiVersion: autoscaling/v2
+apiVersion: autoscaling/v1
```
songy23: v2 will fail in k8s 1.25 and below , looks like chart/datadog needs to be backwards compatible?

truthbk: We need agent-onboarding to review this and confirm.

charts/datadog/ci/otel-agent-gateway-hpa-values.yaml
```diff
@@ -0,0 +1,51 @@
...
+    minReplicas: 2
+    maxReplicas: 10
+    metrics:
+      # Aim for high CPU utilization for higher throughput
+      - type: Resource
+        resource:
+          name: cpu
+          target:
+            type: Utilization
+            averageUtilization: 80
```
truthbk, line 46: I think we might want to look at our SMP tests to check memory evolution as a function of CPU under basic use-cases.

I believe throughout the PR we establish CPU-bound workloads in GW (which are the likely reality), but I'm wondering about the possibility of memory bound workloads too, in which case scaling on CPU and not memory might not be ideal.

charts/datadog/templates/otel-agent-gateway-hpa.yaml
vboulineau: It's available since 1.23, 10 versions ago, given that's a new feature, it's fine to request >= 1.23 for this feature.

khewonc: I think you can use [Capabilities.KubeVersion.Version](https://helm.sh/docs/chart_template_guide/builtin_objects/) in the if to check that the k8s version is compatible

charts/datadog/values.yaml
```diff
@@ -2911,3 +2912,39 @@ otelAgentGateway:
     # If true, checks OTel Collector config for filelog receiver and mounts additional volumes to collect containers
     # and pods logs.
     enabled: false
+
+  autoscaling:
```
khewonc, line 2917: Can you add descriptions for all the configurable fields here? Sorry as I know there are a lot of fields, but we use this info to populate the description of configurable fields in the readme

songy23: Sure, added for all

charts/datadog/templates/otel-agent-gateway-hpa.yaml
songy23: I ended up requiring >= k8s 1.23 instead of supporting older versions with autoscaling/v1, because the configs in autoscaling/v1 vs. autoscaling/v2 are different (autoscaling/v1 is much more limited).  While I can do a conditional check on which configs to support, I don't see an easy way to test different configs in different k8s versions with current CI setup.

E.g. if I want to test
```
    metrics:
      # Aim for high CPU utilization for higher throughput
      - type: Resource
        resource:
          name: cpu
          target:
            type: Utilization
            averageUtilization: 80
    behavior:
      scaleUp:
        stabilizationWindowSeconds: 30
      scaleDown:
        stabilizationWindowSeconds: 60
```
which is only supported in autoscaling/v2, it will break in 1.22.17 https://github.com/DataDog/helm-charts/blob/1b0810c8eaa40c42fbbf87afc63379d5f3a74b24/.github/workflows/ci.yaml#L126

charts/datadog/ci/otel-agent-gateway-hpa-values.yaml
songy23: To clarify, these are just examples rather than default. We don't have any pre-defined metrics or scaling behavior in HPA, and customers need to provide them. I updated the docs to make this clear.

We can definitely provide some recommendations on HPA metrics and/or scaling behavior from SMP. It can be CPU, memory or external (custom) metrics. But IMO it doesn't need to be in this PR (in fact I think we need this PR to go in first to unblock it).

songy23:
CIs failed because http://get.helm.sh/ is unreachable  (not related to the PR)

charts/datadog/templates/otel-agent-gateway-hpa.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if and (semverCompare ">=1.23.0" .Capabilities.KubeVersion.Version) (.Values.otelAgentGateway.autoscaling.enabled) }}
```
levan-m: Perhaps worth adding a [NOTES.txt](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/NOTES.txt) warning or error if autoscaling is enabled on k8s <1.23.

levan-m: Also, I suppose this should be conditionally enabled only when `otelAgentGateway.enabled: true`.

levan-m (approved):

Issue #1979, opened 2025-07-28
Default chart settings cause Argo-CD to constantly be out of sync due to readOnly: false


jwhittem-usb:
Below are listed the volume mounts that have an issue syncing with the default state of the resource which is blank (assumed readOnly: false).  After applying Argo will stay out of sync trying to "correct" the diff with these "readOnly: false" options, but since it's the default state the live object will not contain readOnly: false.

I downloaded the chart, removed these options and the chart seemed to install just the same, so I'm not sure why they are even in the chart?  If there is a reason to keep them LMK, I can propose a PR to make it a helm chart value, instead of hardcoded into the chart.

_suggest adding a helm chart default variable, and code for `readOnly: false`, so it can be turned off via values.yaml, or if not needed from the default daemonset manifest, removed entirely._

Thank you.

-- snip --

<img width="1403" height="656" alt="Image" src="https://github.com/user-attachments/assets/e304dabb-6756-48d8-8df4-d01b9db11ab0" />

- mountPath: /var/log/datadog
name: logdatadog
readOnly: false

- mountPath: /tmp
name: tmpdir
readOnly: false

- mountPath: /etc/datadog-agent
name: config
readOnly: false

- mountPath: /etc/datadog-agent/auth
name: auth-token
readOnly: false

- mountPath: /var/run/datadog
name: dsdsocket
readOnly: false

- mountPath: /var/log/datadog
name: logdatadog
readOnly: false

- mountPath: /tmp
name: tmpdir
readOnly: false

- mountPath: /var/run/datadog
name: dsdsocket
readOnly: false

- mountPath: /var/log/datadog
name: logdatadog
readOnly: false

- mountPath: /var/run/datadog
name: dsdsocket
readOnly: false

- mountPath: /tmp
name: tmpdir
readOnly: false

- mountPath: /var/log/datadog
name: logdatadog
readOnly: false

- mountPath: /tmp
name: tmpdir
readOnly: false

- mountPath: /sys/kernel/debug
mountPropagation: None
name: debugfs
readOnly: false

- mountPath: /var/run/sysprobe
name: sysprobe-socket-dir
readOnly: false

- mountPath: /host/proc
mountPropagation: None
mountPropagation: None
name: runtime-compiler-output-dir
readOnly: false

- mountPath: /var/tmp/datadog-agent/system-probe/kernel-headers
name: kernel-headers-download-dir
readOnly: false

- mountPath: /var/log/datadog
name: logdatadog
[...]

PR #2150, opened 2025-11-07
[CONTP-1089] Enable readOnlyRootFilesystem on all agent containers

Mathew-Estafanous:
### What this PR does / why we need it:

This PR sets `securityContext.readOnlyRootFilesystem: true` for most Datadog Agent containers to improve support for read-only filesystem (ROFS) environments and strengthen the Agent’s overall security posture. This is a quick win for Datadog and our customers because the Agent is already mostly configured to support ROFS with minimal effort.

Agent containers with `readOnlyRootFilesystem` newly set to true:
- agent
- process agent
- otel agent
- trace agent
- system probe
- security agent
- agent data plane
- cluster checks

#### Additional Changes
- Added writable volume mounts to agents that are missing them (`system-probe` and `cluster-checks-runner`)
- Verified ADP writes are the same as other Go-based agents ([Slack confirmation](https://dd.slack.com/archives/C08A7KEN7LL/p1762794444315989))

#### Manual QA 🧪

1. Deployed all agent containers via the helm chart
```yml
datadog:
  clusterName: rofs-helm-dev
  kubelet:
    tlsVerify: false
  logs:
    enabled: false
    containerCollectAll: true
  dogstatsd:
    nonLocalTraffic: true
    originDetection: true
    useSocketVolume: true
    tagCardinality: "high"
  kubeStateMetricsCore:
    enabled: true
  admissionController:
    enabled: true
    mutateUnlabelled: true
  apm:
    socketEnabled: true
    portEnabled: true
  processAgent:
    enabled: true
    processCollection: true
    containerCollection: true
  networkMonitoring:
    enabled: true
  serviceMonitoring:
    enabled: true
  systemProbe:
    enableConntrack: true
    collectDNSStats: true
    enableTCPQueueLength: true
    enableOOMKill: true
  securityAgent:
    compliance:
      enabled: true
    runtime:
      enabled: true
      fimEnabled: true
      network:
        enabled: true
  agentDataPlane:
    enabled: true
  otelCollector:
    enabled: true
agents:
    doNotCheckTag: true
clusterAgent:
  sharedProcessNamespace: false
  admissionController:
    mutateUnlabelled: true
    enabled: true
clusterChecksRunner:
  enabled: true
  doNotCheckTag: true
```

2. Installed the updated helm chart 
```
helm install datadog-agent-helm ./charts/datadog \
--values ./charts/datadog/testing/datadog.yml \
--set datadog.apiKey="$DD_API_KEY" \
--set datadog.appKey="$DD_APP_KEY" \
--namespace datadog-agent
```

3. Validate containers running as expected.
<details><summary>Details</summary>
<p>

image

Screenshot 2025-11-11 at 9 42 10 AM

image

</p>
</details> 

#### Smoke test on k8s-dd-agent ☁️
- Replicated changes in https://github.com/DataDog/k8s-datadog-agent-ops/pull/6740
- Deployed to `entei` staging cluster
- Observed no anomalies or regressions after deployment

### Risks & Considerations

Enabling ROFS on all agent containers has a known level of risk with the worst case being the agent failing to start. There could also be some configuration changing where the agent writes and possibly hitting a read-only filesystem error if a writable volume isn't present.

The following steps were taking to mitigate risk:
- Locally all agents to confirm they continue to work.
- Smoke tested a similar change on an internal staging cluster (See https://github.com/DataDog/k8s-datadog-agent-ops/pull/6740)
- Added missing writable volume mounts (CLC runner & system-probe)
- Enumerated all known agent filesystem writes ([see this RFC](https://datadoghq.atlassian.net/wiki/spaces/CONTP/pages/5637079172/Support+Read-Only+Root+Filesystems+for+Datadog+Agent#%F0%9F%91%A3--Agent-Filesystem-Footprint-(source)))
- Ran [E2E tests](https://gitlab.ddbuild.io/DataDog/helm-charts/-/pipelines/83189128) for GKE and GKE autopilot 

### Checklist

### DIFF
--- a/charts/datadog/templates/_container-security-agent.yaml
+++ b/charts/datadog/templates/_container-security-agent.yaml
@@ -6,16 +6,11 @@
   lifecycle:
 {{ toYaml .Values.agents.lifecycle | indent 4 }}
   {{- end }}
-  {{- if or .Values.agents.containers.securityAgent.securityContext (eq  (include "should-enable-compliance" .) "true") }}
-  securityContext:
-    {{- if eq  (include "should-enable-compliance" .) "true" }}
-    capabilities:
-      add: ["AUDIT_CONTROL", "AUDIT_READ"]
-    {{- end }}
-    {{- if .Values.agents.containers.securityAgent.securityContext }}
-{{ toYaml .Values.agents.containers.securityAgent.securityContext | indent 4 }}
-    {{- end }}
-  {{- end }}
+  {{- $securityContext := .Values.agents.containers.securityAgent.securityContext -}}
+  {{- if eq (include "should-enable-compliance" .) "true" -}}
+    {{- $securityContext = merge $securityContext (dict "capabilities" (dict "add" (list "AUDIT_CONTROL" "AUDIT_READ"))) -}}
+  {{- end -}}
+{{ include "generate-security-context" (dict "securityContext" $securityContext "targetSystem" .Values.targetSystem "seccomp" "" "kubeversion" .Capabilities.KubeVersion.Version) | nindent 2 }}
   command: ["security-agent", "start", "-c={{ template "datadog.confPath" . }}/datadog.yaml"]
   resources:
 {{ toYaml .Values.agents.containers.securityAgent.resources | indent 4 }}
--- a/charts/datadog/templates/_container-agent-data-plane.yaml
+++ b/charts/datadog/templates/_container-agent-data-plane.yaml
@@ -3,6 +3,7 @@
   image: "{{ include "image-path" (dict "root" .Values "image" .Values.datadog.agentDataPlane.image) }}"
   imagePullPolicy: {{ .Values.datadog.agentDataPlane.image.pullPolicy }}
   command: ["agent-data-plane", "run", "--config={{ template "datadog.confPath" . }}/datadog.yaml"]
+{{ include "generate-security-context" (dict "securityContext" .Values.agents.containers.agentDataPlane.securityContext "targetSystem" .Values.targetSystem "seccomp" "" "kubeversion" .Capabilities.KubeVersion.Version) | indent 2 }}
   resources:
 {{- if and (empty .Values.agents.containers.agentDataPlane.resources) .Values.providers.gke.autopilot -}}
 {{ include "default-container-resources" . | indent 4 }}
--- a/test/datadog/baseline/manifests/adp_enabled.yaml
+++ b/test/datadog/baseline/manifests/adp_enabled.yaml
@@ -1147,6 +1147,8 @@ spec:
             successThreshold: 1
             timeoutSeconds: 5
           resources: {}
+          securityContext:
+            readOnlyRootFilesystem: true
           volumeMounts:
             - mountPath: /tmp
               name: tmpdir


charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -70,6 +70,11 @@
       subPath: datadog.yaml
       readOnly: true
     {{- end }}
+    {{- if eq .Values.targetSystem "linux" }}
+    - name: eventsocket
+      mountPath: /opt/datadog-agent/run
+      readOnly: false
```
brycekahle: I don't think `readOnly: false` is needed since it is the default value, and it may cause issues including it: See #1979

charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -74,8 +74,8 @@
       mountPath: /etc/datadog-agent/system-probe.yaml
       subPath: system-probe.yaml
       readOnly: true
-    - name: sysprobe-socket-dir
-      mountPath: /var/run/sysprobe
+    - name: sysprobe-run-dir
+      mountPath: /opt/datadog-agent/run
```
brycekahle:
```suggestion
      mountPath: /opt/datadog-agent/run/sysprobe
```

Mathew-Estafanous: `/opt/datadog-agent/run/event-monitor.sock` and `/opt/datadog-agent/run/runtime-security` are written by the agent hence the `/opt/datadog-agent/run` mountPath.

brycekahle: The config is read by both the agent and system-probe, so the paths need to match between the two. Since we aren't using defaults, if you wanted to revert back to `/var/run/sysprobe` to minimize changes, I would be OK with that.

Mathew-Estafanous: I'm using a subpath of the sysprobe volume when mounting it on the other agents since they look at the sys probe config `/opt/datadog-agent/run/sysprobe/`

https://github.com/DataDog/helm-charts/blob/09ffb46a4b8fde2e48da8115fadd6dfa455cd7c9/charts/datadog/templates/_container-agent.yaml#L321-L324

I'm open to reverting back to simplify this PR as well.

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -77,6 +77,8 @@
   name: apmsocket
 {{- end }}
 {{- if eq (include "should-enable-system-probe" .) "true" }}
+- name: eventsocket
```
brycekahle: What are the other container(s) that need to mount this to use the socket(s) and file(s) within?

Mathew-Estafanous: The process agent needs a mount to `/opt/datadog-agent/run/event-monitor.sock`. I didn't add it because I wasn't sure if it was left out intentionally. I'm guessing it might be an oversight because without a mount, the process agent cannot collect events (ie. `process-agent events pull` fails)

brycekahle: @wiyu thoughts on the missing mount?

Mathew-Estafanous: As discussed, the `process-agent events pull/listen` command is mostly a prototype cmd and it's fine that it doesn't work. Cloud network monitoring does use the same socket for `ProcessEventsCheck` and I've reached out to them regarding this. I'll add the socket as a mount to process-agent if they deem it as a misconfiguration.

brycekahle (approved):

charts/datadog/values.yaml
```diff
@@ -2234,6 +2243,10 @@ agents:
         successThreshold: 1
         failureThreshold: 12
 
+      # agents.containers.agentDataPlane.securityContext -- Allows you to overwrite the default container SecurityContext for the agent-data-plane container.
+      securityContext:
+        readOnlyRootFilesystem: true
```
swang392, line 2248: `agentDataPlane.securityContext` isn't processed anywhere, should that be added to `_container-agent-data-plane.yaml`?

charts/datadog/templates/_container-security-agent.yaml
```diff
@@ -6,10 +6,15 @@
   lifecycle:
 {{ toYaml .Values.agents.lifecycle | indent 4 }}
   {{- end }}
-  {{- if eq  (include "should-enable-compliance" .) "true" }}
+  {{- if or .Values.agents.containers.securityAgent.securityContext (eq  (include "should-enable-compliance" .) "true") }}
   securityContext:
+    {{- if eq  (include "should-enable-compliance" .) "true" }}
     capabilities:
```
swang392: If a user specifies `securityAgent.securityContext.capabilities`, it will conflict with the `capabilities` specified here and override it. Is there a reason the `generate-security-context` helper isn't used here?

brycekahle (approved):

charts/datadog/values.yaml
Mathew-Estafanous: Ahh I didn't notice that. Good catch, thank you.

charts/datadog/templates/_container-security-agent.yaml
Mathew-Estafanous: Hmm good point. I honestly wasn't sure why the security agent doesn't use `generate-security-context`. It actually didn't use `securityAgent.securityContext` at all. Maybe that was overlooked?

PR #2169, opened 2025-11-14
feat: add ability to disable configSync

### LINKED ISSUE #2133: Not able to disable configSync in otel-agent
Hello Datadog team,
Currently, if you enable ddot in helm chart, it automatically configures configSync for you.
https://github.com/DataDog/helm-charts/blob/be378fa68489230924c2f4118eed2a20da4c8937/charts/datadog/templates/_container-otel-agent.yaml#L70-L73
While it is a great feature, I faced some cases where I would like to disable it.
In one of those cases, I want otel-agent to ingest metrics/spans directly to Datadog intake, while agent itself to send them to another otel collector for additional filtering/data masking before ingesting into Datadog via intake endpoints. To do it, I override DD_DD_URL and DD_APM_DD_URL in the agent container, but due to configSync, the otel-agent uses the same endpoints.
Based on the source code, to disable it, I should configure the port or interval to 0 - https://github.com/DataDog/datadog-agent/blob/fe3c3448775c167ca73548088526a7af36c3aaf7/comp/core/configsync/configsyncimpl/module.go#L69C1-L72
But they are hardcoded in the Helm chart.

I think the easiest solution will be to allow overriding of DD_AGENT_IPC_PORT, so you could set it to 0 for otel-agent container.
I can bring a PR if you are okay with adding this feature.
Thank you!

tr3mor:
Currently, if you enable ddot in helm chart, it automatically configures configSync for you. While it is a great feature, I faced some cases where I would like to disable it.
This pr adds config option to disable it
fixes #2133

#### Checklist

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.146.0
+
+* Add ability to disable configSync between otel-agent and agent container via `datadog.otelCollector.enableConfigSync`. Default is true to match previous behaviour.
+
 ## 3.145.1
 
 * [CONS-7793] Add necessary RBAC for ArgoRollout to be provide read access to the admission controller.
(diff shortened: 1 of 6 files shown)


tr3mor:
We found another way to achieve this, not needed anymore

PR #2178, opened 2025-11-18
Truncate part-of label

fanny-jiang:

Truncates part-of label so that they are under 63 characters in length as allowable by kubernetes. 

  - fixes #

#### Checklist


charts/datadog/templates/_helpers.tpl
```diff
@@ -784,7 +784,7 @@ Build part-of label
 */}}
 {{- define "part-of-label" -}}
 {{- $ns := .Release.Namespace | replace "-" "--" -}}
-{{- $name := include "datadog.fullname" . | replace "-" "--" -}}
+{{- $name := include "datadog.fullname" . | trunc 40 | replace "-" "--" | trimSuffix "-" -}}
```
swang392: What happens if the namespace variable is too long? If `ns` is 30+ characters, the final part-of label would still exceed 63 chars.

fanny-jiang: addressed in https://github.com/DataDog/helm-charts/pull/2178/commits/7fe497f361cd7455a2ae2e0cf121baf233d90e78

charts/datadog/templates/_helpers.tpl
```diff
@@ -784,8 +784,8 @@ Build part-of label
 */}}
 {{- define "part-of-label" -}}
 {{- $ns := .Release.Namespace | replace "-" "--" -}}
-{{- $name := include "datadog.fullname" . | replace "-" "--" -}}
-{{ printf "%s-%s" $ns $name }}
+{{- $name := include "datadog.fullname" . | replace "-" "--" | trimSuffix "-" -}}
+{{ printf "%s-%s" $ns $name | trunc 63 | trimSuffix "-" }}
```
swang392, line 788:
```suggestion
{{ printf "%s-%s" $ns $name | trunc 63}}
```
`trimSuffix` is already added to `$name` so I don't think it's needed again

fanny-jiang: A trailing `-` could still be present if the value gets truncated at a `-` char.

swang392: oooh true, nevermind

PR #2182, opened 2025-11-19
Add e2e test for CSI on GKE Autopilot

JatinSingh007:

Adds end-to-end test to validate that the datadog-csi-driver chart can be successfully installed and run on GKE Autopilot with the existing allowlist.

  - fixes # NA

#### Checklist


adel121:
@codex review

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+		var csiDriverPod corev1.Pod
+		containsCsiDriver := false
+		for _, pod := range res.Items {
+			if strings.Contains(pod.Name, "csi-driver") {
+				containsCsiDriver = true
+				csiDriverPod = pod
+				break
+			}
+		}
```
adel121: This logic can be simplified by using kubernetes label selectors so that you retrieve pods that belong to the csi driver daemonset set.

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	output, err = helmCmd.CombinedOutput()
+	v.T().Logf("Helm output: %s", string(output))
+	if err != nil {
+		v.T().Fatalf("Helm install failed: %v", err)
+	}
+	v.T().Log("CSI driver installed")
+
+	// Wait for CSI driver pod to transition from Pending to Running state
+	time.Sleep(5 * time.Second)
```
adel121: No need to Sleep here because we have an EventuallyWithTf directly after it.

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: Yes you're right just for consistency I used this, other tests more or less follow the same flow

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: Actually I thought that too but the tests immediately exited when it saw it in the pending state

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// Check if CSI driver pods exist
+	assert.EventuallyWithTf(v.T(), func(c *assert.CollectT) {
+		res, err := v.Env().KubernetesCluster.Client().CoreV1().Pods("datadog-agent").List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
```
Mathew-Estafanous: Consider using an `require.NoError(c, err)` here because this `EventuallyWithTf` will pass when `err != nil` because technically no assertion failed.

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
```diff
@@ -0,0 +1,108 @@
...
+	output, err := helmCmd.CombinedOutput()
+	v.T().Logf("Helm output: %s", string(output))
+	if err != nil {
+		v.T().Fatalf("Helm repo add failed: %v", err)
+	}
+	v.T().Log("Datadog repository added")
+
+	// Installing the csi driver via helm
+	v.T().Log("Installing CSI driver")
+	helmCmd = exec.Command("helm", "install", "datadog-csi-driver", "datadog/datadog-csi-driver",
```
Mathew-Estafanous: test infra definitions has a [csidriver package](https://github.com/DataDog/test-infra-definitions/blob/main/components/datadog/csi-driver/helm.go#L29C6-L29C25) to install the driver. Can that be used instead?

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
```diff
@@ -0,0 +1,108 @@
...
+		containsCsiDriver := false
+		for _, pod := range res.Items {
+			if strings.Contains(pod.Name, "csi-driver") {
+				containsCsiDriver = true
+				csiDriverPod = pod
+				break
+			}
+		}
+
+		assert.True(v.T(), containsCsiDriver, "CSI Driver pod not found")
```
Mathew-Estafanous:
```suggestion
		assert.True(c, containsCsiDriver, "CSI Driver pod not found")
```
You need to use the provided `(c *assert.CollectT)` passed into the eventually with condition. I believe if you use `v.T()` then the eventually with won't keep on retrying, even if the assertion failed, so your test might fail prematurely without waiting for the timeout.

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: Yes I tried using that with `workloadappfunc` the problem with that is on Autopilot allowlist takes sometime to sync and if while provisioning you try to install, it fails

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: Yes I realised that good catch thanks, it will also solve the problem of time.sleep()

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
swang392: follow up: this installs the latest public version of the `datadog-csi-driver` chart, which won't include any edits made on a local branch which is what we want to test with e2e. Instead, when setting up `e2e.Run` in `TestGKEAutopilotCSISuite`, add something similar to below (example from `e2e_gke_autpilot_test.go`):

```
e2e.Run(t, &gkeAutopilotSuite{}, e2e.WithProvisioner(gcpkubernetes.GKEProvisioner(
		....
		gcpkubernetes.WithAgentOptions(
			kubernetesagentparams.WithGKEAutopilot(),i
			kubernetesagentparams.WithHelmRepoURL(""),
			kubernetesagentparams.WithHelmChartPath(datadogChartPath()),
			kubernetesagentparams.WithHelmValues(`
            ...
```

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: Ahhh yes you're right, I missed assertion here and returning early, we can put an `!assert.NoError` and return or use `require` as you mentioned. Thanks !!

test/e2e/datadog/e2e_gke_autopilot_csi_test.go
JatinSingh007: I considered this, if I understand it correctly those parameters only work for installing the main Datadog Agent chart. This test specifically validates the standalone `datadog-csi-driver` chart installation. Using the provisioner would install `datadog/datadog` with CSI enabled as a subchart, which is a different deployment model than what we're testing here.

swang392: Yes, that code snippet installs the `datadog` chart, but you could create a similar method to install the `datadog-csi-driver` chart and pass that into `WithHelmChartPath()`. Otherwise, if the e2e test runs on a branch where the `datadog-csi-driver` chart was updated, we wouldn't be testing for any possible regressions from that branch, it would just be testing the latest chart on main.

PR #2190, opened 2025-11-20
Update CloudPrem Helm chart

guilload:
Release new version

### DIFF
--- a/charts/cloudprem/values.yaml
+++ b/charts/cloudprem/values.yaml
@@ -62,7 +62,8 @@ image:
   # The CloudPrem image is also available on DockerHub:
   # https://hub.docker.com/r/datadog/cloudprem
   repository: public.ecr.aws/datadog/cloudprem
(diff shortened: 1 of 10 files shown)


charts/cloudprem/values.yaml
```diff
@@ -303,11 +314,11 @@ indexer:
   resources:
     # See https://docs.datadoghq.com/cloudprem/configure/cluster_sizing/
     limits:
-      cpu: "4"
-      memory: "8Gi"
+      cpu: 2
```
PSeitz-dd, line 317: only 2 cpus?

PSeitz-dd (approved):

charts/cloudprem/values.yaml
guilload: Yeah, maybe we should go to 4 / 16 vCPU right away. Let's do that in the next release, next week.

PR #2198, opened 2025-11-26
[ACTP] Configurable Security Context Constraints for OpenShift

merchristK:
## What this PR does / why we need it:
When running the PAR using OpenShift, there is an error at startup because by default each container is executed with a random user ID in the range `[1000660000, 1000669999]`. This is problematic because only the `dog(uid=501)` (and root) has write permission to `/etc/dd-action-runner` directory.

```bash
> ls -la /etc/dd-action-runner/
total 0
drwx------. 1 dog  dog   20 Nov 26 21:38 .
drwxr-xr-x. 1 root root  42 Nov 26 21:38 ..
drwxrwxrwt. 3 root root 100 Nov 26 21:38 config
```

So to be able to run this successfully, we need a [security context constraints](https://docs.redhat.com/en/documentation/openshift_container_platform/4.8/html/authentication_and_authorization/managing-pod-security-policies) that starts the pod with the non root user from the PAR docker image (`dog`)

## Testing
I've created a local oc cluster following this [wiki](https://datadoghq.atlassian.net/wiki/spaces/TS/pages/2520056015/Creating+OpenShift+Environment+With+Code-Ready-Containers) and then applied the chart
```bash
helm install oc-par ~/dd/helm-charts/charts/private-action-runner -f values.yaml
```

Using this values.yaml file
```yaml
runner:
    podSecurity:
        securityContextConstraints:
            create: true
```

### DIFF
--- a/charts/private-action-runner/README.md
+++ b/charts/private-action-runner/README.md
@@ -129,7 +129,6 @@ helm upgrade <RELEASE_NAME> datadog/private-action-runner -f values.yaml
 
 ## OpenShift Deployment
 
-
 ### Enabling Security Context Constraints
 
 When deploying on OpenShift, you need to configure [Security Context Constraints](https://docs.redhat.com/en/documentation/openshift_container_platform/3.11/html/cluster_administration/admin-guide-manage-scc) (SCC) to ensure the runner pods have the necessary permissions to function properly.
@@ -152,7 +151,7 @@ runner:
   config:
     urn: "YOUR_RUNNER_URN"
     privateKey: "YOUR_RUNNER_PRIVATE_KEY"
-  
+
   # Enable SCC creation for OpenShift
   podSecurity:
     securityContextConstraints:
--- a/charts/private-action-runner/README.md.gotmpl
+++ b/charts/private-action-runner/README.md.gotmpl
@@ -128,6 +128,77 @@ helm upgrade <RELEASE_NAME> datadog/private-action-runner -f values.yaml
 * Deploy several runners with different permissions for different teams or environments
 * Learn more about [Private actions](https://docs.datadoghq.com/actions/private_actions/)
 
+## OpenShift Deployment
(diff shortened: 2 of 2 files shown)


merchristK:
* **#2198** <a href="https://app.graphite.com/github/pr/DataDog/helm-charts/2198?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/helm-charts/2198?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

charts/private-action-runner/templates/scc.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.runner.podSecurity.securityContextConstraints.create }}
+kind: SecurityContextConstraints
+apiVersion: security.openshift.io/v1
+metadata:
+  name: {{ include "chart.fullname" . }}
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+users:
+- system:serviceaccount:{{ .Release.Namespace }}:{{ include "chart.fullname" . }}
+priority: 10 # same priority as anyuid SCC
```
merchristK, line 10: I've tried setting this to `null` but it didn't work... When I did the default behavior was still applied. The openshift client used `restricted-v2` instead of picking this scc. The `anyuid` scc has a priority of `10` so I've tried using the same.

charts/private-action-runner/README.md
dd-gplassard, line 1: can you add readme for instructions on the changes required to deploy in openshift ? 🙏

PR #2226, opened 2025-12-08
obs-pipelines: Add support for secrets-management

graphcareful:
- Exposes new options for the user to pass a custom secrets config or easy setup via a secrets file.
- If the user selects secrets file it will be created, formatted correct, filled with values from the `entries` and the OPW bootstrap file will correctly point to it.

- fixes OPA-4166

#### Checklist

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Datadog changelog
 
+## 3.154.0
+
+* Add a field to enable the kubelet orchestrator check
+
+
 ## 3.153.0
 
 * Add support for App & API Protection injector configuration for proxies (Envoy Gateway, Istio) via `datadog.appsec.injector` settings.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: datadog
-version: 3.153.0
+version: 3.154.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.153.0](https://img.shields.io/badge/Version-3.153.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.154.0](https://img.shields.io/badge/Version-3.154.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 [Datadog](https://www.datadoghq.com/) is a hosted infrastructure monitoring platform. This chart adds the Datadog Agent to all nodes in your cluster via a DaemonSet. It also optionally depends on the [kube-state-metrics chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-state-metrics). For more information about monitoring Kubernetes with Datadog, please refer to the [Datadog documentation website](https://docs.datadoghq.com/agent/basic_agent_usage/kubernetes/).
 
@@ -860,6 +860,7 @@ helm install <RELEASE_NAME> \
 | datadog.orchestratorExplorer.container_scrubbing | object | `{"enabled":true}` | Enable the scrubbing of containers in the kubernetes resource YAML for sensitive information |
 | datadog.orchestratorExplorer.customResources | list | `[]` | Defines custom resources for the orchestrator explorer to collect |
 | datadog.orchestratorExplorer.enabled | bool | `true` | Set this to false to disable the orchestrator explorer |
+| datadog.orchestratorExplorer.kubelet_configuration_check.enabled | bool | `true` | Enable the orchestrator kubelet configuration check |
 | datadog.originDetectionUnified.enabled | bool | `false` | Enabled enables unified mechanism for origin detection. Default: false. (Requires Agent 7.54.0+). |
 | datadog.osReleasePath | string | `"/etc/os-release"` | Specify the path to your os-release file |
 | datadog.otelCollector.config | string | `nil` | OTel collector configuration |
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -267,6 +267,10 @@
     {{- end }}
     {{- include "additional-env-entries" .Values.agents.containers.agent.env | indent 4 }}
     {{- include "additional-env-dict-entries" .Values.agents.containers.agent.envDict | indent 4 }}
+    {{- if .Values.datadog.orchestratorExplorer.kubelet_configuration_check.enabled }}
+    - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+      value: "true"
+    {{- end }}
   volumeMounts:
     - name: logdatadog
       mountPath: {{ template "datadog.logDirectoryPath" . }}
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -942,6 +942,12 @@ datadog:
     container_scrubbing:
       enabled: true
 
+    # datadog.orchestratorExplorer.kubelet_configuration_check.enabled -- Enable the orchestrator kubelet configuration check
+
+    ## this enables the collection of the kubelet configuration for viewing in the orchestrator
+    kubelet_configuration_check:
+      enabled: true
+
     # datadog.orchestratorExplorer.customResources -- Defines custom resources for the orchestrator explorer to collect
 
     # customResources is required for RBAC creation if a custom orchestrator explorer configuration is provided in `clusterAgent.confd` or `clusterAgent.advancedConfd`
--- a/test/datadog/baseline/manifests/adp_enabled.yaml
+++ b/test/datadog/baseline/manifests/adp_enabled.yaml
@@ -937,6 +937,8 @@ spec:
               value: "true"
             - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
               value: /var/lib/kubelet/pod-resources/kubelet.sock
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/agent-clusterchecks-deployment_default.yaml
+++ b/test/datadog/baseline/manifests/agent-clusterchecks-deployment_default.yaml
@@ -966,6 +966,8 @@ spec:
               value: "true"
             - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
               value: /var/lib/kubelet/pod-resources/kubelet.sock
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/agent-workload_exclude.yaml
+++ b/test/datadog/baseline/manifests/agent-workload_exclude.yaml
@@ -933,6 +933,8 @@ spec:
               value: "true"
             - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
               value: /var/lib/kubelet/pod-resources/kubelet.sock
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/daemonset_default.yaml
+++ b/test/datadog/baseline/manifests/daemonset_default.yaml
@@ -931,6 +931,8 @@ spec:
               value: "true"
             - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
               value: /var/lib/kubelet/pod-resources/kubelet.sock
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/default_all.yaml
+++ b/test/datadog/baseline/manifests/default_all.yaml
@@ -931,6 +931,8 @@ spec:
               value: "true"
             - name: DD_KUBERNETES_KUBELET_PODRESOURCES_SOCKET
               value: /var/lib/kubelet/pod-resources/kubelet.sock
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/default_all_windows.yaml
+++ b/test/datadog/baseline/manifests/default_all_windows.yaml
@@ -908,6 +908,8 @@ spec:
               value: /var/lib/kubelet/pod-resources/kubelet.sock
             - name: DD_SYSTEM_PROBE_ENABLED
               value: "false"
+            - name: DD_ORCHESTRATOR_EXPLORER_KUBELET_CONFIG_CHECK_ENABLED
+              value: "true"
           image: gcr.io/datadoghq/agent:7.73.0
           imagePullPolicy: IfNotPresent
           livenessProbe:
--- a/test/datadog/baseline/manifests/gdc_daemonset_default.yaml
+++ b/test/datadog/baseline/manifests/gdc_daemonset_default.yaml
@@ -911,6 +911,8 @@ spec:
               value: "true"
             - name: DD_KUBELET_CORE_CHECK_ENABLED
               value: "true"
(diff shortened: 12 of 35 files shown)


charts/observability-pipelines-worker/values.yaml
```diff
@@ -39,6 +39,25 @@ datadog:
...
+  secretFile:
+    enabled: false
+
+    # where to mount secrets.json inside the container
+    mountPath: /etc/observability-pipelines-secrets/secrets.json
+
+    # Build secrets.json from entries
+    entries: {}
+
+  custom:
```
emilychendd: hmm it's a little confusing to me to have a custom vs secretFile section. I wonder if below would be more consistent. 
 
```
#values.yaml
#Comment references the public doc we'll have for secrets backend integration.
secret:
  sgc_path: /opt/datadog/observability-pipelines-worker/bin/datadog-secret-manager
  backend_type: json
  backend_config:
    file_path: /etc/observability-pipelines-secrets/secrets.json

#another section to help populate secrets.json, potentially secrets.yaml also.
secretJson:
  mountPath: /etc/observability-pipelines-secrets/secrets.json
  entries: {}
```

charts/observability-pipelines-worker/values.yaml
```diff
@@ -39,6 +39,25 @@ datadog:
     # if you change this port, you'll need to update the livenessProbe and readinessProbe
     address: "0.0.0.0:8686"
 
+bootstrapBackend:
+  filename: bootstrap.yaml
+
+  # exactly one of these should be true
+  secretFile:
+    enabled: false
```
emilychendd: Do we need to have enabled? can we just check the existence of the block?

emilychendd: actually i see other sections have this too... so it's consistent

charts/observability-pipelines-worker/values.yaml
```diff
@@ -39,6 +39,25 @@ datadog:
     # if you change this port, you'll need to update the livenessProbe and readinessProbe
     address: "0.0.0.0:8686"
 
+bootstrapBackend:
+  filename: bootstrap.yaml
```
gwenaskell: Just to be sure, is this useful that the user can configure the name of the bootstrap file mounted on the worker? Same for the mount path? If not, I'd rather hardcode them in the templates

charts/observability-pipelines-worker/values.yaml
gwenaskell: Agreed, it looks like those could be merged. I would suggest something like this:

```yaml
bootstrapBackend:
  enabled: false

  filename: bootstrap.yaml # if this is necessary

  sgc_path: /opt/datadog/observability-pipelines-worker/bin/datadog-secret-manager

  backend_type: json

  backend_config:
    file_path: /etc/observability-pipelines-secrets/secrets.json

  # used when the backend_type is 'json'
  secretJson:
    # uncomment this to create and mount a Secret containing the provided entries in the worker
    # entries: {}
```

To make this safer:

- you can add validation rules at the beginning of the secret template like so:

```yaml
{{ if and Values.bootstrapBackend.enabled Values.bootstrapBackend.secretJson.entries }}

{{- if ne Values.bootstrapBackend.backend_type "json" }}
{{ fail "bootstrapBackend type should be 'json' to use secretJson.entries" }}
{{ end -}}

{{- if ne Values.bootstrapBackend.backend_config.file_path "/etc/observability-pipelines-secrets/secrets.json" }}
{{ fail "the secrets backend file_path cannot be changed when type is 'json'" }}
{{ end -}}

// secret goes here
{{ end }}
```


Would that make sense to you?

graphcareful: The idea was so that people would never have to touch `file_path` if backend type was `json`, because then we need to ensure the paths line up correctly. So the options are either configure via secrets file or not. Thoughts?

emilychendd: Just noting that we chatted offline and it is more convenient to not have to align the file_path field when using files as backend, although it makes this config for file a little different with other platforms. No strong preference on my side.

charts/observability-pipelines-worker/values.yaml
```diff
@@ -39,6 +39,25 @@ datadog:
...
+
+  # exactly one of these should be true
+  secretFile:
+    enabled: false
+
+    # where to mount secrets.json inside the container
+    mountPath: /etc/observability-pipelines-secrets/secrets.json
+
+    # Build secrets.json from entries
```
ckelner: Just speaking from an end user perspective, proving inline examples is generally really helpful, or at least linking to docs where example configs can be found. The question "What do I put here?" comes to mind.

I assume it should look something like:

```
entries:
   SOURCE_DATADOG_AGENT_ADDRESS: "0.0.0.0:8282"
  # You can add more entries below
  # ANOTHER_VARIABLE: "another_value"
```

Or am I mistaken?

charts/observability-pipelines-worker/values.yaml
ckelner: Is there any reason to make this variable? Could we be opinionated and just say it is always bootstrap.yaml? Why expose it to the end user?

charts/observability-pipelines-worker/values.yaml
ckelner: Links to Docs or comments/examples would be helpful here -- what is a custom secrets backend? What values can I give for `backend_type`? I suspect this is coming and will be in our docs, but without any context I am left wondering when and why I use one vs the other.

charts/observability-pipelines-worker/values.yaml
graphcareful: Sounds good ! Yup thats right

charts/observability-pipelines-worker/values.yaml
graphcareful: Yeah theres no reason to make it configurable, then I guess I just hardcode it in my config maps?

charts/observability-pipelines-worker/values.yaml
graphcareful: Good point. The idea was that this is for any value that isn't `file`. Would writing comment in-line here be the place where users would look for documentation too?

charts/observability-pipelines-worker/values.yaml
ckelner: Yeah that makes most sense to me :+1:

charts/observability-pipelines-worker/values.yaml
ckelner: I think it is pretty common to offer a brief description inline as a nice to have and a link to docs. I don't think we want to be super verbose here, but offer the shortest explanation possible so users understand what it is for. Just to share some prior art, if you look at the Datadog agent helm chart values file, you'll see that almost every field is covered by some comments and in some cases links to docs, e.g. the secrets backend section of the agent: https://github.com/DataDog/helm-charts/blob/main/charts/datadog/values.yaml#L60-L86

Personally I find this really helpful when using a chart! So any color we can add here would be good. If docs won't be published until later I think jotting down a TODO in jira or wherever for later would be good for us to come back to it once we have docs with examples.

charts/observability-pipelines-worker/templates/_pod.tpl
```diff
@@ -91,6 +91,17 @@ containers:
...
       - name: data
         mountPath: "{{ .Values.datadog.dataDir | default "/var/lib/observability-pipelines-worker" }}"
+{{- if or .Values.bootstrapBackend.secretFile.enabled .Values.bootstrapBackend.custom.enabled }}
+      - name: opworker-bootstrap-defaults
+        mountPath: /etc/observability-pipelines-worker
+        readOnly: true
+{{- end }}
+{{- if eq (include "opw.bootstrapMode" .) "secretFile" }}
+      - name: secret-file-backend
+        mountPath: /etc/observability-pipelines-secrets/secrets.json
```
vladimir-dd:
```suggestion
        mountPath: {{ .Values.bootstrapBackend.secretFile.mountPath }}
```
should we use the variable here so that it mounts to the right place if the var changes?

charts/observability-pipelines-worker/templates/secret-file-backend.yaml
```diff
@@ -0,0 +1,12 @@
+{{- $secretFile := .Values.bootstrapBackend.secretFile | default dict }}
+
+{{- if $secretFile.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "opw.fullname" . }}-secret-file-backend
```
vladimir-dd, line 5:
```suggestion
metadata:
  name: {{ include "opw.fullname" . }}-secret-file-backend
  namespace: {{ .Release.Namespace }}
  labels: {{ include "opw.labels" . | nindent 4 }}
```
I guess we could add namespace and labels to keep all resources' metadata in sync.

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
```diff
@@ -0,0 +1,21 @@
+{{- $mode := include "opw.bootstrapMode" . }}
+
+{{- if $mode }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "opw.fullname" . }}-opworker-bootstrap-defaults
```
vladimir-dd:
```suggestion
metadata:
  name: {{ include "opw.fullname" . }}-opworker-bootstrap-defaults
  namespace: {{ .Release.Namespace }}
  labels: {{ include "opw.labels" . | nindent 4 }}
```
same here

vladimir-dd: nit: I'm wondering if we really need this `-defaults` postfix here and in the filename, especially given that we want to "hide" this file from users as per the Chris's comment.

charts/observability-pipelines-worker/values.yaml
vladimir-dd: agree, I think we actually _must_ make it a constant, given that users are aware of this file https://docs.datadoghq.com/observability_pipelines/configuration/install_the_worker/advanced_worker_configurations/

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
```diff
@@ -0,0 +1,21 @@
+{{- $mode := include "opw.bootstrapMode" . }}
```
vladimir-dd: I feel like we are mixing two entities in the same variable here: bootstrap file and secretsBackend, which can exist separately in case we want to add more config settings in the bootstrap file other than "secret".

If my understanding is correct, I'd recommend always having this bootstrap file configmap created and renaming this variable to something like `.Values.secretBackend`.
In this case we would move this `if opw.secretBackend` inside `data:` around `secret: `, so it'd look like:
```
data:
    bootstrap.yaml:
       {{- if .Values.secretBackend }}
             secret:
       {{- end }}
```
So if there is no secretBacked we'd have an empty configmap, which seems fine to me.
I'd also additionally let users override/set any other bootstrap config parameters, as described [here](https://docs.datadoghq.com/observability_pipelines/configuration/install_the_worker/advanced_worker_configurations/) with something like:

```
{{- if .Values.bootstrap.config }}
{{- toYaml .Values.bootstrap.config | nindent 4 }}
{{- end }}
{{- if $.Values.secretBackend }}
  secret:
{{- end }}
```
With this generic `Values.bootstrap.config` I'm wondering if we'd need the "custom" secret backend at all, since users could override the whole "secret" configuration in `Values.yaml`:
```
bootstrap:
  config:
     secret:
       .... 
```
, but I'm not sure what these custom modes are.
On a related note, do we support any "custom" mode today or is it just to accommodate any future modes? If it's only for the future modes, I'd probably remove it to avoid confusion, otherwise we should clearly comment and document what these custom modes are and how to configure them. It seems from the related comments though that we do support them and we are going to comment and document them in the public docs, in which case feel free to disregard my comment related to the "custom" mode.

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
```diff
@@ -0,0 +1,21 @@
...
+{{- if $mode }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "opw.fullname" . }}-opworker-bootstrap-defaults
+data:
+  {{ .Values.bootstrapBackend.filename }}: |
+    secret:
+      {{- if eq $mode "secretFile" }}
+      sgc_path: "/opt/datadog/observability-pipelines-worker/bin/datadog-secret-backend"
```
vladimir-dd: seems like this is a shared constant that could me moved outside "if" blocks? or maybe removed from it completely, since it seems like it's a binary we provide?

charts/observability-pipelines-worker/templates/_pod.tpl
```diff
@@ -121,6 +132,16 @@ volumes:
   - name: data
     emptyDir: {}
 {{- end }}
+{{- if or .Values.bootstrapBackend.secretFile.enabled .Values.bootstrapBackend.custom.enabled }}
```
vladimir-dd: as explained above, I'd remove this "if" around bootstrap configmap and have it always created.

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
vladimir-dd: ah, I get it - I guess there might be external providers.

charts/observability-pipelines-worker/templates/_pod.tpl
graphcareful: I didn't want to expose this to the user since they pass in k/v secrets in `values.yaml` and I just place it for them in this file. Where exactly the secrets go is the implementation detail.

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
graphcareful: Chatted with Vlad offline about this, mainly boils down to weather we also want to expose the bootstrap file as well and keep some of what is here as a convenience to users who want to use the file backend. We could then remove the custom mode, and just expect users to craft the correct bootstrap themselves in the case they don't want to use the file backend. Thoughts @ckelner ?

charts/observability-pipelines-worker/templates/_pod.tpl
vladimir-dd: yeah, agree, that makes sense, I guess we should remove this var from `Values.yaml` then?

charts/observability-pipelines-worker/templates/configmap-opworker-bootstrap-defaults.yaml
graphcareful: `Values.bootstrap.config` has been implemented in latest changes

charts/observability-pipelines-worker/templates/_pod.tpl
graphcareful: Removed

charts/observability-pipelines-worker/templates/_pod.tpl
graphcareful: I left this in because `bootstrap.yaml` will then not be "created" and what will exist is the file as it is in the image. This is the default version of `bootstrap.yaml` found [here](https://github.com/DataDog/observability-pipelines-worker/blob/main/dist/bootstrap.yaml) in our OPW repo.

graphcareful:
Changes in [force-push](https://github.com/DataDog/helm-charts/compare/a400dfd67481c414c2f711089d53fafca723d3ab..d1939f71f2b0e85cea8491d9e8a07301af621337)
- Rebase against main

charts/observability-pipelines-worker/values.yaml
emilychendd: resolving since the latest change addressed this.

charts/observability-pipelines-worker/README.md
```diff
@@ -91,6 +91,9 @@ The command removes all the Kubernetes components associated with the chart and
 | containerPorts | list | `[]` | Manually define ContainerPort array, overriding automated generation of ContainerPorts. |
 | datadog.apiKey | string | `nil` | Specify your Datadog API key. |
 | datadog.apiKeyExistingSecret | string | `""` | Specify a preexisting Secret that has your API key instead of creating a new one. The value must be stored under the `api-key`. |
+| datadog.bootstrap | object | `{"config":{},"secretFileContents":{}}` | Provide a bootstrap file that conforms to the options provided in this documentation:   https://docs.datadoghq.com/observability_pipelines/configuration/install_the_worker/advanced_worker_configurations/#bootstrap-options |
+| datadog.bootstrap.config | object | `{}` | The bootstrap file contents |
+| datadog.bootstrap.secretFileContents | object | `{}` | Additional helper for the "secrets" portion of the bootstrap file. Use if your backend_type is of type 'file'. Helm chart will copy the provided secrets into a new file, and correctly setup the bootstrap to point to the secrets file. |
```
vladimir-dd: nice, I like how you framed it.

charts/observability-pipelines-worker/values.yaml
```diff
@@ -38,6 +38,15 @@ datadog:
...
     address: "0.0.0.0:8686"
+  # datadog.bootstrap -- Provide a bootstrap file that conforms to the options provided in this documentation:
+  #   https://docs.datadoghq.com/observability_pipelines/configuration/install_the_worker/advanced_worker_configurations/#bootstrap-options
+  bootstrap:
+    # datadog.bootstrap.config -- The bootstrap file contents
+    config: {}
+    # datadog.bootstrap.secretFileContents -- Additional helper for the "secrets" portion of the bootstrap file.
+    # Use if your backend_type is of type 'file'. Helm chart will copy the provided secrets into a new file,
+    # and correctly setup the bootstrap to point to the secrets file.
+    secretFileContents: {}
```
vladimir-dd, line 50: nit: could we add one line commented out example of entries in this file?

vladimir-dd (approved):
LGTM!

Tested with both `config` and secrets:
```
# cat /etc/observability-pipelines-secrets/secrets.json
{"SOURCE_DATADOG_AGENT_ADDRESS":"0.0.0.0:8282","dfasdfsdf":"sfsdfdfsf"}
# cat /etc/observability-pipelines-worker/bootstrap.yaml
tags:
- adsf:dsfdsf
secret:
  sgc_path: /opt/datadog/observability-pipelines-worker/bin/datadog-secret-backend
  backend_type: json
  backend_config:
    file_path: /etc/observability-pipelines-secrets/secrets.json#

PR #2242, opened 2025-12-16
set datadog.autoscaling.workload.enabled empty by default

### LINKED ISSUE #2241: datadog-3.154.1 introduces breaking change
The latest version of the helm chart (3.154.1) introduces a breaking change resulting in CRD deletion. I can reliably rollback to the previous release (3.154.0) and have the CRDs get restored.

~~I am still investigating the root cause. I believe it may be resulting from downstream [changes](https://github.com/DataDog/helm-charts/pull/2221/changes) in datadog-crds chart adding the `keepCrds` annotation but I don't have a fix yet~~

I believe I have found the root cause and it is a small detail in how conditions are evaluated for dependencies:
> The first condition path that exists wins and subsequent ones for that chart are ignored.  ([Docs](https://helm.sh/docs/topics/charts/#tags-and-condition-resolution))

The CRDs are conditionally rendered based on:
```
condition: datadog.autoscaling.workload.enabled,clusterAgent.metricsProvider.useDatadogMetrics
```
While `datadog.autoscaling.workload.enabled` used to be omitted it is now explicitly false: https://github.com/DataDog/helm-charts/blame/main/charts/datadog/values.yaml#L1295

This results in the CRDs being removed from any existing configurations without autoscaling enabled

Release Diffs: https://github.com/DataDog/helm-charts/compare/datadog-3.153.0..datadog-3.154.1

celenechang:
Fixes #2241

#### Checklist
- [X] Chart Version semver bump label added (use `<chartName>/minor-version`, `<chartName>/patch-version`, or `<chartName>/no-version-bump`)
- [X] For `datadog` or `datadog-operator` chart or value changes, update the test baselines (run: `make update-test-baselines`)

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -1296,7 +1296,7 @@ datadog:
   autoscaling:
     workload:
       # datadog.autoscaling.workload.enabled -- Enable Workload Autoscaling.
-      enabled: false
+      enabled:
 
 ## This is the Datadog Cluster Agent implementation that handles cluster-wide
 ## metrics more cleanly, separates concerns for better rbac, and implements
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.155.1
+
+* Change default value for `datadog.workload.autoscaling.enabled` to be empty. Fixes issue [#2241](https://github.com/DataDog/helm-charts/issues/2241) in chart 3.154.1.
+
 ## 3.155.0
 
 * Allow activation of cluster autoscaling.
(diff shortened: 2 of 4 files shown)


charts/datadog/values.yaml
```diff
@@ -1296,7 +1296,7 @@ datadog:
   autoscaling:
     workload:
       # datadog.autoscaling.workload.enabled -- Enable Workload Autoscaling.
-      enabled: false
+      enabled:
```
atmask, line 1299: Was it validated that nil won't be implicitly evaluated as false still by helm when evaluating the `condition`? The path `datadog.autoscaling.workload.enabled` still exists in the values with this change (@fanny-jiang & @celenechang)

celenechang: Thanks @atmask - unfortunately it gives a warning that it is a `non-bool value`, but it does maintain the previous behavior

PR #2266, opened 2025-12-24
Update datadog-csi-driver chart to support SSI

iamluc:

This PR adds a persistent storage to the Datadog CSI driver DaemonSet.
The [CSI driver](https://github.com/DataDog/datadog-csi-driver/pull/48) now needs to store downloaded data and an associated database that must persist across pod rollouts.

It also adds a new config key `disableSSI` to disable Single Step Instrumentation support

https://datadoghq.atlassian.net/browse/INPLAT-891

#### Checklist

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.163.1
+
+* Update default Agent version to 7.75.0 ([#2326](https://github.com/DataDog/helm-charts/pull/2326)).
+
 ## 3.163.0
 
 * [AGTMETRICS-393] Update Agent Data Plane support to reflect new "Data Plane" terminology. ([#2313](https://github.com/DataDog/helm-charts/pull/2313)).
@@ -26,7 +30,7 @@
 
 ## 3.160.4
 
-* Add "watch" permission for cluster roles and cluster role bindings in the Cluster Agent RBAC when the CSPM feature is enabled.
+* Add "watch" permission for role bindings and cluster role bindings in the Cluster Agent RBAC when the CSPM feature is enabled.
 
 ## 3.160.3
 
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: datadog
-version: 3.163.0
+version: 3.163.1
 appVersion: "7"
 description: Datadog Agent
 keywords:
--- a/charts/datadog/README.md
+++ b/charts/datadog/README.md
@@ -1,6 +1,6 @@
 # Datadog
 
-![Version: 3.163.0](https://img.shields.io/badge/Version-3.163.0-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
+![Version: 3.163.1](https://img.shields.io/badge/Version-3.163.1-informational?style=flat-square) ![AppVersion: 7](https://img.shields.io/badge/AppVersion-7-informational?style=flat-square)
 
 > [!WARNING]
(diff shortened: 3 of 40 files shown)


betterengineering (approved):

charts/datadog-csi-driver/templates/daemonset.yaml
```diff
@@ -56,6 +59,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            {{- if .Values.disableSSI }}
+            - name: DD_DISABLE_SSI
+              value: {{ .Values.disableSSI | quote  }}
+            {{- end }}
```
adel121:
```suggestion
            - name: DD_DISABLE_SSI
              value: {{ .Values.disableSSI | quote  }}
```

I believe this should be sufficient.

charts/datadog-csi-driver/values.yaml
```diff
@@ -50,6 +50,9 @@ sockets:
   # Should correspond to `datadog.dsd.hostSocketPath`
   dsdHostSocketPath: /var/run/datadog/dsd.socket
 
+# disableSSI -- Disable Single Step Instrumentation (SSI) support for the CSI driver.
+disableSSI: false
```
adel121: I am thinking if it would be better to have something like:

```
apm:
    enabled: true
```

On the longterm I think what we want is a flag to indicate if the feature is enabled or not, rather than a flag that disables the feature. I understand they serve the same thing, but `disableSSI` is more of a debugging flag IMHO, so it might not be the best way to expose the feature once it is GA. 

WDYT ?

charts/datadog-csi-driver/values.yaml
```diff
@@ -50,6 +50,9 @@ sockets:
   # Should correspond to `datadog.dsd.hostSocketPath`
   dsdHostSocketPath: /var/run/datadog/dsd.socket
 
+# disableSSI -- Disable Single Step Instrumentation (SSI) support for the CSI driver.
```
adel121: Should we also indicate the feature is in BETA ? like add a `[Beta]` prefix ?

adel121:
General comment: 

CSI driver has a gke autopilot allowlist. 

If we deploy the chart as it is on gke autopilot, it will probably fail (not 100% sure though). We need to update the gke allowlist and request google's approval to make the new changes defined in the allowlist. 

It is not necessary to do this now since the feature is not yet mature and this will slow us down. But we could at least ensure the csi chart can still be deployed on gke autopilot (for example by omitting the introduced changes if we are on gke autopilot) and document that SSI is still not supported on gke autopilot.  

We already have [an e2e test for CSI on gke autopilot](https://github.com/DataDog/helm-charts/blob/main/test/e2e/datadog/e2e_gke_autopilot_csi_test.go), I am surprised why it didn't fail 🤔

adel121:
> General comment:
> 
> CSI driver has a gke autopilot allowlist.
> 
> If we deploy the chart as it is on gke autopilot, it will probably fail (not 100% sure though). We need to update the gke allowlist and request google's approval to make the new changes defined in the allowlist.
> 
> It is not necessary to do this now since the feature is not yet mature and this will slow us down. But we could at least ensure the csi chart can still be deployed on gke autopilot (for example by omitting the introduced changes if we are on gke autopilot) and document that SSI is still not supported on gke autopilot.
> 
> We already have [an e2e test for CSI on gke autopilot](https://github.com/DataDog/helm-charts/blob/main/test/e2e/datadog/e2e_gke_autopilot_csi_test.go), I am surprised why it didn't fail 🤔

Actually the test did fail, it was just not automatically invoked https://gitlab.ddbuild.io/DataDog/helm-charts/-/jobs/1365221627

So we are good.

charts/datadog-csi-driver/values.yaml
iamluc: :+1: Done

charts/datadog-csi-driver/values.yaml
iamluc: I feel this could be a bit confusing. The Helm chart option itself is not in beta and is enabled by default.

The feature in the CSI driver is, but customers should use it via the cluster-agent configuration, where the experimental status is documented.

WDYT?

PR #2269, opened 2025-12-24
fix(datadog-operator): lookup endpoint-config ConfigMap by exact name

### LINKED ISSUE #2268: [datadog-operator subchart] Multiple releases in same namespace produce malformed secret names
## Description
When running multiple Datadog Helm releases in the same namespace (e.g., `datadog` and `karpenter-datadog`), the operator subchart's secret name resolution produces malformed names by concatenating values from ALL matching ConfigMaps.

## Expected Behavior
The operator should only read from its own parent release's endpoint-config ConfigMap.

## Actual Behavior
Secret names are concatenated:
- Expected: `karpenter-datadog` or `datadog`
- Actual: `datadogkarpenter-datadog`

This causes `CreateContainerConfigError` because the secret doesn't exist:
```
Error: secret "datadogkarpenter-datadog" not found
```

## Root Cause
In `charts/datadog-operator/templates/_helpers.tpl`, the `get-endpoint-config-data-key` template:

```yaml
{{- define "get-endpoint-config-data-key" -}}
{{- $ctx := index . 0 }}
{{- $key := index . 1 }}
{{- $ns := $ctx.Release.Namespace -}}
{{- $list := lookup "v1" "ConfigMap" $ns "" -}}
{{- if $list }}
  {{- range $list.items }}
    {{- if hasSuffix "endpoint-config" .metadata.name }}
      {{- get .data $key -}}
    {{- end }}
  {{- end }}
{{- end }}
{{- end -}}
```

This iterates over ALL ConfigMaps ending in `endpoint-config` and outputs ALL matching values without any delimiter or early exit. With two releases:
- `datadog-endpoint-config` → `api-key-secret-name: datadog`
- `karpenter-datadog-endpoint-config` → `api-key-secret-name: karpenter-datadog`

The template outputs: `datadogkarpenter-datadog`

## Reproduction
1. Install two Datadog releases in the same namespace with operator enabled:
   ```bash
   helm install datadog datadog/datadog -n dev-tools --set datadog.operator.enabled=true
   helm install karpenter-datadog datadog/datadog -n dev-tools --set datadog.operator.enabled=true
   ```
2. Observe operator pods fail with `CreateContainerConfigError`
3. Check pod events: `secret "datadogkarpenter-datadog" not found`

## Suggested Fix
The template should look up the specific ConfigMap by exact name rather than iterating over all matching suffixes:

```yaml
{{- define "get-endpoint-config-data-key" -}}
{{- $ctx := index . 0 }}
{{- $key := index . 1 }}
{{- $ns := $ctx.Release.Namespace -}}
{{- $expectedName := printf "%s-endpoint-config" $ctx.Release.Name -}}
{{- $cm := lookup "v1" "ConfigMap" $ns $expectedName -}}
{{- if $cm }}
  {{- get $cm.data $key -}}
{{- end }}
{{- end -}}
```

## Environment
- Datadog Helm Chart: 3.157.0, 3.157.1
- Kubernetes: 1.27+
- First appeared in: 3.157.0 (PR #2112 - Enable Datadog Operator chart dependency)

## Workaround
Pin chart version to `3.156.3` which doesn't include the operator subchart:
```yaml
spec:
  chart:
    spec:
      version: "3.156.3"  # instead of ^3.132.0
```

chad-fossa:
## Summary

When multiple Datadog Helm releases exist in the same namespace (e.g., `datadog` and `karpenter-datadog`), the `get-endpoint-config-data-key` template iterates over ALL ConfigMaps ending in `endpoint-config` and concatenates their values, producing malformed secret names.

**Before (broken):**
- `datadog-endpoint-config` → `api-key-secret-name: datadog`
- `karpenter-datadog-endpoint-config` → `api-key-secret-name: karpenter-datadog`
- Template output: `datadogkarpenter-datadog` (concatenated!)

**After (fixed):**
- Looks up ConfigMap by exact name: `<release-name>-endpoint-config`
- Each release's operator gets only its own ConfigMap's values

## Changes

Modified `charts/datadog-operator/templates/_helpers.tpl`:
- Changed `get-endpoint-config-data-key` to use `lookup` with exact ConfigMap name instead of iterating over all matching suffixes

## Testing

1. Deploy two Datadog releases in the same namespace with operator enabled
2. Verify operator pods start correctly without `CreateContainerConfigError`
3. Verify each operator references the correct secret

## Related

Fixes #2268

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.17.0-dev.2
+
+* Fix endpoint-config ConfigMap lookup to use exact name instead of suffix matching, preventing value concatenation when multiple Datadog releases exist in the same namespace.
+
 ## 2.17.0-dev.1
 
 * Update Datadog Operator chart for to 1.22.0-rc.1.
--- a/charts/datadog-operator/README.md
+++ b/charts/datadog-operator/README.md
@@ -1,6 +1,6 @@
 # Datadog Operator
 
-![Version: 2.17.0-dev.1](https://img.shields.io/badge/Version-2.17.0--dev.1-informational?style=flat-square) ![AppVersion: 1.22.0-rc.1](https://img.shields.io/badge/AppVersion-1.22.0--rc.1-informational?style=flat-square)
+![Version: 2.17.0-dev.2](https://img.shields.io/badge/Version-2.17.0--dev.2-informational?style=flat-square) ![AppVersion: 1.22.0-rc.1](https://img.shields.io/badge/AppVersion-1.22.0--rc.1-informational?style=flat-square)
 
 ## Values
 
--- a/test/datadog-operator/baseline/Operator_Deployment_default.yaml
+++ b/test/datadog-operator/baseline/Operator_Deployment_default.yaml
@@ -7,7 +7,7 @@ metadata:
   namespace: datadog-agent
   labels:
     app.kubernetes.io/name: datadog-operator
-    helm.sh/chart: datadog-operator-2.17.0-dev.1
+    helm.sh/chart: datadog-operator-2.17.0-dev.2
     app.kubernetes.io/instance: datadog-operator
     app.kubernetes.io/version: "1.22.0-rc.1"
     app.kubernetes.io/managed-by: Helm


chad-fossa:
Let me figure that out <3 I'll definitely enable writes from maintainers

chad-fossa:
Huzzah!  box checked.  Today I learned :D

chad-fossa:
currently building/pushing requested changes

chad-fossa:
Should be up now :D sorry if we conflict'd

PR #2295, opened 2026-01-15
[ACTP] add support for taskTimeoutSeconds and httpTimeoutSeconds

dd-gplassard:

Allow customers to override taskTimeoutSeconds and httpTimeoutSeconds configuration for their private action runner

  - fixes #

#### Checklist

### DIFF
--- a/.github/workflows/release-crds.yaml
+++ b/.github/workflows/release-crds.yaml
@@ -48,3 +48,4 @@ jobs:
           mark_as_latest: ${{ steps.is_prerelease.outputs.mark_as_latest }}
         env:
           CR_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
+          CR_GENERATE_RELEASE_NOTES: true
--- a/.github/workflows/release-operator.yaml
+++ b/.github/workflows/release-operator.yaml
@@ -48,3 +48,4 @@ jobs:
           mark_as_latest: ${{ steps.is_prerelease.outputs.mark_as_latest }}
         env:
           CR_TOKEN: '${{ secrets.GITHUB_TOKEN }}'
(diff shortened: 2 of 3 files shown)


charts/private-action-runner/values.yaml
```diff
@@ -54,6 +54,10 @@ runner:
     actionsAllowlist: []
     # -- List of tags to be added to metrics and logs published by the runner. The tags must be specified in a 'key:value' format.
     tags: []
+    # -- Global timeout for task executions. Use 0 for no timeout.
+    taskTimeoutSeconds: 0
+    # -- Global http client timeout for http based actions.
```
merchristK, line 59: can we also use 0 for no timeout in http?

dd-gplassard: we would need to do another release and change the behavior in the PAR and I'm not sure there is that much of a use case. 

The "annoying" thing is that for global tasks, no value meant infinity because we had no previous support while for http no value meant 30 seconds because it was hardcoded like that

merchristK: I see. I was just curious. I think we should surface it. Or remove the comment that tells user to use 0 for no value. it might be misleading.

merchristK: They can think that 0 also means no timeout for http

dd-gplassard: I think its even more confusing if we don't have `Use 0 for no timeout.` ? 

Then it becomes
```
# -- Global timeout for task executions. 
taskTimeoutSeconds: 0
# -- Global http client timeout for http based actions.
httpTimeoutSeconds: 30
```
And we kind of want to put the `taskTimeoutSeconds` otherwise its not discoverable

PR #2312, opened 2026-01-21
Conditionally set env vars to match datadog-operator: logs, prometheus

fanny-jiang:

* `DD_LOGS_CONFIG_K8S_CONTAINER_USE_FILE`: only set when logs.enabled=true
* `DD_PROMETHEUS_SCRAPE_SERVICE_ENDPOINTS`: added to agent container to match the [datadog-operator logic](https://github.com/DataDog/datadog-operator/blob/ac6240550b21f86607c0f3f99488f22474318515/internal/controller/datadogagent/feature/prometheusscrape/feature.go#L132-L135) (before, it was only added to the cluster-agent deployment)

These changes align the datadog chart behavior with that of the datadog-operator and avoids setting env vars when they're not explicitly needed. 

  - fixes #

#### Checklist

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2.18.0-dev.4
+
+* Update Datadog Operator chart for 1.23.0-rc.3.
+
+## 2.18.0-dev.3
+
+* Add POD_NAMESPACE environment variable via downward API.
+
 ## 2.18.0-dev.2
 
 * Update Datadog Operator chart for 1.23.0-rc.2.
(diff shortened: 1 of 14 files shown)


charts/datadog/templates/_processes-common-env.yaml
```diff
@@ -7,8 +7,10 @@
   value: {{ .Values.datadog.processAgent.containerCollection | quote }}
 - name: DD_PROCESS_AGENT_DISCOVERY_ENABLED
   value: {{ .Values.datadog.processAgent.processDiscovery | quote }}
+{{- if and (eq .Values.datadog.processAgent.processCollection true) .Values.datadog.processAgent.stripProcessArguments }}
```
daniel-taf: Maybe the operator needs updating, but we don't want to check `Values.datadog.processAgent.processCollection` here. Due to the newer Service Discovery feature, processes may be collected without this config set. We still want to strip args in that case.

fanny-jiang: Got it, thanks for the context! I'll revert those changes

PR #2315, opened 2026-01-22
[ACTP] bump helm charts to 1.21.1

chrisnad:
bump PAR version to v1.17.1: Fix Private Action Runner script validation

### DIFF
--- a/charts/observability-pipelines-worker/CHANGELOG.md
+++ b/charts/observability-pipelines-worker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.12.3
+
+- Change the default podManagementPolicy to Parallel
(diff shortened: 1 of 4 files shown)


charts/private-action-runner/Chart.yaml
```diff
@@ -3,8 +3,8 @@ name: private-action-runner
 description: A Helm chart to deploy the private action runner
 
 type: application
-version: 1.21.0
-appVersion: "v1.17.0"
+version: 1.21.1
```
merchristK, line 6: I would have bumped the minor instead of the patch

chrisnad: actually this matches other examples in the git history

PR #2354, opened 2026-02-05
[OTAGENT-794] Host Profiler bundled support

mackjmr:
This PR adds experimental support for the host profiler

OTAGENT-794

Not sure how to note that this feature is experimental and subject to change (e.g. we only have a public dev image for now). In the operator, we used annotations support rather than full feature support in order to convey this.

#### Checklist

### DIFF
--- a/charts/datadog/values.yaml
+++ b/charts/datadog/values.yaml
@@ -1411,7 +1411,7 @@ clusterAgent:
     name: cluster-agent
 
     # clusterAgent.image.tag -- Cluster Agent image tag to use
-    tag: 7.75.4
+    tag: 7.76.1
 
     # clusterAgent.image.digest -- Cluster Agent image digest to use, takes precedence over tag if specified
     digest: ""
@@ -2002,7 +2002,7 @@ agents:
     name: agent
 
     # agents.image.tag -- Define the Agent version to use
-    tag: 7.75.4
+    tag: 7.76.1
 
     # agents.image.digest -- Define Agent image digest to use, takes precedence over tag if specified
     digest: ""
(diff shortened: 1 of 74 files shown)


charts/datadog/values.yaml
```diff
@@ -698,6 +698,27 @@ datadog:
       # datadog.otlp.logs.enabled -- Enable logs support in the OTLP ingest endpoint
       enabled: false
 
+  ## Host Profiler related configuration for the host-profiler in Agent Daemonset. Note this is experimental and subject to change
+  hostProfiler:
+    # datadog.hostProfiler.enabled -- Enable the Host Profiler
+    enabled: false
+    # datadog.hostProfiler.image -- Image the Host Profiler. This parameter will be removed once official image is available.
+    image: ""
```
fanny-jiang, line 724:
```suggestion
  hostProfiler:
    # datadog.hostProfiler.enabled -- Enable the Host Profiler. This feature is experimental and subject to change.
    enabled: false
    # datadog.hostProfiler.image -- Image the Host Profiler. This parameter is experimental and will be removed once official image is available.
    image: ""
```

charts/datadog/templates/_helpers.tpl
```diff
@@ -152,6 +152,17 @@ false
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return true if the Host Profiler needs to be deployed
+*/}}
+{{- define "should-enable-host-profiler" -}}
+{{- if and .Values.datadog.hostProfiler.enabled (eq .Values.targetSystem "linux") -}}
```
fanny-jiang:
```suggestion
{{- if and .Values.datadog.hostProfiler.enabled (eq .Values.targetSystem "linux") (not .Values.providers.gke.gdc) (not .Values.providers.gke.autopilot) -}}
```

fanny-jiang: Can we also add a failure gate to the `NOTES.txt` file to prevent upgrade/install if hostProfiler is enabled for autopilot and GKE GDC environments?

charts/datadog/values.yaml
mackjmr: Updated 👍

charts/datadog/templates/_helpers.tpl
mackjmr: Added failure gate

PR #2363, opened 2026-02-09
Add label-based fallback for endpoint-config ConfigMap lookup

fanny-jiang:

Adds label-based discovery fallback to the Operator's chart's endpoint-config ConfigMap lookup logic to support aliased datadog chart installations. The Operator uses the fields defined in the parent Datadog chart's endpoint-config to configure API access to Datadog for surfacing Helm metadata in the [Fleet Automation](https://docs.datadoghq.com/agent/fleet_automation/#control-access-to-fleet-automation) feature. 

Previously, the implementation looked up the ConfigMap by exact name: `<releaseName>-endpoint-config`. With the datadog chart changes in https://github.com/DataDog/helm-charts/pull/2362, the ConfigMap name now is prepended with the alias name (if installed with chart alias). 

**The problem**: The Operator chart only shares the`.Release.Name` with the parent Datadog chart and has no way to reference the parent chart's alias, so exact-name lookup fails.

This PR adds label-based discovery as a fallback option in aliased installs. The lookup logic now:
1. Try exact name first (`<releaseName>-endpoint-config`): standard non-aliased installs
2. Fall back to label-based discovery: aliased sub-chart installs
`lookup` ConfigMaps matching:
    - `datadoghq.com/component: endpoint-config`
    - `app.kubernetes.io/instance: <releaseName>`
3. If multiple matches, pick the alphabetically first (via `sortAlpha`) for deterministic behavior. 
**Caveat**: If dual-shipping using the aliased-chart installation mode, only 1 endpoint-config CM is used. 

Backward compatibility is preserved for multi-release installations.

This PR also adds the linux nodeSelector to the values defaults to prevent the operator pod from being scheduled on Windows nodes. 

  - fixes #

#### Checklist


charts/datadog-operator/templates/deployment.yaml
```diff
@@ -186,8 +186,8 @@ spec:
           securityContext:
             {{- toYaml .Values.containerSecurityContext | nindent 12 }}
           {{- end }}
-      {{- with .Values.nodeSelector }}
       nodeSelector:
```
swang392: not sure if this change is needed, it would break helm templating if there is no `nodeSelector` defined.

Issue #2338, opened 2026-01-29
Use finegrained kubelet API authorization


vinayakankugoyal:
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/DataDog/helm-charts/issues) and this is not a duplicate
- [x] This is a Helm chart issue, not a Datadog product/service problem

### Helm chart name

datadog-operator

### Helm chart version

3.x.x

### Bug Report

**What happened:**
Granting nodes/proxy with even the get verb can be used by attackers to gain privileges. This risk has been highlighted in the kubernetes documentation in https://kubernetes.io/docs/concepts/security/rbac-good-practices/#access-to-proxy-subresource-of-nodes and https://kubernetes.io/docs/concepts/security/api-server-bypass-risks/#kubelet-api and more recently in - https://grahamhelton.com/blog/nodes-proxy-rce.

We found the following instances of nodes/proxy being used in this repository

https://grep.app/search?f.lang=YAML&f.path=charts%2F&f.repo=DataDog%2Fhelm-charts&q=nodes%2Fproxy

**What I expected:**
Please update the workloads to use fine-grained kubelet API authorization [KEP](https://kep.k8s.io/2862) and replace nodes/proxy with more fine-grained permissions. See [this](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/#fine-grained-authorization) link for how to update your RBAC config.

### Environment

**Kubernetes version:** 
**Helm version:** 
**Cloud provider:**

vinayakankugoyal:
https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/rbac.yaml handles this correctly and hopefully we can update this chart as well

PR #2049, opened 2025-09-15
Added datadog.kubelet.fineGrainedAuthorization flag for supporting fine grained Kubelet API authorization

triviajon:
#### What this PR does / why we need it:
This PR updates the `datadog` helm chart to add a new flag `datadog.kubelet.fineGrainedAuthorization` which is disabled by default. If enabled, this replaces the `nodes/proxy` endpoint with the `nodes/pods`, `nodes/healthz`, `nodes/configz`, and `nodes/log` endpoints in the RBAC. This is possible by this [new feature](https://github.com/kubernetes/enhancements/issues/2862) introduced in K8s 1.32+. 

#### Special notes for your reviewer:
Tested changes by:

1. Created a local cluster and deploying the Datadog agent onto it using the updated Helm chart in this PR: `helm install datadog-agent -f values.yaml . --set datadog.apiKey="$DD_API_KEY" --set datadog.appKey="$DD_APP_KEY" --set datadog.kubelet.fineGrainedAuthentication=true --set datadog.kubelet.tlsVerify=false --set datadog.clusterName="jrosario-dev-cluster-other"`
2. Checked that the kubelet check works by running `agent check kubelet` in the datadog-agent pod, also checked that pod collection is still working at https://dddev.datadoghq.com/. 

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version semver bump label added (use `<chartName>/minor-version`, `<chartName>/patch-version`, or `<chartName>/no-version-bump`)
- [x] For `datadog` or `datadog-operator` chart or value changes, update the test baselines (run: `make update-test-baselines`)

GitHub CI takes care of the below, but are still required:
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] `CHANGELOG.md` has been updated 
- [x] Variables are documented in the `README.md`

PR #2364, opened 2026-02-10
[charts/datadog-operator] Add clusterRole.kubelet.fineGrainedAuthorization option

horjulf:
  - fixes https://github.com/DataDog/helm-charts/issues/2338

Follow up to https://github.com/DataDog/helm-charts/pull/2049

#### Checklist

### DIFF
--- a/charts/datadog-operator/templates/clusterrole.yaml
+++ b/charts/datadog-operator/templates/clusterrole.yaml
@@ -55,12 +55,15 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes/configz
-  - nodes/healthz
-  - nodes/logs
   - nodes/metrics
+  {{- if .Values.clusterRole.kubelet.fineGrainedAuthorization }}
   - nodes/pods
+  - nodes/healthz
+  - nodes/configz
+  - nodes/logs
+  {{- else }}
   - nodes/proxy
+  {{- end }}
   - nodes/spec
   - nodes/stats
   verbs:
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.18.0-dev.5
+
+* Add `clusterRole.kubelet.fineGrainedAuthorization` option, allowing for finer grained kubelet API authorization (Kubernetes 1.32+).
+
 ## 2.18.0-dev.4
 
 * Update Datadog Operator chart for 1.23.0-rc.3.
--- a/charts/datadog-operator/Chart.yaml
+++ b/charts/datadog-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: datadog-operator
-version: 2.18.0-dev.4
+version: 2.18.0-dev.5
 appVersion: 1.23.0-rc.3
 description: Datadog Operator
 keywords:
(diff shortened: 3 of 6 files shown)


charts/datadog-operator/templates/clusterrole.yaml
```diff
@@ -55,12 +55,15 @@ rules:
...
   resources:
-  - nodes/configz
-  - nodes/healthz
-  - nodes/logs
   - nodes/metrics
+  {{- if .Values.clusterRole.kubelet.fineGrainedAuthorization }}
   - nodes/pods
+  - nodes/healthz
+  - nodes/configz
+  - nodes/logs
```
horjulf, line 63: The individual permissions were added in https://github.com/DataDog/helm-charts/pull/2108.
Should the condition just remove `nodes/proxy` and leave existing resources ?

horjulf:
@triviajon could you review this since you added support on the agent side previously ? 🙏

tbavelier:
Hello @horjulf ,

Thank you for the contribution!

At the moment, we won't proceed with this solution as the operator is not able to re-concile if it tries to grant a permission (e.g. `nodes/proxy`) it doesn't have itself. This permission is granted by default unless the annotation `agent.datadoghq.com/fine-grained-kubelet-authorization-enabled: "true"` is added on the `DatadogAgent` custom resource (reference https://github.com/DataDog/datadog-operator/pull/2188) which can't be templated since this custom resource is not part of the deployment. Such a flag risks breaking users not aware of the requirements to use it (Kubernetes 1.33, or 1.32 with the flag set to true, as well as the annotation on the `DatadogAgent` object).

With that in mind, if you'd prefer not to use `nodes/proxy` due to the exposure, we currently recommend creating the service account and RBAC for the operator yourself (using the base `rbac.yaml` template from this chart/https://github.com/DataDog/datadog-operator/blob/main/config/rbac/role.yaml) and making sure to use the provided annotation on your `DatadogAgent` resource.

```yaml
rbac:
  create: false
serviceAccount:
  create: false
  name: "custom-datadog-operator"
```

edit: to clear any confusion, please note the above reply is not our final stance on this matter, we plan to indeed eventually provide a safe opt-in option to avoid granting `nodes/proxy` and have the operator not try to grant it to the Agent, but we have yet to decide on the approach. Thank you for your patience and understanding!

horjulf:
:wave: @tbavelier

Thank you for the review and context.
I understand the concern but maintaining a role outside of the chart can lead to bigger issues.

Until there is a better approach, would you consider keeping the option with stronger wording on documentation or even undocumented ?
The operator and agent are very likely maintained by the same team, would be good to allow the administrator to do an educated choice and improve cluster security.

tbavelier:
@horjulf 

I understand and I _personally_ agree a temporary solution templated in the chart might work better until we get to a proper solution/decision. I created https://github.com/DataDog/helm-charts/pull/2392 which is pretty much the same PR as yours with more warnings.

However, this needs to be discussed within our team if the approach is sufficient in terms of warnings, maintainability and that it stays on our radar.
I will update this thread accordingly after.

Regards,

tbavelier:
Hello again,

Closing this as we indeed went with the flag to allow this. As it's only available in the devel chart which is our lifecycle for the release candidate versions of the operator, it won't be available in the stable version until we complete `1.24.0` release cycle. With that being said, in the meantime, you could use the `--devel` version while pinning a non release candidate version of the operator using `image.tag` set to `1.23.1`: https://github.com/DataDog/helm-charts/blob/66db4923e80778e6190bac6aec7e937af0c7ca20/charts/datadog-operator/values.yaml#L50, e.g. `helm upgrade --install datadog-operator datadog/datadog-operator --devel --set image.tag=1.23.1 --set clusterRole.kubeletFineGrainedAuthorization=true`

Otherwise, you can wait for the next stable version that should be in a few weeks.

Thanks!

PR #2421, opened 2026-03-02
[BARX-1655] Add registryMigrationMode for per-site DD_REGISTRY_OVERRIDE_* env vars

avonengel:

Adds helm chart support for the per-site `DD_REGISTRY_OVERRIDE_*` env vars that configure the operator to use `registry.datadoghq.com` as the default image registry for agent images, replacing site-based regional mirrors (GCR, ECR, ACR).

A new `registryMigrationMode` setting controls the behavior:
- `"auto"` (default): enables overrides for sites where this is enabled by default (currently Asia)
- `"all"`: enables all per-site overrides (`DD_REGISTRY_OVERRIDE_ASIA`, `_DEFAULT`, `_EU`, `_AZURE`)
- `"none"`: disables all overrides

**Stack: 1/2** — base PR introducing the registry migration infrastructure.

  - fixes https://datadoghq.atlassian.net/browse/BARX-1655

This depends on the operator-side change in [DataDog/datadog-operator#2669](https://github.com/DataDog/datadog-operator/pull/2669) which adds the `DD_REGISTRY_OVERRIDE_*` env vars. The env vars are harmless if the operator doesn't recognize them yet.

PR #2430 builds on this to switch the operator's own image to `registry.datadoghq.com`.

#### Checklist

### DIFF
--- a/charts/datadog-operator/CHANGELOG.md
+++ b/charts/datadog-operator/CHANGELOG.md
@@ -1,8 +1,11 @@
 # Changelog
 
-## 2.19.0-dev.6
+## 2.19.0-dev.7
 
 * Add `registryMigrationMode` to control gradual migration of Agent image pulls to `registry.datadoghq.com`. Defaults to "auto", which currently enables this behaviour for the AP1 datacenter (when `global.site: ap1.datadoghq.com`). More sites will be enabled in future releases. Set it to "" to disable. ([#2421](https://github.com/DataDog/helm-charts/pull/2421)).
+
+## 2.19.0-dev.6
+
 * Update Datadog Operator chart for RBACs for 1.24.0-rc.4.
 
 ## 2.19.0-dev.5
--- a/charts/datadog-operator/Chart.yaml
+++ b/charts/datadog-operator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: datadog-operator
-version: 2.19.0-dev.6
+version: 2.19.0-dev.7
 appVersion: 1.24.0-rc.4
 description: Datadog Operator
 keywords:
--- a/charts/datadog-operator/README.md
+++ b/charts/datadog-operator/README.md
@@ -1,6 +1,6 @@
 # Datadog Operator
 
-![Version: 2.19.0-dev.6](https://img.shields.io/badge/Version-2.19.0--dev.6-informational?style=flat-square) ![AppVersion: 1.24.0-rc.4](https://img.shields.io/badge/AppVersion-1.24.0--rc.4-informational?style=flat-square)
+![Version: 2.19.0-dev.7](https://img.shields.io/badge/Version-2.19.0--dev.7-informational?style=flat-square) ![AppVersion: 1.24.0-rc.4](https://img.shields.io/badge/AppVersion-1.24.0--rc.4-informational?style=flat-square)
 
 ## Values
 


charts/datadog-operator/values.yaml
```diff
@@ -102,6 +102,10 @@ datadogSLO:
 remoteConfiguration:
   # remoteConfiguration.enabled -- If true, enables Remote Configuration in the Datadog Operator (beta). Requires clusterName, API and App keys to be set.
   enabled: false
+# useDatadogRegistry -- If set, explicitly enables or disables registry.datadoghq.com as the default
+# container image registry for agent images managed by the operator. When not set, automatically enabled
+# for supported sites. Set to false to opt out.
+useDatadogRegistry:
```
AliDatadog: I am not sure about adding a new field in the values.yaml. There will be some confusion: `useDatadogRegistry` in the operator chart controls what registry the operator uses for agent images, not the operator's own image. 

Is it a temporary change ? If yes, IMO we should probably rename it / add a comment about it being temporary.
@levan-m what do you think ?

charts/datadog-operator/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.19.0
```
AliDatadog: As discussed it should be a dev version.

charts/datadog-operator/values.yaml
AliDatadog: Also, I think there might be some confusion:
- The default value is nil.
- The doc says it's a string
- It takes `true` or `false` as value (bool)
Not sure how yaml handles that

AliDatadog:
@codex review

AliDatadog:
@codex review

charts/datadog-operator/templates/_helpers.tpl
```diff
@@ -166,6 +166,13 @@ Return the appropriate apiVersion for PodDisruptionBudget policy APIs.
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the registry migration mode: "auto", "all", or "none".
+*/}}
+{{- define "datadog-registry-mode" -}}
+{{- default "auto" .Values.registryMigration.mode -}}
```
tbavelier: The default auto is not necessary since it's already the value set in `values.yaml`, so it will use it if not modified. If set to anything (including invalid value), it will use said defined value. I'd recommend using a helper with "real" value, e.g. (to double-check if it works, it's pseudo code):

```
  {{- define "datadog-registry-mode" -}}
  {{- $mode := .Values.registryMigration.mode  -}}
  {{- if and $mode (not (has $mode (list "auto" "all"))) -}}
    {{- fail (printf "registryMigration.mode must be \"auto\" or \"all\". Got: %q" $mode) -}}
  {{- end -}}
  {{- $mode -}}
  {{- end -}}
```
In the above example, if registryMigration.mode is ""/nil, it is not defined (so disabled). If it is a valid string, it is compared to auto/all to ensure a correct value is set (otherwise we silently "fail", e.g. mode = foo, we don't set the env vars)

charts/datadog-operator/values.yaml
```diff
@@ -102,6 +102,12 @@ datadogSLO:
 remoteConfiguration:
   # remoteConfiguration.enabled -- If true, enables Remote Configuration in the Datadog Operator (beta). Requires clusterName, API and App keys to be set.
   enabled: false
+registryMigration:
+  # registryMigration.mode -- Controls registry.datadoghq.com overrides for agent images.
+  # "auto" (default): enable overrides for sites where this is enabled by default.
+  # "all": enable all per-site overrides (DD_REGISTRY_OVERRIDE_*).
+  # "none": disable all overrides.
+  mode: "auto"
```
tbavelier:
```suggestion
registryMigration:
  # registryMigration.mode -- Controls registry.datadoghq.com overrides for agent images.
  # "auto" (default): enable overrides for sites where this is enabled by default.
  # "all": enable all per-site overrides (DD_REGISTRY_OVERRIDE_*).
  # "": disable all overrides.
  mode: "auto"
```
* Think it's a clearer/more common pattern to not set a registryMigration.mode instead of "explicitly" setting `none` (`--set registryMigration.mode=""`).

* Why are we using a nested key `registryMigration.mode` instead of directly `registryMigrationMode` ? Do we expect other children keys under `registryMigration` on top of `mode` ?

* I don't really like the `auto` naming scheme (maybe worth asking #documentation if they have ideas ?): I'm not sure if it's clear from a customer perspective this would be no-op for instance if they don't use `global.registry` set to Asia or w/e. Can we add 2-3 sentences explaining what this option does (more than "Controls registry.datadoghq.com overrides for agent images.") and how it could be no-op (reasoning being customers might remove it because of misunderstanding what it does)?
Also we would need to add the list of overrides when using `auto` (e.g. for this release, only Asia, but following releases will use more default) in the docstring

charts/datadog-operator/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.19.0-dev.6
+
+* Add registryMigration.mode setting with per-region DD_REGISTRY_OVERRIDE_* env vars, enable ASIA override by default ([#2421](https://github.com/DataDog/helm-charts/pull/2421)).
```
tbavelier: See comment further below on customer clarity, this is not clear IMHO

charts/datadog-operator/values.yaml
```diff
@@ -102,12 +102,17 @@ datadogSLO:
...
+# registryMigrationMode -- Controls gradual migration of agent image pulls from
+# third-party registries (GCR, ACR) to registry.datadoghq.com. When enabled, the
+# operator injects DD_REGISTRY_OVERRIDE_* env vars so that agent images are pulled
+# from registry.datadoghq.com for matching sites, unless the user has set a custom
+# registry. This has no effect on sites not covered by the active overrides.
+# More sites will be enabled by default in future helm-chart releases.
+# "auto" (default): enable overrides for sites where migration is rolled out.
+#   Currently enabled: ASIA region (ap1.datadoghq.com).
+# "all": enable all per-site overrides (ASIA, DEFAULT, EU, AZURE).
+# "" or unset: disable all overrides.
```
tbavelier:
```
# registryMigrationMode -- Controls gradual migration of agent image pulls from
# third-party registries (GCR, ACR) to registry.datadoghq.com. When enabled, the
# operator injects DD_REGISTRY_OVERRIDE_* env vars so that agent images are pulled
# from registry.datadoghq.com for matching sites, unless the user has set a custom
# registry. This has no effect on sites not covered by the active overrides.
# More sites will be enabled by default in future helm-chart releases.
# "auto" (default): enable overrides for sites where migration is rolled out.
#   Currently enabled: ASIA region (ap1.datadoghq.com).
# "all": enable all per-site overrides (ASIA, DEFAULT, EU, AZURE).
# "" or unset: disable all overrides.
```
Not sure why I can't quote/do in line suggestion
> Controls gradual migration of agent

Controls gradual migration of **A**gent

> When enabled, the operator injects DD_REGISTRY_OVERRIDE_* env vars so that agent images [...]

When enabled, DD_REGISTRY_OVERRIDE_* environment variables are added to the Datadog Operator deployment to pull Agent images from the global CDN-backed registry `registry.datadoghq.com` based on the setting `global.site` unless `global.registry` is specified in the `DatadogAgent` custom resource which takes precedence.
[...]

Feel free to adjust or w/e, but I think we can clarify further. Might be worth adding a note to https://docs.datadoghq.com/containers/kubernetes/installation/?tab=datadogoperator#container-registries and also mentioning ECR+ DockerHub on top of GCR and ACR ?

Also not sure about ASIA, DEFAULT, EU, AZURE. I'd recommend aligning with https://docs.datadoghq.com/getting_started/site/#access-the-datadog-site

charts/datadog-operator/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.19.0-dev.6
+
+* Add `registryMigrationMode` to control gradual migration of agent image pulls to registry.datadoghq.com. Defaults to "auto", which currently enables the ASIA region override (ap1.datadoghq.com). More sites will be enabled in future releases. Set to "" to disable. ([#2421](https://github.com/DataDog/helm-charts/pull/2421)).
```
tbavelier:
```suggestion
* Add `registryMigrationMode` to control gradual migration of Agent image pulls to `registry.datadoghq.com`. Defaults to "auto", which currently enables this behaviour for the AP1 datacenter (when `global.site: ap1.datadoghq.com`). More sites will be enabled in future releases. Set it to "" to disable. ([#2421](https://github.com/DataDog/helm-charts/pull/2421)).
```

tbavelier (approved):

PR #2471, opened 2026-03-11
Enable remote configuration by default on cluster check runners

piochelepiotr:
## Summary
Remote configuration was previously hardcoded to `false` on cluster check runners.
Some integration are using remote configuration: Kafka consumer integration for example, that is recommended to be setup as a Cluster check. So default behavior is no longer working well.

Add a new `clusterChecksRunner.remoteConfiguration.enabled` value (default `true`) so users can disable it if needed.

### DIFF
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: datadog
-version: 3.185.0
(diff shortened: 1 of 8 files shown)


chnn (approved):

gabedos (approved):
enabling rc on clc runners by default lgtm

PR #2395, opened 2026-02-18
[ACTP] Install the private action runner in the cluster agent via helm charts

merchristK:
#### What this PR does / why we need it:

This PR adds support for Private Action Runner (PAR) in the Datadog Cluster Agent, enabling users to execute workflow actions in their Kubernetes clusters.

**Key features:**
- **Self-enrollment mode**: Cluster Agent can automatically register itself with Datadog using API/APP keys and manage its identity in a Kubernetes secret
- **Manual credentials**: Support for providing URN and private key directly via values or existing secrets
- **Actions allowlist**: Configuration to restrict which actions the runner can execute
- **RBAC**: Automatic creation of Role and RoleBinding for secret management when PAR is enabled

**Changes:**
- Add `clusterAgent.privateActionRunner.*` configuration options in `values.yaml`
- Update cluster agent deployment template to inject PAR environment variables
- Create RBAC resources for PAR identity secret management
- Add comprehensive test coverage for all PAR configurations
- Update CHANGELOG.md for version 3.172.0

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
- fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] All commits are signed (see: [signing commits][1])
- [x] Chart Version semver bump label has been added (use `<chartName>/minor-version`, `<chartName>/patch-version`, or `<chartName>/no-version-bump`)
- [x] For `datadog` or `datadog-operator` chart or value changes, update the test baselines (run: `make update-test-baselines`)

GitHub CI takes care of the below, but are still required:
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] `CHANGELOG.md` has been updated
- [x] Variables are documented in the `README.md`

[1]: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

PR #2478, opened 2026-03-13
[ACTP] Add Private Action Runner support for node agent

martin-prevost1:
### Description

Deploy the Private Action Runner (PAR) as a standalone sidecar container in the node agent DaemonSet, matching the Datadog Operator approach https://github.com/DataDog/datadog-operator/pull/2659.
Cluster agent: https://github.com/DataDog/helm-charts/pull/2395

Instead of embedding PAR in the agent process via env vars, this PR:
- Adds a dedicated `private-action-runner` container running the standalone PAR binary (`/opt/datadog-agent/embedded/bin/privateactionrunner`)
- Creates a ConfigMap (`privateactionrunner.yaml`) with PAR configuration, mounted via the `-E` flag 
- Supports all enrollment modes: self-enrollment, manual credentials, and existing secrets
- Auto-enables remote configuration when PAR is enabled
- Adds a checksum annotation on the DaemonSet pod template so that PAR config changes trigger automatic pod rollouts

Configuration validation is delegated to the agent at runtime.

### Changes

| File | Description |
|------|-------------|
| `_container-private-action-runner.yaml` | New sidecar container template with standalone PAR binary |
| `private-action-runner-configmap.yaml` | ConfigMap with `privateactionrunner.yaml` config (no k8s secret fields) |
| `daemonset.yaml` | Conditionally includes PAR container, ConfigMap volume, and checksum annotation |
| `_helpers.tpl` | Auto-enable remote config when PAR is enabled; validation for manual mode credentials |
| `values.yaml` | Added `datadog.privateActionRunner` and `agents.containers.privateActionRunner` config |

### Test plan
Test performed using Datadog Agent version 7.77

**Docker Desktop Kubernetes**:
- [x] Self-enrollment: identity persisted to local file, Script connection auto-created, health check OK
- [x] `identityFromExistingSecret`: URN/privateKey injected via env vars from secret, health check OK
- [x] Inline credentials (urn + privateKey in ConfigMap): runner starts with correct URN, health check OK

**OpenShift**:
- [x] Self-enrollment: identity persisted to local file, Script connection auto-created, health check OK
- [x] `identityFromExistingSecret`: URN/privateKey injected via env vars from secret, health check OK
- [x] Inline credentials (urn + privateKey in ConfigMap): runner starts with correct URN, health check OK

**Unit tests** (19 tests passing):
- [x] `helm template` renders correctly for all scenarios (disabled, self-enroll, manual creds, existing secret, actions allowlist)
- [x] Self-enrollment works without leader election (node agent stores identity in local file)
- [x] Validation rejects manual mode without credentials

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 * Add Private Action Runner support in Node Agent as a sidecar container with configuration options for self-enrollment, manual credentials, and existing secrets.
 
+## 3.191.1
+
+* fix(gke_autopilot): Use pointerdir volume for GKE autopilot clusters ([#2495](https://github.com/DataDog/helm-charts/pull/2495)).
+
 ## 3.190.1
 
 * Add log volume to full host profiler ([#2461](https://github.com/DataDog/helm-charts/pull/2461)).
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -137,7 +137,7 @@
       value: {{  (default false (or .Values.datadog.logs.enabled .Values.datadog.logsEnabled)) | quote}}
     - name: DD_LOGS_CONFIG_CONTAINER_COLLECT_ALL
       value: {{  (default false (or .Values.datadog.logs.containerCollectAll .Values.datadog.logsConfigContainerCollectAll)) | quote}}
-    {{- if .Values.datadog.logs.enabled }} 
+    {{- if .Values.datadog.logs.enabled }}
     - name: DD_LOGS_CONFIG_K8S_CONTAINER_USE_FILE
       value: {{ .Values.datadog.logs.containerCollectUsingFiles | quote }}
     {{- end }}
@@ -364,6 +364,10 @@
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
     {{- end }}
+    {{- else if or .Values.providers.gke.autopilot .Values.providers.gke.gdc }}
+    - name: pointerdir
+      mountPath: /opt/datadog-agent/run
+      readOnly: false
     {{- else }}
     - name: datadogrun
       mountPath: /opt/datadog-agent/run
--- a/charts/datadog/templates/_daemonset-volumes-linux.yaml
+++ b/charts/datadog/templates/_daemonset-volumes-linux.yaml
@@ -227,7 +227,7 @@
   name: logdockercontainerpath
 {{- end }}
 {{- end }}
-{{- if or .Values.datadog.logs.enabled .Values.datadog.logsEnabled }}
+{{- if or .Values.datadog.logs.enabled .Values.datadog.logsEnabled .Values.providers.gke.autopilot .Values.providers.gke.gdc }}
 - hostPath:
     path: {{ template "datadog.hostMountRoot" . }}/logs
   name: pointerdir
--- a/test/datadog/baseline/manifests/gdc_daemonset_default.yaml
+++ b/test/datadog/baseline/manifests/gdc_daemonset_default.yaml
@@ -1471,7 +1471,8 @@ spec:
               name: auth-token
               readOnly: false
             - mountPath: /opt/datadog-agent/run
-              name: datadogrun
+              name: pointerdir
+              readOnly: false
             - mountPath: /certs
               name: kubelet-cert-volume
       initContainers:
@@ -1560,8 +1561,9 @@ spec:
           name: tmpdir
         - emptyDir: {}
           name: s6-run
-        - emptyDir: {}
-          name: datadogrun
+        - hostPath:
+            path: /var/datadog/logs
+          name: pointerdir
         - name: kubelet-cert-volume
           secret:
             secretName: datadog-kubelet-cert
--- a/test/datadog/baseline/manifests/gke_autopilot_allowlistedv2workload_default.yaml
+++ b/test/datadog/baseline/manifests/gke_autopilot_allowlistedv2workload_default.yaml
@@ -1521,7 +1521,8 @@ spec:
               name: passwd
               readOnly: true
             - mountPath: /opt/datadog-agent/run
-              name: datadogrun
+              name: pointerdir
+              readOnly: false
       initContainers:
         - args:
             - cp -r /etc/datadog-agent /opt
@@ -1636,8 +1637,9 @@ spec:
         - hostPath:
             path: /var/run/containerd
           name: runtimesocketdir
-        - emptyDir: {}
-          name: datadogrun
+        - hostPath:
+            path: /var/autopilot/addon/datadog/logs
+          name: pointerdir
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 10%
--- a/test/datadog/baseline/manifests/gke_autopilot_allowlistedv2workload_kubelet_apiserver.yaml
+++ b/test/datadog/baseline/manifests/gke_autopilot_allowlistedv2workload_kubelet_apiserver.yaml
@@ -1493,7 +1493,8 @@ spec:
               name: passwd
               readOnly: true
             - mountPath: /opt/datadog-agent/run
-              name: datadogrun
+              name: pointerdir
+              readOnly: false
       initContainers:
         - args:
             - cp -r /etc/datadog-agent /opt
@@ -1606,8 +1607,9 @@ spec:
         - hostPath:
             path: /var/run/containerd
           name: runtimesocketdir
-        - emptyDir: {}
-          name: datadogrun
+        - hostPath:
+            path: /var/autopilot/addon/datadog/logs
+          name: pointerdir
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 10%
--- a/test/datadog/baseline/manifests/gke_autopilot_npm.yaml
+++ b/test/datadog/baseline/manifests/gke_autopilot_npm.yaml
@@ -1754,7 +1754,8 @@ spec:
               name: passwd
               readOnly: true
             - mountPath: /opt/datadog-agent/run
-              name: datadogrun
+              name: pointerdir
+              readOnly: false
         - command:
             - process-agent
             - -config=/etc/datadog-agent/datadog.yaml
@@ -2123,8 +2124,9 @@ spec:
         - hostPath:
             path: /var/run/containerd
           name: runtimesocketdir
-        - emptyDir: {}
-          name: datadogrun
+        - hostPath:
+            path: /var/autopilot/addon/datadog/logs
+          name: pointerdir
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 10%
--- a/test/datadog/baseline/manifests/gke_autopilot_system_probe.yaml
+++ b/test/datadog/baseline/manifests/gke_autopilot_system_probe.yaml
@@ -1754,7 +1754,8 @@ spec:
               name: passwd
               readOnly: true
             - mountPath: /opt/datadog-agent/run
-              name: datadogrun
+              name: pointerdir
+              readOnly: false
         - command:
             - process-agent
             - -config=/etc/datadog-agent/datadog.yaml
@@ -2187,8 +2188,9 @@ spec:
         - hostPath:
             path: /var/run/containerd
           name: runtimesocketdir
-        - emptyDir: {}
-          name: datadogrun
+        - hostPath:
+            path: /var/autopilot/addon/datadog/logs
+          name: pointerdir
   updateStrategy:
     rollingUpdate:
       maxUnavailable: 10%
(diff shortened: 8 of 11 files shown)


charts/datadog/templates/daemonset.yaml
```diff
@@ -187,6 +190,11 @@ spec:
           name: {{ include "agents-install-info-configmap-name" . }}
       - name: config
         emptyDir: {}
+      {{- if .Values.datadog.privateActionRunner.enabled }}
+      - name: {{ template "datadog.fullname" . }}-privateactionrunner-config
+        configMap:
+          name: {{ template "datadog.fullname" . }}-privateactionrunner
```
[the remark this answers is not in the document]
merchristK: this is true. can you update the checksum @martin-prevost1 ?

charts/datadog/templates/private-action-runner-configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+data:
+  privateactionrunner.yaml: |
+    private_action_runner:
+      enabled: true
+      {{- if .Values.datadog.privateActionRunner.selfEnroll }}
+      self_enroll: true
+      identity_use_k8s_secret: true
```
merchristK: this is a config only needed for DCA. Node agent doesn't need that.

charts/datadog/templates/private-action-runner-configmap.yaml
```diff
@@ -0,0 +1,36 @@
...
+      identity_secret_name: {{ .Values.datadog.privateActionRunner.identitySecretName | quote }}
```
merchristK: same. this is not used by the node agent at all. Instead we self enrol and save in a file https://github.com/DataDog/datadog-agent/blob/2cab6a680057ee033dd9b3ec50fc7c22082c59dc/pkg/config/setup/privateactionrunner.go#L20

So IMO, if users want to reuse an existing secret, it should be like:
- helm values updated to specify the PAR identity (either urn or from a secret)
- we update the envvar with the par urn / private key

If this is not the case, users should just add `self_enroll: true` and that's it. The runner will be self enrolled and become a zombie when the agent is terminated.

charts/datadog/templates/rbac.yaml
```diff
@@ -237,4 +237,36 @@ roleRef:
...
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ template "datadog.fullname" . }}-node-private-action-runner
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "agents.serviceAccountName" . }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
```
merchristK: none of these are needed. Only makes sense for the DCA

merchristK (changes requested):
Thanks for the changes. I believe the missing piece is the flow for self enrollment. What to keep in mind is that storing PAR identity in a k8s secret is only done for the DCA.
For the node agent we only use this https://github.com/DataDog/datadog-agent/blob/2cab6a680057ee033dd9b3ec50fc7c22082c59dc/pkg/config/setup/privateactionrunner.go#L20

The flow is:
- we try to get private_key, urn from `config.private_action_runner.identity_file_path`
- if it exists, we use it
- if not, we check if private key and urn are directly provided in the config (`config.private_action_runner.private_key` and `config.private_action_runner.urn`)
- if they are provided, we use it, if not we check if self_enroll is enabled
- if enabled, we enroll the runner and store the private_key and urn in the config
- if not enabled, we fail and the runner doesn't start.

Check the full flow here https://github.com/DataDog/datadog-agent/blob/2cab6a680057ee033dd9b3ec50fc7c22082c59dc/comp/privateactionrunner/impl/privateactionrunner.go#L134-L163


I believe the helm chart installation should reflect that.

charts/datadog/templates/daemonset.yaml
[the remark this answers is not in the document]
martin-prevost1: I’m adding a checksum.

martin-prevost1:
> Thanks for the changes. I believe the missing piece is the flow for self enrollment. What to keep in mind is that storing PAR identity in a k8s secret is only done for the DCA. For the node agent we only use this https://github.com/DataDog/datadog-agent/blob/2cab6a680057ee033dd9b3ec50fc7c22082c59dc/pkg/config/setup/privateactionrunner.go#L20
> 
> The flow is:
> 
> * we try to get private_key, urn from `config.private_action_runner.identity_file_path`
> * if it exists, we use it
> * if not, we check if private key and urn are directly provided in the config (`config.private_action_runner.private_key` and `config.private_action_runner.urn`)
> * if they are provided, we use it, if not we check if self_enroll is enabled
> * if enabled, we enroll the runner and store the private_key and urn in the config
> * if not enabled, we fail and the runner doesn't start.
> 
> Check the full flow here https://github.com/DataDog/datadog-agent/blob/2cab6a680057ee033dd9b3ec50fc7c22082c59dc/comp/privateactionrunner/impl/privateactionrunner.go#L134-L163
> 
> I believe the helm chart installation should reflect that.

Thanks for the detailed review. I made the changes and tested everything again (see the test section in the PR description).

charts/datadog/templates/_container-private-action-runner.yaml
```diff
@@ -0,0 +1,87 @@
...
+{{- end }}
+{{- end }}
+  env:
+    {{- include "containers-common-env" . | nindent 4 }}
+    {{- include "containers-cluster-agent-env" . | nindent 4 }}
+    {{- include "fips-envvar" . | nindent 4 }}
+    - name: DD_LOG_LEVEL
+      value: {{ .Values.agents.containers.privateActionRunner.logLevel | default .Values.datadog.logLevel | quote }}
+    {{- if or .Values.datadog.appKey .Values.datadog.appKeyExistingSecret }}
+    - name: DD_APP_KEY
```
merchristK, line 26: do we have an api key as well?

charts/datadog/templates/_container-private-action-runner.yaml
```diff
@@ -0,0 +1,87 @@
...
+      valueFrom:
+        secretKeyRef:
+          name: {{ template "datadog.appKeySecretName" . }}
+          key: app-key
+    {{- end }}
+    {{- if and (not .Values.datadog.privateActionRunner.selfEnroll) .Values.datadog.privateActionRunner.identityFromExistingSecret }}
```
merchristK, line 32: nit: can we rename to `identitySecretName`? to have consistency with `appKeySecretName`. We would need to change the dca config as well. Fine to do as a follow up.

charts/datadog/templates/_helpers.tpl
```diff
@@ -1323,6 +1323,19 @@ Validate Private Action Runner configuration
...
 {{- end -}}
 
+{{/*
+Validate Node Agent Private Action Runner configuration
+*/}}
+{{- define "validate-node-private-action-runner-config" -}}
+{{- if .Values.datadog.privateActionRunner.enabled -}}
+{{- if not .Values.datadog.privateActionRunner.selfEnroll -}}
+{{- if and (not .Values.datadog.privateActionRunner.identityFromExistingSecret) (or (not .Values.datadog.privateActionRunner.urn) (not .Values.datadog.privateActionRunner.privateKey)) -}}
+{{- fail "Node Agent Private Action Runner: when selfEnroll is disabled, you must provide either datadog.privateActionRunner.identityFromExistingSecret or both datadog.privateActionRunner.urn and datadog.privateActionRunner.privateKey" }}
```
merchristK, line 1333: nit: Can we also validate that if `identityFromExistingSecret` we shouldn't provide `datadog.privateActionRunner.urn` and `datadog.privateActionRunner.privateKey`?

It doesn't really matter since the first has precedence but maybe log a warn? 🤷

charts/datadog/templates/private-action-runner-configmap.yaml
```diff
@@ -0,0 +1,34 @@
...
+  namespace: {{ .Release.Namespace }}
+  labels:
+    helm.sh/chart: "{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" }}"
+    app.kubernetes.io/name: "{{ template "datadog.fullname" . }}"
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+data:
+  privateactionrunner.yaml: |
+    private_action_runner:
```
merchristK, line 15: this will be hard to scale. It means anytime we change the config (from dd-agent), we will also need to update the helm chart. I think we should consider extracting some required fields like `self_enroll` .. and then expose a base config that we will take as is.

NOTE: we need to do this for the DCA also so fine to do in a follow up

charts/datadog/templates/_container-private-action-runner.yaml
martin-prevost1: Yes, the API key is already injected via 
https://github.com/DataDog/helm-charts/blob/bbe8a23181ba3507d12ca356e944fd8ee093492c/charts/datadog/templates/_containers-common-env.yaml#L4-L8
This helper includes DD_API_KEY for all Agent containers. Only the APP key requires an explicit addition, as it is not included by default in the Node Agent containers.

charts/datadog/templates/_container-private-action-runner.yaml
martin-prevost1: I will do it in another PR.

charts/datadog/templates/_helpers.tpl
martin-prevost1: I like the idea of a warning. I will add it.

fanny-jiang:
@codex review

charts/datadog/templates/daemonset.yaml
```diff
@@ -167,6 +170,9 @@ spec:
         {{- if eq (include "should-enable-data-plane" .) "true" }}
           {{- include "container-agent-data-plane" . | nindent 6 }}
         {{- end }}
+        {{- if .Values.datadog.privateActionRunner.enabled }}
+          {{- include "container-private-action-runner" . | nindent 6 }}
```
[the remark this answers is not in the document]
martin-prevost1: The PAR sidecar follows the same pattern as other opt-in sidecars in the DaemonSet (security-agent, otel-agent, host-profiler, data-plane)

charts/datadog/templates/_container-private-action-runner.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - name: logdatadog
+      mountPath: {{ template "datadog.logDirectoryPath" . }}
+      readOnly: false
+    {{- if (not .Values.providers.gke.autopilot) }}
+    - name: auth-token
+      mountPath: {{ template "datadog.confPath" . }}/auth
+      readOnly: false
+    {{- end }}
+    - name: {{ template "datadog.fullname" . }}-privateactionrunner-config
+      mountPath: /etc/privateactionrunner
```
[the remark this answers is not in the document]
martin-prevost1: Windows containers in Kubernetes are not supported for the Private Action Runner at this time.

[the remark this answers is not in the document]
fanny-jiang: The private action runner container still gets rendered in the daemonset when I set `targetSystem=windows` and `datadog.privateActionRunner.enabled=true`. I'd expect that the private action runner container and its associated volumes shouldn't get rendered when `targetSystem` is set to `windows`.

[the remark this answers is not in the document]
martin-prevost1: I agree with you, and I made the change. I also added a warning in NOTES.txt.

charts/datadog/templates/_container-private-action-runner.yaml
```diff
@@ -0,0 +1,87 @@
...
+{{- include "validate-node-private-action-runner-config" . }}
+- name: private-action-runner
+  image: "{{ include "image-path" (dict "root" .Values "image" .Values.agents.image) }}"
+  imagePullPolicy: {{ .Values.agents.image.pullPolicy }}
+  {{- if eq .Values.targetSystem "linux" }}
+  command: ["/opt/datadog-agent/embedded/bin/privateactionrunner", "run", "-c={{ template "datadog.confPath" . }}", "-E=/etc/privateactionrunner/privateactionrunner.yaml"]
+  {{- end -}}
+  {{- if eq .Values.targetSystem "windows" }}
+  command: ["privateactionrunner", "run", "-c={{ template "datadog.confPath" . }}", "-E=/etc/privateactionrunner/privateactionrunner.yaml"]
+  {{- end -}}
```
fanny-jiang: If PAR is not supported for windows in kubernetes, then we could leave this out, right?

PR #2479, opened 2026-03-13
Add system-probe-lite support for discovery in Helm chart

vitkyrka:

Adds support for starting [system-probe-lite](https://datadoghq.atlassian.net/wiki/spaces/DSCVR/pages/6332384226/Starting+system-probe-lite+and+system-probe) (SPL) instead of the full system-probe in the Helm chart when only discovery is enabled.

SPL is a privileged Rust binary that implements just the discovery module. The goal of this support is to ensure that SPL is started instead of system-probe whenever only discovery is enabled.

**How it works:**

If any other system-probe features than discovery are also enabled (NPM, USM, etc.), the regular `system-probe` binary is always used directly

When only discovery is enabled (no other system-probe features):
 - The system-probe container automatically runs SPL as the entry point
 - If SPL fails to start (either because it doesn't exist in the image or potentially due to some other issue):
   -  If `discovery.enabled` was explicitly set by the user, we fall back to the full `system-probe` binary (since the user opted in to discovery knowingly)
    - If only `discovery.enabledByDefault` is set (i.e. discovery was turned on without the user's explicit choice), we fall back to `sleep infinity` to avoid starting system-probe unexpectedly (or crashing) when using an older agent image without `system-probe-lite` (or without the discovery feature altogether).

The `discovery-enabled` helper uses `kindIs "invalid"` to distinguish between `enabled: false` (explicitly disabled) and `enabled` being unset/nil. This is important because a simple `or` would treat `false` and nil the same way, causing `enabledByDefault: true` to override an explicit `enabled: false`.

Note that discovery is not turned on by default in this PR, that's a separate, future step.

#### Checklist


vitkyrka:
@codex review

vitkyrka:
@codex review

charts/datadog/templates/_helpers.tpl
```diff
@@ -590,11 +590,60 @@ Return the image for the otel-agent in gateway based on `.Values` (passed as .)
...
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if system-probe-lite should be used instead of full system-probe.
+This is the case when discovery is enabled and no other system-probe feature
+requires the full binary.
+*/}}
+{{- define "should-use-system-probe-lite" -}}
+{{- if and (eq (include "discovery-enabled" .) "true") (eq (include "system-probe-feature-except-discovery" .) "false") -}}
```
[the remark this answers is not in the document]
vitkyrka: Should be fixed now

gjulianm (approved):

test/datadog/baseline/manifests/gke_autopilot_usm.yaml
```diff
@@ -1753,8 +1753,9 @@ spec:
               name: pointerdir
               readOnly: false
         - command:
-            - system-probe
-            - --config=/etc/datadog-agent/system-probe.yaml
+            - /bin/sh
+            - -c
+            - system-probe-lite run --socket /var/run/sysprobe/sysprobe.sock --log-level ${DD_LOG_LEVEL:-info} || system-probe --config=/etc/datadog-agent/system-probe.yaml
```
fanny-jiang: This diff indicates that the changes will break the system-probe container command in GKE autopilot when Service Discovery is enabled: https://github.com/DataDog/helm-charts/blob/main/test/datadog/baseline/values/gke_autopilot_usm.yaml

Can you please gate the new command from getting enabled when `providers.gke.autopilot` or `providers.gke.gdc` are enabled until the GKE Autopilot Datadog WorkloadAllowlist can be updated to support system-probe-lite?

test/datadog/gke_autopilot_workloadallowlist_test.go
```diff
@@ -96,10 +96,10 @@ func Test_autopilotWorkloadAllowlistConfigs(t *testing.T) {
...
 				Values:      []string{"../../charts/datadog/values.yaml"},
 				Overrides: map[string]string{
-					"datadog.envDict.HELM_FORCE_RENDER":                 "true",
-					"datadog.apiKeyExistingSecret":                      "datadog-secret",
-					"datadog.appKeyExistingSecret":                      "datadog-secret",
-					"providers.gke.autopilot":                           "true",
+					"datadog.envDict.HELM_FORCE_RENDER":        "true",
+					"datadog.apiKeyExistingSecret":             "datadog-secret",
+					"datadog.appKeyExistingSecret":             "datadog-secret",
+					"providers.gke.autopilot":                  "true",
```
fanny-jiang, line 103: Let's also add `datadog.service.enabled` to the overrides in this autopilot go test, as well as a new test case for the system-probe container's expected command in the `verifyAutopilotWorkloadAllowlistConstraints() test func`. 

```suggestion
                    "datadog.service.enabled":                  "true",
					"providers.gke.autopilot":                  "true",
```

https://github.com/DataDog/helm-charts/pull/2479/changes#diff-a99ee983a512a9c1a67ee68b4c3ad260e2bbe1182dd0c2cfbfc1f6ed379b65a8R146

charts/datadog/templates/_helpers.tpl
```diff
@@ -590,11 +590,60 @@ Return the image for the otel-agent in gateway based on `.Values` (passed as .)
...
+true
+{{- else -}}
+false
+{{- end -}}
+{{- end -}}
```
fanny-jiang, line 624: Gate the breaking change for autopilot:

```suggestion
{{- define "should-use-system-probe-lite" -}}
{{- if and (eq (include "discovery-enabled" .) "true") (eq (include "system-probe-feature-except-discovery" .) "false")
 (not (or .Values.providers.gke.autopilot .Values.providers.gke.gdc)) -}}                                             
true                                                                                                                   
{{- else -}}                                                                                                           
false                                                                                                                  
{{- end -}}                                                                                                           
{{- end -}}
```

fanny-jiang (changes requested):
LGTM, but the changes as they are now are breaking in GKE Autopilot due to the restricted container, `command` value, being modified without a corresponding update to the Datadog GKE Autopilot WorkloadAllowlist. 

https://docs.cloud.google.com/kubernetes-engine/docs/reference/crds/workloadallowlist#containers (see `containers[].command`)

Contributions to update the WorkloadAllowlist are welcome! We have internal docs on how updates can be submitted. Until then, the changes should be gated for both GKE Autopilot and GKE GDC. Thanks!

Yumasi (approved):

wdhif (changes requested):
Blocking until we validate https://github.com/DataDog/datadog-operator/pull/2610

fanny-jiang (approved):

PR #2262, opened 2025-12-22
[CWS] add securityAgent.runtime.enforcement paramter

safchain:
#### What this PR does / why we need it:

Expose the datadog.securityAgent.runtime.enforcement.enabled parameter and adjust the capabilities and seccomp profile accordingly. Add `KILL` capability in case of enabled.

#### Which issue this PR fixes
*(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close that issue when PR gets merged)*
  - fixes #

#### Checklist
[Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.]
- [x] Chart Version semver bump label added (use `<chartName>/minor-version`, `<chartName>/patch-version`, or `<chartName>/no-version-bump`)
- [x] For `datadog` or `datadog-operator` chart or value changes, update the test baselines (run: `make update-test-baselines`)

GitHub CI takes care of the below, but are still required:
- [x] Documentation has been updated with helm-docs (run: `.github/helm-docs.sh`)
- [x] `CHANGELOG.md` has been updated 
- [x] Variables are documented in the `README.md`

PR #2486, opened 2026-03-17
fix(datadog): gate KILL capability WorkloadAllowlist violation for GKE Autopilot and GDC

fanny-jiang:
## What this PR does / why we need it

PR #2262 (CWS enforcement) added the `KILL` Linux capability and the `kill` syscall to `system-probe`'s seccomp profile when `datadog.securityAgent.runtime.enforcement.enabled=true`. Since this option **defaults to `true`**, any Helm install that enables a system-probe feature (e.g. `networkMonitoring.enabled=true`) will include `KILL` in `system-probe`'s capabilities. This capability is not present in WorkloadAllowlist v1.0.3, causing the Warden admission webhook to reject the DaemonSet on GKE Autopilot clusters:

```
WorkloadAllowlist Mismatch:
Container[2] (system-probe): SecurityContext: Capabilities: Add Mismatch: KILL missing from allowlist.
```

**Fix:** gate both the `KILL` capability and the `kill` seccomp syscall on `securityAgent.runtime.enabled`. Since enforcement is a CWS sub-feature that has no effect without the CWS runtime, and `runtime.enabled=true` is already blocked on GKE Autopilot via a NOTES.txt validation error, this gate disables enforcement end-to-end on Autopilot.

The fix touches two templates:
- `_container-system-probe.yaml` — `KILL` capability now only added when `runtime.enabled=true` and `enforcement.enabled=true`
- `system-probe-configmap.yaml` — `enforcement.enabled` config field and `kill` seccomp syscall follow the same gate

### Testing

- Manually validated before/after on GKE Autopilot sandbox with `networkMonitoring.enabled=true`:
  - Before: Warden rejects with `KILL` mismatch
  - After: `STATUS: deployed`, agent pods `3/3 Running`, `KILL` absent from capabilities
- Unit tests pass: `make unit-test-datadog`

## Checklist
- [x] `datadog/patch-version`
- [x] Baseline manifests updated (`make update-test-baselines`)

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,8 +1,16 @@
 # Datadog changelog
 
-## 3.193.1
+## 3.195.1
 
-* Fix GKE Autopilot and GDC installs: gate `KILL` capability on `system-probe` when `securityAgent.runtime.enforcement.enabled=true`, as it is not included in the Datadog WorkloadAllowlist v1.0.3.
+* Gate `KILL` capability on `system-probe` when `securityAgent.runtime.enforcement.enabled=true` and `securityAgent.runtime.enabled=true`.
+
+## 3.195.0
+
+* Extend `registryMigrationMode: "auto"` to US1 (`datadoghq.com`) users with APM disabled (the default). If you experience image pull issues, set `registryMigrationMode: ""` to revert to the previous registry.
+
+## 3.194.0
+
+* [CONTP-1361] add admission controller probe configuration ([#2449](https://github.com/DataDog/helm-charts/pull/2449)).
 
 ## 3.193.0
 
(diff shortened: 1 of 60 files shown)


charts/datadog/templates/_container-agent.yaml
```diff
@@ -364,7 +364,7 @@
       mountPropagation: {{ .Values.datadog.hostVolumeMountPropagation }}
       readOnly: true
     {{- end }}
-    {{- else }}
+    {{- else if not (or .Values.providers.gke.autopilot .Values.providers.gke.gdc) }}
     - name: datadogrun
       mountPath: /opt/datadog-agent/run
```
Mathew-Estafanous, line 373: `/opt/datadog-agent/run` also needs to be writable for other features like runtime security profiles and sockets for communication between the agent and system probe. Is there any way we could instead add `datadogrun` to the allowlist?

PR #2488, opened 2026-03-17
[INPLAT-1004]: expose registry allow list for DatadogLibrary volumes

knusbaum:
## Summary
- Adds `apm.registryAllowList` to the `datadog-csi-driver` Helm chart
- When set, the value is passed to the driver container as `DD_REGISTRY_ALLOW_LIST` (comma-joined), restricting which container registries can be used as sources for DatadogLibrary CSI volumes
- Defaults to empty list (all registries allowed, preserving backward compatibility)
- Bumps chart version to `0.10.0`

## Test plan
- [x] `TestRegistryAllowList` verifies `DD_REGISTRY_ALLOW_LIST` is set correctly when `registryAllowList` is non-empty
- [x] `TestRegistryAllowList` verifies `DD_REGISTRY_ALLOW_LIST` is absent when `registryAllowList` is explicitly set to `[]` (via values file)
- [x] `TestRegistryAllowList` verifies `DD_REGISTRY_ALLOW_LIST` is absent when `registryAllowList` is undefined (no values file)
- [x] All existing baseline tests continue to pass

Closes INPLAT-1004


charts/datadog-csi-driver/values.yaml
```diff
@@ -66,6 +66,13 @@ apm:
   # apm.enabled -- Enable APM/SSI support for the CSI driver.
   enabled: true
 
+  # apm.registryAllowList -- List of allowed container registries for DatadogLibrary CSI volumes.
+  # When empty, all registries are allowed (default). When non-empty, only the listed registries
+  # are permitted as sources for instrumentation libraries.
+  registryAllowList: []
```
knusbaum: Initially, I was just going to do this in the CSI driver, but I realized it's better to do it both there and the admission controller.

I'm not super convinced that this is the right place for registryAllowList, given that it affects both the csi driver and the admission controller, but I'm not sure if there's a better place.

adel121:
@iamluc seems related to what you are working on ?

Could you please have a look 🙇

iamluc (approved):
LGTM!

adel121:
I think [charts/datadog/charts/datadog-csi-driver-0.10.0.tgz‎](https://github.com/DataDog/helm-charts/pull/2488/changes#diff-46ceb6d063d3f63904c643249cda6c5a5a8860bafdfec33d4f6ec46f0cfa6311) was committed by mistake ?

charts/datadog/values.yaml
```diff
@@ -5,6 +5,14 @@
...
 ## ONLY SET THE VALUES YOU WANT TO OVERRIDE IN YOUR values.yaml.
 
+# global.containerRegistryAllowList -- Restrict which registries can be used for APM library injection.
+## When non-empty, only libraries from the listed registries will be injected. Enforced by both the
+## admission controller webhook and the CSI driver. An empty list allows all registries (default).
+global:
+  containerRegistryAllowList: []
+  #   - public.ecr.aws/datadog
+  #   - gcr.io/datadoghq
+
```
adel121, line 15: Does the allowlist get enforced only on CSI volumes? or on all autoinstrumentation ?

iamluc: All autoinstrumentation (init containers, CSI volumes, image volumes)

adel121 (approved):

tedkahwaji (approved):

charts/datadog/values.yaml
adel121: I see that this is already approved. 

But to me the name is confusing. It shows no mention of APM at all (and it is not in the autoinstrumentation section), and doesn't explain what features this allowlist applies to. 

I would suggest to change the name to something more expressive (`apmSSIContainerRegistryAllowlist` or `apmRegistryAllowlist`), or at least add a comment to make it more explicit what this configuration option does.

Worth considering that we have similar configurations:
1- [Admission controller level container registry config option](https://github.com/DataDog/helm-charts/blob/87138559b85ca0310d9f8fe55999cb9bfd92e45d/charts/datadog/values.yaml#L1654-L1656)
2- [Webhook level override (exists for agent sidecar webhook only)](https://github.com/DataDog/helm-charts/blob/87138559b85ca0310d9f8fe55999cb9bfd92e45d/charts/datadog/values.yaml#L1714-L1715)

So to me, the option being added in this PR, at least currently, makes sense only for APM webhook.

Ideally, I would have preferred that the config option lives inside autoinstrumentation section directly, but that will risk inconsistent config (we may mitigate this by adding a comment above the option saying something like `to enforce this allowlist on the CSI driver level, append [this] config at the end`, but this is not perfect either). So I leave it to you to decide what's the best for this feature's config.

But in all cases, if we decide to keep it global, I would strongly recommend making the intent explicit so that the user instantly know what this does without any confusion.

iamluc: Even if it's not perfect, my view is that having a single global is safer than having two that could become inconsistent.
That said, I've renamed the config to `apmRegistryAllowlist` as you suggested

adel121: This looks much better. 

Thanks!

PR #2586, opened 2026-04-21
[PROF-14068] Remove privileges for host-profiler

theomagellan:
#### What this PR does / why we need it

This PR removes the Host Profiler's needs for `privileges: true` by only allowing strictly minimal capabilities and restricting available syscalls.

#### Special notes for your reviewer
- Capabilities: `SYS_ADMIN, SYS_PTRACE, SYS_RESOURCE, DAC_READ_SEARCH, SYSLOG`
- Seccomp: allowlist with `SCMP_ACT_ERRNO` default. `kill` restricted to signal 0 via arg filter. `setns` excluded.
- AppArmor: wired into `generate-security-context` (defaults to `unconfined`, matching `system-probe`)
- Egress: added to the FQDN allow-list for symbol uploads and OTLP metrics
  -  `sourcemap-intake.%s`: symbol intake
  - `otlp.%s`: OTLP metrics intake

#### Test plan
- Deployed on EKS 1.35 (kernel 6.17) with seccomp + AppArmor localhost profiles enforced
- `--go-runtime-metrics` OTLP self-loop works

Profiles using the chart can be found [here](https://ddstaging.datadoghq.com/profiling/explorer?query=datadog.host.name%3Ai-07c1951615c0817e3%20OR%20datadog.host.name%3Ai-09516ab7ee1385240&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=false&my_code=enabled&refresh_mode=sliding&viz=stream&from_ts=1776846237443&to_ts=1776849837443&live=true) for both supported architectures.

**Note**: self profiles (the profiler profiling its own process) are currently broken since non-privileged containers with hostPID get their own cgroup namespace and `container.id` extraction fails. The cluster used for testing is using a custom image with a hotfix for this.

#### Checklist


charts/datadog/templates/_container-host-profiler.yaml
r1viollet: why not read only ?

r1viollet:
We discussed a couple of points here.
It might be worth trying out CAP_BPF + CAP_PERFMON instead of SYS_ADMIN, we can follow up on this later.
Endpoints look good.
There is follow up work on the UIDs and child processes, so we might not need some of the UID syscalls (`setuid/setgid/setresuid/setresgid/setgroups`).

r1viollet:
The PR needs to be aligned

r1viollet (approved):
LGTM, some minor improvement proposals
This is a step in the correct direction (even if we will no longer be able to do things like gdb / ptrace etc 😄 )

charts/datadog/templates/_container-host-profiler.yaml
theomagellan: Hm, I didn't really challenge the existing deployment configuration.
`cilium/ebpf` seems to rely on writing to `tracefs` as fallback to old kernel limitations (cf. [here](https://github.com/cilium/ebpf/blob/9f87aafaeb37c579ad1e3f5729462d8540200504/link/kprobe.go#L294-L299)).

The 5.10 kernel's [`alloc_trace_kprobe`](https://github.com/torvalds/linux/blob/2c85ebc57b3e1817b6ce1a6b703928e113a90442/kernel/trace/trace_kprobe.c#L256) is not able to return `-EINVAL` based on symbol name, so we should be safe even if the fallback still exists in current `cilium/ebpf` versions.
Overall I would lean towards keeping it writeable in case something changes or odd behaviors occur somehow, but no strong opinions. I can make it read only if you feel strongly about this

theomagellan:
> It might be worth trying out CAP_BPF + CAP_PERFMON instead of SYS_ADMIN, we can follow up on this later.


Yes since upstream [now has kernel 5.10 as minimum](https://github.com/DataDog/opentelemetry-ebpf-profiler#supported-linux-kernel-version). While testing though I found that we also needed `CAP_CHECKPOINT_RESTORE` to be able to read the contents in `/proc/pid/map_files` for other processes
Implemented in [78cd53a](https://github.com/DataDog/helm-charts/pull/2586/commits/78cd53a2e771d0fa1631d0672a3b73571364bd2e)[78cd53a](https://github.com/DataDog/helm-charts/pull/2586/commits/78cd53a2e771d0fa1631d0672a3b73571364bd2e) and profiles can be found [here](https://ddstaging.datadoghq.com/profiling/explorer?query=%28host%3Ai-09516ab7ee1385240%20OR%20host%3Ai-07c1951615c0817e3%29&agg_m=%40prof_ebpf_cpu_cores&agg_m_source=base&agg_t=sum&color_by=total&extra_search_fields=%7B%22filters_query%22%3A%22%22%2C%22sample_type%22%3A%22cpu-time%22%7D&fromUser=true&my_code=enabled&refresh_mode=paused&viz=flame_graph&from_ts=1777997555294&to_ts=1778001155294&live=false)
I'm re-requesting a review from you @r1viollet since I changed the capability set 🙇 sorry for the trouble!


> There is follow up work on the UIDs and child processes, so we might not need some of the UID syscalls (`setuid/setgid/setresuid/setresgid/setgroups`).


I looked into this and it's still pretty unclear to me if they are needed at all. Earlier today the container would fail without these but now it seems to be running fine in my testing environment. The only changes made were the capabilities, but I don't think this should change much 🤔
Caught this in a jira ticket for investigation https://datadoghq.atlassian.net/browse/PROF-14540

charts/datadog/templates/_container-host-profiler.yaml
theomagellan: Actually, deployments in `datadog-operator` seem to be [mounting it in read only](https://github.com/DataDog/datadog-operator/blob/59f9e3d84127430ecc0c9a1b05ae5b0cf7a36541/internal/controller/datadogagent/feature/hostprofiler/feature.go#L96-L100) so let's align to that as well. I'll comment previous findings in the yaml so we can keep an eye on it just in case.

charts/datadog/values.yaml
```diff
@@ -2395,7 +2404,16 @@ agents:
       # agents.containers.hostProfiler.securityContext -- Allows you to overwrite the default container SecurityContext for the host-profiler container.
       securityContext:
         readOnlyRootFilesystem: true
-        privileged: true
+        privileged: false
+        capabilities:
```
r1viollet, line 2408: this is great, thanks!

r1viollet (approved):
LGTM

theomagellan:
hey @gpalmz 👋 this is ready for a review from your team! Thank you 🙇

swang392 (approved):

PR #2589, opened 2026-04-22
[CONS-8251]  Delete unused resources when node agent or cluster agent is disabled

aquiladayc:
Raised in CONS-8251
Some resources are not in use when node agent or cluster agent is disabled.

- Install info (ConfigMap)
- kpi telemetry config 
- ksm rbac


aquiladayc:
See: https://datadoghq.atlassian.net/browse/CONS-8251?focusedCommentId=3186924

PR #2598, opened 2026-04-24
enable discovery by default on supported agent versions

Yumasi:

This PR enables Service Discovery by default on agents >=7.78, which will use the new System-Probe-Lite binary for low-resource overhead. The version is determined based on the image tag. We treat every unparseable tag as "latest". This is to be consistent with the [operator behaviour](https://github.com/DataDog/helm-charts/compare/guillaume.pagnoux/service-discovery-defaulting?expand=1).

The change makes the distinction between default enablement, and an explicit one. This allows users of an agent <7.78 to still enable discovery if they truly wants it, via system-probe (much higher overhead).

DSCVR-396

This PR uses the fallback mechanism of System-Probe to decide to start System-Probe-Lite. This is done in order to avoid a drift between the Helm chart & the agent in the handling of the fallback decision (i.e we would have to maintain a list of system-probe feature which could be easily missed when adding new features. Missing it could cause wrong behaviors like not enabling system-probe when we actually should for another feature, for example).

Using system-probe to control the fallback decision, like in bare-Linux & Docker installations, avoid this sort of issues, but this means that we will see a spike in memory consumption due to system-probe's size itself (~70MiB). Once the fallback to System-Probe-Lite happens, the memory use is expected to fallback to ~3.4MiB (measured in dogfooding).

Fallback behaviour + system-probe container creation tested with Kind and the following cases:

<table>
<tr>
 <td><b>Agent version</b>
 <td><b>Config scenario</b>
 <td><b>system-probe<br>container status</b>
 <td><b>Running binary</b>
<tr>
 <td>master-py3
 <td>Default configuration, non-semver tag fallback
 <td>Present
 <td>System-Probe-Lite
<tr>
 <td>latest
 <td>Default configuration, chart <code>get-agent-version</code> policy
 <td>Present
 <td>System-Probe-Lite
<tr>
 <td>7.78
 <td>Default configuration
 <td>Present
 <td>System-Probe-Lite
<tr>
 <td>7.77
 <td>Default configuration
 <td>Absent
 <td>N/A
<tr>
 <td>7.78
 <td>Discovery explicitly disabled
 <td>Absent
 <td>N/A
<tr>
 <td>7.77
 <td>Discovery explicitly enabled
 <td>Present
 <td>System-Probe
<tr>
 <td>7.78
 <td>Other system-probe feature enabled
 <td>Present
 <td>System-Probe
</table>

#### Checklist


Yumasi:
@codex review

charts/datadog/templates/_helpers.tpl
```diff
@@ -15,6 +15,100 @@
...
+
+{{/*
+  Returns a semver-ish version for discovery defaulting.
+  Discovery reuses the chart's existing agent-version resolution for supported tags.
+  If that resolution still returns a non-semver-ish value, discovery treats it as latest.
+*/}}
+{{- define "get-agent-version-for-discovery" -}}
+{{- $tag := include "get-effective-agent-image-tag-for-discovery" . | trimSuffix "-jmx" -}}
+{{- $ctx := dict "Values" (dict "agents" (dict "image" (dict "tag" $tag))) -}}
+{{- $version := include "get-agent-version" $ctx -}}
```
brycekahle: `get-effective-agent-image-tag-for-discovery` + `trimSuffix "-jmx"` seems to do exactly what `get-agent-version` does with respect to the tag. I think you could just use `get-agent-version` without the extra helper and `ctx` creation.

vitkyrka (approved):

tedkahwaji (approved):

Jansen-w (approved):

PR #2570, opened 2026-04-14
[observability-pipelines-worker] bump to 2.15.1

lisaqvu:
## Summary
- Bump OPW helm chart to 2.15.1 (Chart.yaml, values.yaml, README, CHANGELOG)

## Test plan
- [ ] Helm template renders correctly with new version
- [ ] Image tag points to published 2.15.1 image

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Jira: https://datadoghq.atlassian.net/browse/OPA-5017

PR #2610, opened 2026-04-27
[observability-pipelines-worker] fix livenessProbe for gRPC API on :8686

Jansen-w:
#### What this PR does / why we need it

Upstream Vector replaced its HTTP/GraphQL observability API with a gRPC server ([vectordotdev/vector#24364](https://github.com/vectordotdev/vector/pull/24364)), and the change shipped in OPW image **2.15.0+**. The chart's default `httpGet :8686/health` probes are incompatible with the gRPC (HTTP/2) server now bound to 8686, so the kubelet probe receives an HTTP/2 SETTINGS frame, fails to parse it as HTTP/1, and reports:

> `Liveness probe failed: net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x12\x04..."`

Failure cadence: `delay 15s + (failureThreshold 5 × period 10s) = 65s` → SIGKILL → restart loop. Customers running chart **2.15.0** or **2.15.1** with default values (`workerAPI.enabled: true`, default probes) are affected — pods restart every ~1 minute in production.

This PR switches both probes to `tcpSocket` on :8686. TCP probes are protocol-agnostic, work on every Kubernetes version in the `install-chart` matrix, and don't require image changes.

#### Why `tcpSocket` and not `grpc`

`grpc:` probes would technically be more precise (Vector exposes the standard `grpc.health.v1.Health` service since [vectordotdev/vector#25139](https://github.com/vectordotdev/vector/pull/25139), and `set_not_serving()` is wired up for early-drain). However, `grpc:` probes require K8s ≥ 1.24 and would force dropping `1.22.17` from the install-chart matrix. `tcpSocket` ships the fix immediately for every supported K8s version with no matrix changes; we can iterate to `grpc:` in a follow-up if/when the matrix is pruned.

#### Validation

Tested locally on kind `v1.28.13` (matching the CI version that fails on main) with image `gcr.io/datadoghq/observability-pipelines-worker:2.15.1` and `ci/api-values.yaml`:

| Probe | Pod reaches Ready | helm install --wait | Restart count |
|---|---|---|---|
| `httpGet` (current main) | ❌ never | ❌ timeout 300s | climbs |
| `tcpSocket` (this PR) | ✅ ~51s | ✅ exit 0 | 0 |

Confirmed probe spec on running pod: `tcp-socket :8686 delay=15s timeout=15s period=10s #success=1 #failure=5`.

#### Why CI on #2570 didn't catch this

When chart 2.15.1 was bumped, [`lint-chart` failed](https://github.com/DataDog/helm-charts/actions/runs/24459496933/job/71469447427) (`chart version not ok. Needs a version bump!` because `2.15.1` was already on main). The `install-chart` job has `needs: [lint-chart, kubeconform-chart]`, so it was **skipped**. GitHub branch protection treats SKIPPED as passing, and the merge queue went through. The actual integration test never ran. Worth a follow-up to make `install-chart` a required check (or restructure the dependency) so this can't slip through again.

#### Customer mitigation while this PR is in flight

Affected users on chart 2.15.0 / 2.15.1 can override probes locally:

```yaml
livenessProbe:
  tcpSocket: { port: 8686 }
readinessProbe:
  tcpSocket: { port: 8686 }
```

Unblocks #2608 (icon URL migration) and any other PR touching the OPW chart, which currently hit this regression in `install-chart` CI.

#### Special notes for your reviewer

- Chart-only fix — no image change.
- `appVersion` stays at `"2.15.1"` (image not bumping).
- `version` bumps `2.15.1 → 2.15.2`, following the precedent of `2.14.1` ([#2492](https://github.com/DataDog/helm-charts/pull/2492)) and `2.13.3`/`2.13.4` for chart-only patches.

#### Checklist

- [x] Chart Version semver bump label has been added (`observability-pipelines-worker/patch-version`)


tedkahwaji (approved):

Jansen-w:
@codex review

lisaqvu (approved):

Jansen-w:
@codex review

PR #2664, opened 2026-05-15
[opw] Set default CPU/memory resource requests

vladimir-dd:

Sets default resource requests on the Observability Pipelines Worker
container so customers deploying the chart with no overrides get sane scheduling and HPA behavior out of the box:

```yaml
resources:
  requests:
    cpu: 2
    memory: 4Gi
```

**Limits are intentionally not set by default:**

- **No CPU limit** — A CPU limit would cause CFS throttling on the
  Worker's bursty pipeline workload. It would also break HPA:
  with `autoscaling.enabled: true`,
  `autoscaling.targetCPUUtilizationPercentage` computes `usage /
  requests.cpu`, but the throttled pod can't generate the utilization
  signal needed to scale up — you get throttling instead of scaling.
- **No memory limit** — OPW absorbs downstream sink backpressure in
  memory. A hard default cap (e.g. 4Gi) would risk OOMKilling existing
  installs that have organically grown above that ceiling on upgrade,
  and would kill the Worker exactly when buffering is most needed
  during downstream incidents. Recommended example values are kept as
  commented examples in `values.yaml`.

**Why this is also an HPA fix:** Today, any user who flipped
`autoscaling.enabled: true` without also setting `resources.requests.cpu`
got a silently broken HPA — the CPU utilization metric is computed as
`usage / requests.cpu`, which can't be evaluated when `requests.cpu` is
absent. The new `requests.cpu: 2` default makes the chart's existing
HPA defaults actually work.

#### Checklist

### DIFF
--- a/charts/observability-pipelines-worker/CHANGELOG.md
+++ b/charts/observability-pipelines-worker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.15.7
+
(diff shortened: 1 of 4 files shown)


gwenaskell (approved):
seems legit to me.

ckelner (approved):
LGTM! Thanks @vladimir-dd

PR #2686, opened 2026-05-26
chore(cloudprem): bump to 0.4.1 with multi-cloud docs + kubeVersion enforcement

Mallets:
## Summary

Doc-audit follow-up surfaced by an audit of the public BYOC Logs docs against the chart and the pomsky storage layer. Bundles four changes to the `cloudprem` chart so the README + Chart.yaml match the chart's actual multi-cloud support and the public docs.

Audit evidence: https://github.com/DataDog/claude-marketplace/pull/2216#issuecomment-4543572562

### Changes

1. **`kubeVersion: ">= 1.25.0-0"`** added to `Chart.yaml`. Today the K8s 1.25+ requirement is documented in two places (README Prereqs and the public install page) but not enforced — `helm install` accepts older clusters and customers hit obscure runtime failures. The `-0` suffix accepts cloud-vendor variants like `1.25.0-eks-abc`.

2. **README Prerequisites section** rewritten from AWS-only to multi-cloud:
   - The chart's `values.yaml` has had a full `azure:` block; the public install docs cover EKS / GKE / AKS / self-managed K8s; `quickwit-storage` registers S3 + Azure + GCS + S3-compat factories.
   - The README listing "AWS account / AWS Load Balancer Controller / S3 bucket" as bullet prereqs was out of sync with all of those.
   - New Prereqs cover supported clouds, both supported ingress controllers (ALB + NGINX), the per-cloud PostgreSQL options, and the full object-store list.

3. **`image.tag` default** in the values table — was documented as `devel`, actually defaults to the chart's `appVersion` when unset (currently `v0.1.26`). Customers reading the README expected one tag and got another.

4. **`azure.*` rows added** to the values table so the Azure configuration surface is discoverable from the README. (No `gcp.*` rows because GCP uses Workload Identity, wired via `serviceAccount.extraAnnotations` — no chart-key gap there.)

Both `README.md` and `README.md.gotmpl` are edited so the change survives whichever path future regeneration takes.

## Test plan

- [ ] `helm lint charts/cloudprem` passes locally.
- [ ] `helm install` against a K8s 1.24 cluster now fails with `chart requires kubeVersion ">= 1.25.0-0"`.
- [ ] `helm install` against a K8s 1.25+ cluster still installs cleanly with no `image.tag` override.
- [ ] Visual review of the regenerated `README.md` (e.g. via artifacthub preview) matches the values-table additions.
- [ ] CHANGELOG entry under `0.4.1` reads cleanly to the chart-team reviewer.

### DIFF
--- a/charts/cloudprem/CHANGELOG.md
+++ b/charts/cloudprem/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.4.1
+
+* Enforce `Kubernetes 1.25+` via `Chart.yaml` `kubeVersion` so `helm install` rejects older clusters with a clear error instead of failing later at runtime.
+* Rewrite the README Prerequisites section to reflect the chart's actual multi-cloud support (AWS / GCP / Azure / self-managed), supported ingress controllers (ALB + NGINX), and full object-storage list (S3, GCS, Azure Blob, MinIO, Ceph, any S3-compatible).
+* Fix the `image.tag` default documented in the README values table — it follows the chart's `appVersion`, not the literal string `devel`.
+* Add `azure.*` rows to the README values table so the Azure configuration surface is discoverable.
+
 ## 0.4.0
 
 * Update Docker image to `v0.1.26`.
--- a/charts/cloudprem/Chart.yaml
+++ b/charts/cloudprem/Chart.yaml
@@ -2,9 +2,15 @@ apiVersion: v2
 name: cloudprem
 description: Datadog CloudPrem
 type: application
-version: 0.4.0
+version: 0.4.1
 # This is the version of the "application". Right now, we follow the image version.
 appVersion: v0.1.26
+# Minimum Kubernetes version supported. The `-0` suffix accepts cloud-vendor
+# pre-release tags like `1.25.0-eks-abc`. Matches the prerequisite stated in the
+# public install docs (https://docs.datadoghq.com/byoc-logs/install/) and in the
+# README; surfacing it here makes `helm install` reject older clusters with a
+# clear error instead of failing later at runtime.
+kubeVersion: ">= 1.25.0-0"
 home: https://www.datadoghq.com/
 icon: https://static.datadoghq.com/static/images/logos/_datadog_avatar.svg
 maintainers:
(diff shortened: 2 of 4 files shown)


Mallets:
## CI status — initial run

Pushed `7091313` to fix the only legitimate failure (`lint-docs`); the rest of the red are expected.

| Check | Status | Note |
| --- | --- | --- |
| `lint-docs` | 🟢 should pass after `7091313` | `helm-docs` auto-expands `{{ template "chart.requirementsSection" . }}` (already present in `README.md.gotmpl:14`) into a `## Requirements` block now that `Chart.yaml` declares `kubeVersion`. The committed `README.md` was missing the rendered output. The `.gotmpl` itself was already correct. |
| `kubeconform-chart (v1.16.4 → v1.24.17)` | 🔴 expected — **this is the feature working** | Logs show `Error: chart requires kubeVersion: >= 1.25.0-0 which is incompatible with Kubernetes v<X>`. That is exactly what the `kubeVersion` constraint is meant to do at install time. The CI matrix lints every chart against every K8s version regardless of declared `kubeVersion`. |
| `kubeconform-chart (v1.25.16 → v1.31.1)` | 🟢 | All supported versions pass. |
| All other checks | 🟢 | Lint, version bump, mergegate, etc. — all green. |

## On the pre-1.25 kubeconform failures

I deliberately did **not** soften the `kubeVersion` constraint to make these pass, because that would defeat the whole point of the change (install-time rejection rather than runtime breakage on EOL K8s versions). Two follow-up options for the chart team to decide:

1. **Per-chart matrix exclusion** — update the workflow to skip kubeconform matrix entries for a chart when those versions fall outside the chart's declared `kubeVersion`. Cleanest long-term, but a workflow-wide change.
2. **Drop pre-1.25 versions from the matrix repo-wide** — K8s 1.16 was EOL in 2020-09, 1.18 in 2021-04, 1.22 in 2022-12, 1.24 in 2024-07. None are supported by any current cloud-managed K8s service. Reduces matrix run time too.

Happy to open a follow-up PR for either, but it touches the workflow file (`.github/workflows/lint-test.yaml`), which is out of scope for this audit follow-up.

Issue #2353, opened 2026-02-05
Datadog SVC has duplicate ports when deploying DDOT with NetworkPolicy successful deploy required


AkselAllas:
### Pre-submission Checklist

- [x] I have searched [existing issues](https://github.com/DataDog/helm-charts/issues) and this is not a duplicate
- [x] This is a Helm chart issue, not a Datadog product/service problem

### Helm chart name

datadog

### Helm chart version

3.131.0

### Bug Report

**What happened:**
This config does not work:
```
  datadog:
    otelCollector:
      enabled: true
      ports: []
    otlp:
      receiver:
        protocols:
          grpc:
            enabled: true
            endpoint: "0.0.0.0:4317"
          http:
            enabled: true
            endpoint: "0.0.0.0:4318"
```

**What I expected:**
service "datadog" to apply without duplicates.

What works:
```
  datadog:
    otelCollector:
      enabled: true
      ports: []
    otlp:
      receiver:
        protocols:
          grpc:
            enabled: true
            endpoint: "0.0.0.0:4317"
          http:
            enabled: true
            endpoint: "0.0.0.0:4318"
```

Problematic code:

1. Default values for otelCollector
<img width="766" height="281" alt="Image" src="https://github.com/user-attachments/assets/9b79818f-c0b4-40a3-8db7-c20251f7d226" />

2. svc https://github.com/DataDog/helm-charts/blob/datadog-3.131.0/charts/datadog/templates/agent-services.yaml#L91-L108

I need this config I presented due to Network Policy required by Istio
https://github.com/DataDog/helm-charts/blob/datadog-3.131.0/charts/datadog/templates/agent-network-policy.yaml#L28-L40

AkselAllas:
So I was using both the trace-agent and otel-collector. 

But still, the options for using otel-collector are not good.

It's not possible to get a Network Policy for otel-collector.

Issue #1237, opened 2023-11-13
No network policy rules created for OTLP


mseiwald:
**Describe what happened:**

The datadog chart's [network policy template](https://github.com/DataDog/helm-charts/blob/main/charts/datadog/templates/agent-network-policy.yaml) does not contain rules for the OTLP ports (4317 & 4318). When OTLP receivers are enabled, Datadog agent will not be able to receive any traffic on those ports.

**Describe what you expected:**

Allow rules for ports 4317 and 4318 should be added to the network policy when OTLP is enabled.

**Steps to reproduce the issue:**

1. Set `datadog.networkPolicy.create` to `true`
2. Set  `datadog.otlp.receiver.protocols.grpc.enabled` and `datadog.otlp.receiver.protocols.http.enabled` to `true`
3. Try to send OTLP messages to port 4317 or 4318 of the Datadog agent

**Additional environment details (Operating System, Cloud provider, etc):**

marcusteixeira:
https://github.com/DataDog/helm-charts/commit/e2698925b5af8ff7ccf07ef8f76c3e70a653d1f5 (Fixed in version ## 3.120.1 of Datadog Chart)

PR #2696, opened 2026-05-31
[datadog] Include otelCollector ports in agent NetworkPolicies

jumpe1:

`datadog.otelCollector.ports` is already reflected in the agent Service (`agent-services.yaml`, guarded by `should-enable-otel-agent`), but the NetworkPolicy templates only open ports for the *native* Agent OTLP receiver (`datadog.otlp.receiver.protocols.grpc/http.enabled`) — a different code path from the DDOT collector.

As a result, on clusters where these policies are enforced (Calico, Cilium, GKE Dataplane V2), ingress traffic to the embedded DDOT receiver is silently dropped: the otel-agent container is `Ready` and the OTLP ports are exposed as container ports (and as hostPorts when configured), while OTLP clients still fail to connect.

```bash
telemetrygen logs --otlp-endpoint=$(HOST_IP):4317 --otlp-insecure  # fails with DeadlineExceeded
```

This PR adds the ports declared in `datadog.otelCollector.ports` as ingress rules to both the agent `NetworkPolicy` and `CiliumNetworkPolicy`, mirroring the existing `agent-services.yaml` logic under the same `should-enable-otel-agent` guard. This also removes the need for the workaround of enabling the native Agent OTLP receiver just to open NetworkPolicy ingress for DDOT ports, which can cause duplicate Service ports (#2353).

* Relates to #1237
* Relates to #2353

* No `values.yaml` changes; this reuses the existing `datadog.otelCollector.ports` and adds no new API surface.
* NetworkPolicy ingress entries are additive allow rules, so duplicated rules are harmless when both the native receiver and the DDOT collector are enabled with the same ports — unlike the Service-level duplication in #2353, where Kubernetes rejects two Service ports sharing the same number/protocol.
* Added a baseline values file combining `networkPolicy` + `otelCollector` (`test/datadog/baseline/values/otel-agent_network_policy.yaml`) and regenerated baselines; no pre-existing baseline manifests changed.

#### Checklist
 - [ ] Chart Version semver bump label has been added (use `datadog/patch-version`)

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.215.2
+
+* Add ports defined in `datadog.otelCollector.ports` to the agent NetworkPolicy and CiliumNetworkPolicy ingress when `datadog.otelCollector.enabled` and `datadog.networkPolicy.create` are set, so OTLP traffic reaches the embedded DDOT collector on clusters that enforce NetworkPolicy.
+
 ## 3.215.1
 
 * Add `datadog.kubeStateMetricsCore.useApiServerCache` to enable the use of the API server cache in the Kube Metrics Core check.
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.239.2
+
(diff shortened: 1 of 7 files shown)


jumpe1:
@tedkahwaji Rebased onto `main` and regenerated the baseline — `make unit-test-datadog` passes locally. Could you take a look when you have a moment? Since this is a fork PR the workflows need approval to run, so no CI has executed yet.

jumpe1:
Rebased again onto `main`; chart bumped to 3.239.2 and `make unit-test-datadog` passes locally.

@tedkahwaji 
Sorry to follow up on this. Since the PR comes from a fork, the workflows stay in `action_required` and CI can't start on its own — that's likely why this one looks stalled. Whenever you (or anyone with the access) have a spare moment, could the workflow run be approved and the `datadog/patch-version` label added? No rush on the review itself; I'd just like to get CI green so it's easier to look at.

tedkahwaji (approved):

PR #2717, opened 2026-06-08
[CONTP-1710] Install DatadogInstrumentation CRD and add RBAC permissions when controller is enabled

Mathew-Estafanous:
Adds support for installing the `DatadogInstrumentation` CRD and required RBAC permissions when `datadog.instrumentationCrd.enabled` is set.

The new datadog-crds dependency in `requirements.yaml` is intentional. Helm resolves conditions left-to-right and stops at the first non-nil value, so if any condition is true then all the CRDs would be installed, even if the other two conditions are false (and therefore unused).

Instead of having this side effect, I've opted to split out the `datadog-crds` dependency into a second one that is just for  `datadog.instrumentationCrd.enabled`, which is expected to be on by default.

  - fixes #

#### Checklist

### DIFF
--- a/charts/datadog/requirements.yaml
+++ b/charts/datadog/requirements.yaml
@@ -2,7 +2,10 @@ dependencies:
   - name: datadog-crds
     version: 2.21.0
     repository: https://helm.datadoghq.com
-    condition: datadog.autoscaling.workload.enabled,clusterAgent.metricsProvider.useDatadogMetrics
+    # Ordering matters here. Helm uses the first condition with a defined (non-nil) value. Conditions with an
+    # explicit false default (e.g. clusterAgent.metricsProvider.useDatadogMetrics) must come last, otherwise
+    # the condition stops evaluation even if the subsequent condition would've been true.
+    condition: datadog.autoscaling.workload.enabled,datadog.instrumentationCrd.enabled,clusterAgent.metricsProvider.useDatadogMetrics
     tags:
     - install-crds
   - name: kube-state-metrics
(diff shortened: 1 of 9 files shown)


charts/datadog/values.yaml
```diff
@@ -1439,6 +1439,11 @@ datadog:
     #   - Do not install the CSI driver separately if this is enabled, or you may hit conflicts.
     enabled: false
 
+  instrumentationCrd:
+    # datadog.instrumentationCrd.enabled -- Enable the DatadogInstrumentation CRD controller and reconciliation platform.
+    # Requires version 7.80.0 or later of both cluster and node agent.
+    enabled:
+
```
Mathew-Estafanous, line 1451: Honestly not a big fan of the name here, but couldn't come up with anything better.

When it's switched on by default I don't expect many customers to disable it as this is more of a platform level feature not a product.

charts/datadog/requirements.yaml
```diff
@@ -5,6 +5,13 @@ dependencies:
     condition: datadog.autoscaling.workload.enabled,clusterAgent.metricsProvider.useDatadogMetrics
     tags:
     - install-crds
+  - name: datadog-crds
+    version: 2.21.0
+    repository: https://helm.datadoghq.com
+    alias: datadog-crds-instrumentation
+    condition: datadog.instrumentationCrd.enabled
```
Mathew-Estafanous: This `datadog-crds` dependencies is split from the above because `datadog.instrumentationCrd.enabled` is expected to be enabled by default and we'd like to avoid installing other CRDs as a side effect.

Splitting it ensures only the CRD that's needed (`DatadogInstrumentation`) is installed. Otherwise the DPA and metrics crds would be installed on all customer's clusters by default.

levan-m (approved):

PR #2742, opened 2026-06-17
default to CNM direct send if >=7.81.0

brycekahle:

Enables CNM direct send if the agent version is >= 7.81.0

#### Checklist


gjulianm (approved):

charts/datadog/templates/_helpers.tpl
```diff
@@ -97,6 +97,17 @@ false
 {{- end -}}
 {{- end -}}
 
+{{- define "cnm-use-direct-send" -}}
+{{- if not .Values.agents.image.doNotCheckTag -}}
```
gpalmz, line 101: With `doNotCheckTag=true` this always renders `false`, so a pinned `>=7.81` image or `latest` can't get direct send. The guard avoids `semverCompare` on a non-semver tag, but it's more conservative than discovery defaulting (which treats `latest` as enabled via `get-agent-version-for-discovery`). Intentional?

gpalmz (approved):
LGTM, left one minor question on the helper.

charts/datadog/templates/_helpers.tpl
brycekahle: Yes, this matches the operator behavior of defaulting to `false` if we cannot parse the tag.

brycekahle: Once 7.81 is released, we may change this behavior to allow `latest`

PR #2764, opened 2026-07-01
Add GKE COS NVIDIA library mounts

gjulianm:

Mounts GKE COS NVIDIA driver libraries into the Agent and system-probe containers when GPU monitoring runs with `providers.gke.cos=true`.

This lets GPU monitoring find NVIDIA libraries on GKE COS nodes where they live under `/home/kubernetes/bin/nvidia/lib64`.

Validated with `helm template` and a temporary deployment on a GKE COS GPU cluster. The Agent DaemonSet reached `3/3 Running`, and the Agent container could read NVIDIA libraries from the mounted path.

The test cluster did not have a `nvidia` RuntimeClass, so deployment validation used `--set-string datadog.gpuMonitoring.runtimeClassName=`.

#### Checklist

- [x] All commits are signed and show as "Verified" on GitHub (see: [signing commits](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))

### DIFF
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -337,7 +337,7 @@
       mountPath: /var/run/nvidia-container-devices/all
     {{- if .Values.providers.gke.cos }}
     - name: gke-nvidia-driver-lib64
-      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu
+      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu # the agent automatically looks for the nvidia driver in this path, and it's also not likely to conflict with other libraries
       readOnly: true
     {{- end }}
     {{- end }}
--- a/charts/datadog/templates/_container-agent.yaml
+++ b/charts/datadog/templates/_container-agent.yaml
@@ -335,6 +335,11 @@
       readOnly: false
     - name: gpu-devices
       mountPath: /var/run/nvidia-container-devices/all
+    {{- if .Values.providers.gke.cos }}
+    - name: gke-nvidia-driver-lib64
+      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu
+      readOnly: true
+    {{- end }}
     {{- end }}
     {{- if not .Values.providers.gke.gdc }}
     - name: dsdsocket
--- a/charts/datadog/templates/_daemonset-volumes-linux.yaml
+++ b/charts/datadog/templates/_daemonset-volumes-linux.yaml
@@ -249,6 +249,12 @@
 - name: gpu-devices
   hostPath:
     path: /dev/null
+{{- if .Values.providers.gke.cos }}
+- name: gke-nvidia-driver-lib64
+  hostPath:
+    path: /home/kubernetes/bin/nvidia/lib64
+    type: DirectoryOrCreate
+{{- end }}
 {{- end }}
 {{- if .Values.providers.openshift.controlPlaneMonitoring }}
 - name: etcd-certs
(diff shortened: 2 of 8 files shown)


raymondeah (approved):

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -249,6 +249,12 @@
 - name: gpu-devices
   hostPath:
     path: /dev/null
+{{- if .Values.providers.gke.cos }}
+- name: gke-nvidia-driver-lib64
+  hostPath:
+    path: /home/kubernetes/bin/nvidia/lib64
+    type: DirectoryOrCreate
```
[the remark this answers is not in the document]
gjulianm: GPU Monitoring doesn't work on autopilot/GDC in general so it doesn't make sense to gate this.

charts/datadog/templates/_container-agent.yaml
```diff
@@ -335,6 +335,11 @@
       readOnly: false
     - name: gpu-devices
       mountPath: /var/run/nvidia-container-devices/all
+    {{- if .Values.providers.gke.cos }}
+    - name: gke-nvidia-driver-lib64
+      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu
```
mbertrone: question: the container path `/host/run/nvidia/driver/usr/lib/x86_64-linux-gnu` has to match what the agent's NVML detection actually searches, and nothing in the chart wires the two together. Grepping the whole chart, this path appears only in these three template lines (here, `_container-system-probe.yaml:153`, and the hostPath source at `_daemonset-volumes-linux.yaml:255`); there's no `LD_LIBRARY_PATH` or path-pointing env var, so the agent binary must have this directory hardcoded in its search logic. The `/host/run/nvidia/driver` prefix mirrors the GPU-Operator driver-container root (`/run/nvidia/driver`) with the chart's `/host` host-mount convention, so it reads as deliberate rather than arbitrary.

Could you confirm against the agent's NVML library-search code that this exact directory is on the search path? The live validation shows the `.so` files are present and readable at the mount, but not that NVML loaded them. Worth confirming GPU metrics flow end-to-end, not just that `ls` finds the libraries.

charts/datadog/templates/_daemonset-volumes-linux.yaml
```diff
@@ -249,6 +249,12 @@
 - name: gpu-devices
   hostPath:
     path: /dev/null
+{{- if .Values.providers.gke.cos }}
+- name: gke-nvidia-driver-lib64
```
mbertrone, line 253: suggestion (non-blocking): this new `cos` block renders in zero baselines, so it ships with no regression protection. The only GPU baseline, `test/datadog/baseline/values/gpu_monitoring.yaml`, enables `gpuMonitoring.{enabled,privilegedMode,configureCgroupPerms}` but never sets `providers.gke.cos`, and `baseline_test.go` auto-discovers values files from that directory — so `make update-test-baselines` produces no diff here precisely because nothing exercises the branch. Correctness rests on an exact path string that no golden manifest currently pins.

Add a `test/datadog/baseline/values/gpu_monitoring_cos.yaml` (GPU monitoring + privilegedMode + `providers.gke.cos: true`) and regenerate; the harness picks it up automatically and locks in both mounts plus the volume. Low effort for a mount whose whole value depends on that path being right.

mbertrone (approved):

charts/datadog/templates/_container-agent.yaml
gjulianm: Yes, this is part of the default search paths of the agent. I deliberately chose this mount path as it's less likely to conflict with other libraries. Adding a comment mentioning it.

charts/datadog/templates/_daemonset-volumes-linux.yaml
gjulianm: Added!

PR #2776, opened 2026-07-08
[observability-pipelines-worker] expose termination grace period to the worker

clementd-dd:

in 2.19, OPW will honor a termination grace period provided through an environment variable. This commit exposes the pod termination grace period to the worker.

Previously this limit was hardcoded to 60s.

OPW will stay up and try to drain its buffers for up to the configured threshold before exiting. This is useful when the pod termination grace period is longer than 60s. Before this change, the worker would not take advantage of this extra time.

Note: this is not released in the worker yet, and will be part of 2.19.

Note: for now this duplicates the pod termination grace period config, but maybe more control would be useful?

#### Checklist

### DIFF
--- a/charts/observability-pipelines-worker/CHANGELOG.md
+++ b/charts/observability-pipelines-worker/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 2.19.0
+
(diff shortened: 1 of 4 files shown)


gwenaskell:
handled in https://github.com/DataDog/helm-charts/pull/2814

PR #2782, opened 2026-07-09
[CONTP-1612] support enabling untaint controller in datadog chart

adel121:

For Helm-managed clusters (agent installed via charts/datadog, no operator managing the agent), users need to opt the agent into tolerating the startup taint so it can schedule on tainted nodes. The untaint controller's pod label selector already works for Helm-managed agents — charts/datadog uses agent.datadoghq.com/component: agent on the node agent DaemonSet — so no label change is needed, only the toleration.

In this PR, we:
- Add `operator.untaintController.enabled` value (default false) to values.yaml
- Inject the toleration into the agent `DaemonSet` template when enabled, and activate the untaint controller feature in the operator.
- Add Helm unit test verifying the toleration appears in the rendered `DaemonSet`

#### Testing

Added unit test

Manual testing:

Deploy the agent using helm chart with the added option enabled. Ensure the toleration is added properly:

Deployed agent with appended at the end of `values.yml`:
```
operator:
  untaintController:
    enabled: true
```

Checked that the toleration was properly added:
```
      tolerations:
      - effect: NoSchedule
        key: agent.datadoghq.com/not-ready
        operator: Equal
        value: presence
```

The CSI driver may also need a similar change, but since CSI is usually used as a subchart (dependency of the agent chart), and the user can already set tolerations manually on the CSI by editing `datadog_csi_driver` section in datadog values.yaml, I am keeping it for a future PR to asses if it is worth adding a separate config option for CSI or not. 

#### Checklist


charts/datadog/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.238.0
+
+* Add `operator.untaintController.enabled` (default `false`). When enabled, the node Agent DaemonSet tolerates the `agent.datadoghq.com/not-ready=presence:NoSchedule` startup taint and the Datadog Operator untaint controller is enabled to remove that taint once the Agent is ready. Requires Operator v1.28.0+.
```
levan-m: nit - can we add doc reference similar to operator changelog?

levan-m (approved):
this looks good for when agent is enabled so approving. We probably need to add toleration for CSI too right?

adel121:
> this looks good for when agent is enabled so approving. We probably need to add toleration for CSI too right?

Yes I think CSI also needs one. But CSI is more complex because it is a subchart of the agent. I don't think there is a straight-forward way to pass `operator.untaintController.enabled` as a value to the csi chart. 

The only way I see is duplicating the config, that is: add `datadog_csi_driver.untaintController.enabled` , which I personally don't like. I preferred to keep CSI out of scope for now because of 2 things:
- CSI adoption is almost negligible compared to Agent adoption, and on top of this, the untaint controller is not something enabled by default or even expected to be widely adopted.
- Customers who really need to implement this for CSI driver may simply manually set the toleration via `datadog_csi_driver.tolerations` (see [here](https://github.com/DataDog/helm-charts/blob/5bd0c3ccb56f1a58f283105b81a725ea380e148c/charts/datadog-csi-driver/values.yaml#L87-L91))

PR #2790, opened 2026-07-15
[datadog] Fix GKE COS system-probe NVML mount

gjulianm:

Mounts the GKE COS NVIDIA driver library volume at system-probe's `HOST_ROOT` path so GPU monitoring can find NVML when `providers.gke.cos=true`.

Validated on a GKE COS T4 cluster: the Agent GPU check found NVML and emitted GPU metrics, while system-probe required this `HOST_ROOT` path for its own NVML lookup.

#### Checklist

[1]: https://docs.github.com/en/authentication/managing-commits/signing-commits

### DIFF
--- a/charts/datadog/templates/_container-system-probe.yaml
+++ b/charts/datadog/templates/_container-system-probe.yaml
@@ -150,9 +150,6 @@
       mountPath: /var/run/nvidia-container-devices/all
 {{- if .Values.providers.gke.cos }}
     - name: gke-nvidia-driver-lib64
-      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu
-      readOnly: true
-    - name: gke-nvidia-driver-lib64
       mountPath: /host/root/run/nvidia/driver/usr/lib/x86_64-linux-gnu
       readOnly: true
 {{- end }}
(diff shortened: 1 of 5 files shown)


charts/datadog/templates/_container-system-probe.yaml
```diff
@@ -150,7 +150,7 @@
       mountPath: /var/run/nvidia-container-devices/all
 {{- if .Values.providers.gke.cos }}
     - name: gke-nvidia-driver-lib64
-      mountPath: /host/run/nvidia/driver/usr/lib/x86_64-linux-gnu
+      mountPath: /host/root/run/nvidia/driver/usr/lib/x86_64-linux-gnu
```
brycekahle, line 153: Shouldn't this already be available if the entire `/` is mounted at `/host/root`?

gjulianm: In this case we're not mounting the same directory, but instead mounting `/home/kubernetes/nvidia/lib464` into this path. I'll update the search paths in the agent too but agents without that change require this mount to be defined.

swang392 (approved):

brycekahle (approved):

PR #2815, opened 2026-07-23
Add cumulativetodelta processor to default DDOT Collector metrics pipelines

RiantZ:

Adds the `cumulativetodelta` processor to the default DDOT сollector configuration rendered by the `datadog` chart, in the metrics pipelines that export to the Datadog exporter:
- standalone Agent config: `metrics` and `metrics/prometheus` pipelines;
- Gateway config: `metrics` pipeline.

DDOT is moving to the metrics v3 payload format. In v3 the Datadog exporter stops converting cumulative counters to delta internally, so a `cumulativetodelta` processor in the pipeline becomes mandatory. Adding it to the default config now is safe before v3 — the exporter consumes already-delta input directly, without a second conversion — and prepares the chart for the v3 switch.

Placement — the processor is added only where the `datadog` exporter is present:
- It is not added to the daemonset Agent config used in gateway mode, because that config
  forwards metrics to the Gateway over `otlphttp`; the Gateway performs the conversion. This
  matches the chart's existing minimal-forwarding-agent design.
- It is not added to `traces`/`logs` pipelines. The processor only implements metrics
  processing, and the collector fails to start if it is placed in a traces/logs pipeline.

#### Checklist
- [x] For datadog or datadog-operator chart or value changes, update the test baselines (run: make update-test-baselines)
- [ ] For datadog chart changes, received ✅ from a member of your team

https://datadoghq.atlassian.net/browse/OTAGENT-1178

### DIFF
--- a/charts/datadog/CHANGELOG.md
+++ b/charts/datadog/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Datadog changelog
 
+## 3.232.0
+
+* Add the `cumulativetodelta` processor to the default DDOT Collector metrics pipelines (standalone Agent and Gateway), converting cumulative counters and histograms to delta upstream of the Datadog exporter. This prepares DDOT for the metrics v3 payload format, where the exporter no longer performs this conversion internally.
+
 ## 3.231.6
 
 * Add `privateActionRunner.apiKeyOnlyEnrollment` for node agent and cluster agent, wiring it to `api_key_only_enrollment` in the PAR ConfigMap and `DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT` in the cluster agent deployment respectively.
--- a/charts/datadog/Chart.yaml
+++ b/charts/datadog/Chart.yaml
@@ -1,7 +1,7 @@
 ---
 apiVersion: v1
 name: datadog
-version: 3.231.6
+version: 3.232.0
 appVersion: "7"
 description: Datadog Agent
 keywords:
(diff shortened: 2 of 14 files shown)


mackjmr (commented):
Do we need this change in helm / operator ? If we want to add it always, just adding it in the converter should be enough, and we would need to only maintain it in one place ?

RiantZ:
> Do we need this change in helm / operator ? If we want to add it always, just adding it in the converter should be enough, and we would need to only maintain it in one place ?

Automatic converter injection does not cover everything (as far I know, if I'm wrong - please tell me)

Standalone Agent (no gateway): the converter will inject c2d itself — but only from a future agent version (the datadog-agent converter change isn't in any release yet). Since the c2d processor is already compiled into every DDOT image, writing it in the config works on any version immediately. Here the explicit helm change is redundant once that agent version ships, but adds version-independence and visibility.

Gateway: the converter won't inject it at all. The chart overrides DD_OTELCOLLECTOR_CONVERTER_FEATURES (zpages,pprof,datadog), replacing the default list and omitting c2d. So the gateway needs either explicit config or c2d added to that env — otherwise it never gets it.

Core reason: the processor is in every image; converter auto-injection is a new, version-specific feature.

@mackjmr

mackjmr:
> Standalone Agent (no gateway): the converter will inject c2d itself — but only from a future agent version (the datadog-agent converter change isn't in any release yet)

I guess this depends when we want to support this by, wait for new agent version or add it now via helm but remove when new agent version is out.

> Gateway: the converter won't inject it at all.

Can we inject it in converter for gateway ?

If we want to support this asap without waiting for new agent image im fine with adding it now as long as we add a todo to remove it with new agent version. But for changes like this that need to be added everywhere, I prefer that we add it only in the converter. Otherwise if we update the config we need to update it in 3/4 different places rather than just the converter.

PR #2818, opened 2026-07-23
fix(gpu): force disable eBPF probes

betterengineering:
This commit forces the disablement of the eBPF probes when privileged mode is enabled.

The eBPF probes for GPU Monitoring are deprecated. Some customers still have the previously recommended setting gpu.privilegedMode enabled. These probes have caused crashes on NVIDIA GB300 Grace Blackwell Ultra machines. We want to disable the probes by default moving forward so they no longer cause problems. We still need to support privilegedMode in some environments for cgroup permission patching.

We've had two customer cases of this and we want to prevent further cases.

#### Checklist


charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -90,6 +90,9 @@ data:
     gpu_monitoring:
       enabled: {{ and $.Values.datadog.gpuMonitoring.enabled $.Values.datadog.gpuMonitoring.privilegedMode }}
       configure_cgroup_perms: {{ $.Values.datadog.gpuMonitoring.configureCgroupPerms }}
+      {{- if $.Values.datadog.gpuMonitoring.privilegedMode }}
+      enable_ebpf_probes: false
```
mbertrone: question: this forces `enable_ebpf_probes: false` (plus the matching core-agent env var) with no `values.yaml` key to override it. A customer who set `privilegedMode: true` specifically to run the eBPF probes loses them on upgrade with no supported way back, and the config-file value in particular can't be reached through Helm values at all. Is the hard force intentional? From the description it reads as deliberate (the probes are deprecated), which is fine to ship. If so, a one-line note in the changelog/values calling it a deliberate kill-switch would help. And if we expand the eBPF path again in the future, we'd want to reintroduce a proper `gpuMonitoring.enableEbpfProbes` flag defaulting to false, rather than un-hardcoding this after the fact.

charts/datadog/templates/system-probe-configmap.yaml
```diff
@@ -90,6 +90,9 @@ data:
     gpu_monitoring:
       enabled: {{ and $.Values.datadog.gpuMonitoring.enabled $.Values.datadog.gpuMonitoring.privilegedMode }}
       configure_cgroup_perms: {{ $.Values.datadog.gpuMonitoring.configureCgroupPerms }}
+      {{- if $.Values.datadog.gpuMonitoring.privilegedMode }}
```
mbertrone: nit (non-blocking): this guard keys off `privilegedMode` alone, while the `enabled:` line right above (L91) and every other GPU/system-probe gate in the chart use `(and gpuMonitoring.enabled gpuMonitoring.privilegedMode)`, e.g. the `configure_cgroup_perms` mount guards at L221/L354 and `should-enable-system-probe` in `_helpers.tpl`. Harmless in practice: the only divergent combo is `enabled: false, privilegedMode: true` with system-probe pulled in by another feature, which just emits a stray `enable_ebpf_probes: false` under a disabled module. Matching L91's condition keeps the block internally consistent and saves a future reader from wondering whether the asymmetry is meaningful.

mbertrone (approved):
LGTM — the configmap `enable_ebpf_probes: false` correctly stops the deprecated probes from loading. Two non-blocking notes inline.

charts/datadog/templates/system-probe-configmap.yaml
betterengineering: Updated to have a flag defaulting to false

charts/datadog/values.yaml
```diff
@@ -1209,6 +1209,12 @@ datadog:
     # datadog.gpuMonitoring.configureCgroupPerms -- Configure cgroup permissions for GPU monitoring
     configureCgroupPerms: false
 
+    # datadog.gpuMonitoring.enableEbpfProbes -- DEPRECATED. Enable the GPU monitoring eBPF probes in system-probe
+    # The eBPF probes are deprecated and disabled by default, even when `datadog.gpuMonitoring.privilegedMode` is
+    # enabled. This option only applies in privileged mode and exists so that users who still rely on the probes can
+    # opt back in; expect it to be removed in a future release.
+    enableEbpfProbes: false
```
Mathew-Estafanous, line 1216: I'm not familiar with gpu monitoring using EBPF probes, but this comment sounds like this could be a breaking change for customers using the deprecated eBPF probes. If they upgrade their helm version without setting this new config, then they'd see everything suddenly stop, correct?

Is that ok, especially for a 'patch' version bump?

Mathew-Estafanous (approved):
change looks good. left one question

charts/datadog/values.yaml
betterengineering: In terms of the metrics they receive, it will be a noop. We have several ways to collect the same metrics and the eBPF probes have the least precedence. I agree it should be a minor bump, will update.

PR #2831, opened 2026-08-01
Add component to CloudPrem OTLP resources

hushy:
## Summary

Add the CloudPrem workload component to `OTEL_RESOURCE_ATTRIBUTES` using the existing `KUBERNETES_COMPONENT` downward-API environment variable.

This replaces the broader Pomsky-side implementation in DataDog/pomsky#729 for the OTLP telemetry path. The OpenTelemetry SDK reads `OTEL_RESOURCE_ATTRIBUTES`, so no application-level configuration plumbing is needed.

After rebasing onto chart `0.5.0`, the new `component` attribute is composed with the existing `chart.version` attribute. The chart is bumped to `0.5.1` as required by the repository release checks.

## Validation

- `helm lint charts/cloudprem --values charts/cloudprem/ci/kubeconform-values.yaml`
- Rendered the indexer with BYOC telemetry enabled and verified `OTEL_RESOURCE_ATTRIBUTES` contains both `component=$(KUBERNETES_COMPONENT)` and `chart.version=0.5.1`.
- Verified `KUBERNETES_COMPONENT` is declared before `OTEL_RESOURCE_ATTRIBUTES`.
- `git diff --check`

### DIFF
--- a/charts/cloudprem/CHANGELOG.md
+++ b/charts/cloudprem/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+## 0.5.1
+
+* Add the CloudPrem component name to OpenTelemetry resource attributes.
+
 ## 0.5.0
 
 * **Breaking**: remove the `medium` pod size. `indexer.podSize` / `searcher.podSize` now accept `large`, `xlarge`, `2xlarge`, `4xlarge`, `6xlarge`, and `8xlarge` only; rendering fails with an explicit error if any other size is set. Move to `large` before upgrading.
--- a/charts/cloudprem/Chart.yaml
+++ b/charts/cloudprem/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: cloudprem
 description: Datadog CloudPrem
 type: application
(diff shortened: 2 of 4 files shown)


Mallets (commented):
LGTM

Note that this PR has no effect since `OTEL_RESOURCE_ATTRIBUTES` is not considered in `quickwit-metrics`. Nonetheless, I'm ok to capture the intended behaviour right now while the support for env-provided metrics labels is being added. Most probably `OTEL_RESOURCE_ATTRIBUTES` will be replaced by another env var more generic to all exporters.

Maybe I'd remove the changelog addition to not give the impression this PR has any effect.

Mallets (approved):

guilload:
It doesn't even work for logs?

Mallets:
My bad. The `OTEL_RESOURCE_ATTRIBUTES` is read directly from the `opentelemetry-otlp` dependency and not by `quickwit`. So, it should be good in this case.

In the future it would be good to harmonise this behaviour across exporters though.

shuheiktgw (approved):

PR #2902, opened 2026-09-04
feat(datadog): set per-remote-agent log levels on the Agent container

rahulkaukuntla:

Re-targets `agents.containers.<name>.logLevel` for the containers that consume the Agent's config stream. `DD_LOG_LEVEL` is rendered on every container from `agents.containers.<name>.logLevel | default datadog.logLevel`, so it is divergent by design and cannot be hoisted to a single Agent-container value. Instead, where a remote container's `logLevel` is set, the corresponding namespaced variable is also rendered on the Agent container — `DD_SECURITY_AGENT_LOG_LEVEL`, `DD_PROCESS_CONFIG_LOG_LEVEL`, `DD_APM_LOG_LEVEL` — so the Agent streams the per-agent level.

Without this, streaming discards the remote agent's own `DD_LOG_LEVEL` and every remote agent falls back to the global level. The values stay where they are and `logLevel` is not deprecated. The mirrored variables are gated on `datadog.configStream.enabled` through the `configstream-enabled` helper from #2919.

**System Probe is omitted deliberately.** `system_probe_config.log_level` is not known to the core Agent's config object, so the Agent never streams the key and setting it on the Agent container would be inert. Its container `logLevel` keeps working outside the stream, so it is not regressed — it is simply the one member of the family that cannot be made live-reconfigurable this way.

**Third of five**, each targeting the previous one: #2896 → #2919 → this PR → #2901 → #2900. Needs DataDog/datadog-agent#55903, which declares and consumes the three namespaced keys; until an Agent carrying it ships, the vars are unbound and therefore ignored. The version bump, changelog and README updates for all five are on #2900.

**Testing.** `make unit-test-datadog` — `ok github.com/DataDog/helm-charts/test/datadog`, with a new baseline exercising a per-container `logLevel`, pinned above the gate's 7.85 floor.

#### Checklist

- [x] Chart Version semver bump label has been added (`datadog/no-version-bump`; the bump is on #2900)
- [x] Test baselines have been updated


charts/datadog/templates/_helpers.tpl
```diff
@@ -1981,3 +1981,25 @@ streamed snapshot reports the value each container actually started with.
...
 {{- end -}}
+
+{{/*
+Per-remote-agent log levels, set on the Agent container so the config stream carries them.
+system-probe is omitted deliberately — system_probe_config.log_level is not known to the core
+Agent's config, so it is never streamed; its container logLevel keeps working locally.
+*/}}
+{{- define "remote-agent-log-level-env" -}}
+{{- if eq (include "configstream-enabled" .) "true" }}
+{{- if .Values.agents.containers.securityAgent.logLevel }}
```
fanny-jiang, line 1990: Non-blocking: just checking if this fallback change is intentional. With streaming enabled, remote agents without an explicit `logLevel` now inherit the core Agent’s effective log level rather than `datadog.logLevel`. For example, setting only `agents.containers.agent.logLevel: debug` would also make them use `debug`, even if `datadog.logLevel` is `info`. Either behavior is fine with me—just wanted to confirm this is expected.

fanny-jiang (approved):

charts/datadog/templates/_helpers.tpl
rahulkaukuntla: Yes, that's intentional — and not really avoidable. The stream carries the core Agent's resolved `log_level`, and the consumer wipes each remote container's own env layer before applying the snapshot, so the `DD_LOG_LEVEL` we render on the remote container is discarded. The core Agent's value comes from `agents.containers.agent.logLevel | default datadog.logLevel `(`_container-agent.yaml:94`), so your example is right: with only agents.containers.agent.logLevel: debug set, the remotes follow it.

That's why the namespaced vars in this PR exist — `agents.containers.<name>.logLevel` still wins, it just gets there through `DD_SECURITY_AGENT_LOG_LEVEL / DD_PROCESS_CONFIG_LOG_LEVEL / DD_APM_LOG_LEVEL` on the Agent container instead of `DD_LOG_LEVEL `on its own. The only case that changes is "no per-container logLevel, and `agents.containers.agent.logLevel` differs from `datadog.logLevel`", which seems like a reasonable reading of an explicit Agent-level override anyway.

PR #2925, opened 2026-09-17
[datadog-operator] Update for operator v1.31.0-rc.1

gabedos:


avonengel (approved):

gabedos:
The workflow's CI failed because of a unit test failure... we should probably still create this PR automatically when there are unit test failures to make the process easier.

Creating this PR locally with the script for some reason wasn't working smoothly because the dependencies were taking forever to install and led to commands timing out and the final steps of the script never finished.

Adding this to the log of workflow improvements

